PR #29, opened 2026-04-03
feat: block all routes until initial setup is complete

ravirajsinh45:
## Summary

- Backend `SetupGuardMiddleware` returns 503 for all API routes until a superadmin exists
- Frontend middleware redirects to `/setup` if `needs_setup` is true
- Exempt paths: `/setup/*`, `/health`, `/docs`, `/share/*`
- Cached after first check — zero DB overhead after setup is done

## Test plan

- [x] Fresh instance (no users) → all routes return 503 except `/setup/status` and `/health`
- [x] Navigate to any page → redirected to `/setup`
- [x] Complete setup → all routes work normally
- [x] Restart server → cached flag resets, re-checks DB once, then caches again


Kashisk2 (approved):
LGTM

PR #70, opened 2026-05-21
Fix public share comments response handling

### LINKED ISSUE #67: Single-asset public share: comments never render due to response-shape mismatch
Reported by @ducknoodledance in #62.

## Problem
Open a public single-asset share link → comments panel is always empty,
even when comments exist. Project / folder shares are unaffected.

## Root cause
Shape mismatch between API and the single-asset share page:

- API (`apps/api/routers/comments.py:567`) returns a bare list:
  ```py
  ```
- Frontend (`apps/web/app/share/[token]/page.tsx:200-207`) expects an
  object:
  ```ts
  .then((data: CommentsResponse) => setComments(data.comments ?? []))
  ```
  `data.comments` is `undefined` on an array, so it silently falls
  back to `[]`.

The folder-share viewer uses a different code path (`ShareReviewScreen`)
that handles the array correctly, which is why project shares work.

## Proposed fix
Either:
- (frontend) treat the response as an array: `setComments(Array.isArray(data) ? data : data.comments ?? [])`, or
- (backend) wrap the response in `{ comments: [...] }` — but this
  would be a breaking change for any other consumer.

Frontend fix is the cleaner option here.

GChaucer:
## Summary
- Normalize public share comments responses on the single-asset share page
- Support both the current bare array response and `{ comments: [...] }` fallback shape
- Keep behavior aligned with the folder share viewer

Fixes #67

## Testing
- `corepack pnpm install --frozen-lockfile --ignore-scripts`
- `corepack pnpm exec tsc --noEmit`
- `corepack pnpm exec eslint 'app/share/[token]/page.tsx'`

### DIFF
--- a/apps/web/app/share/[token]/page.tsx
+++ b/apps/web/app/share/[token]/page.tsx
@@ -61,9 +61,7 @@ interface GuestComment {
   timecode_start?: number | null
 }
 
-interface CommentsResponse {
-  comments: GuestComment[]
-}
+type CommentsResponse = GuestComment[] | { comments?: GuestComment[] }
 
 // ─── Utility ──────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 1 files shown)


ravirajsinh45 (approved):
LGTM — minimal, well-scoped fix that correctly handles both response shapes the API returns. Backend inconsistency at comments.py:557 vs :567 is the real root cause; worth a follow-up to normalize, but the frontend handling here is defensive and correct.

PR #73, opened 2026-05-22
Normalize share comments response shape

### LINKED ISSUE #72: Normalize `/share/{token}/comments` response shape
## Problem

`apps/api/routers/comments.py:544` (`list_share_comments`) returns two different shapes from the same endpoint:

- **Success path** (`comments.py:567`) — bare array:
  ```py
  ```
- **No-asset-id fallback** (`comments.py:557`) — object:
  ```py
  return {"comments": []}
  ```

This forced the frontend in #70 to handle both shapes via a union type. Any future consumer of this endpoint will hit the same trap.

## Proposed fix

Normalize both branches to return a bare array (matches the success path and the format used by `_build_comment_response`):

```py
if not target_asset_id:
    return []
```

Then simplify the frontend in `apps/web/app/share/[token]/page.tsx` to drop the union type:

```ts
type CommentsResponse = GuestComment[]
```

## Notes

- Bare array is the existing success shape, so this normalizes toward the path real consumers already use.
- No other consumers of this endpoint exist in the repo (verified at #70 review time, but worth re-checking before merging).
- Tests should cover both the asset-share (asset_id from link) and folder/project-share (asset_id from query) paths returning the same shape.

Discovered during review of #70 (which fixes the symptom on the frontend).

GChaucer:
## Summary
- Normalize `GET /share/{token}/comments` so the no-target fallback returns a bare array.
- Simplify the single-asset share page comments response type now that the endpoint returns a consistent array shape.
- Add backend regression coverage for asset-share, folder/project-share query asset, and no-target fallback paths.

Closes #72

## Testing
- `. .venv/bin/activate && pytest apps/api/tests/test_share_comments.py`
- `. .venv/bin/activate && pytest apps/api/tests/test_share_comments.py apps/api/tests/test_smoke.py`
- `git diff --check`

## Notes
- Installed API test dependencies in a local virtualenv.
- Frontend lint was not run because `npm ci --ignore-scripts` fails due package-lock/package.json drift, and `pnpm install --frozen-lockfile --ignore-scripts` fails because pnpm-lock.yaml is out of sync with package.json.

### DIFF
--- a/apps/api/routers/comments.py
+++ b/apps/api/routers/comments.py
@@ -554,7 +554,7 @@ def list_share_comments(
     # Determine the asset_id to list comments for
     target_asset_id = link.asset_id or asset_id
     if not target_asset_id:
-        return {"comments": []}
+        return []
     asset_id = target_asset_id
 
     # Get top-level comments — reuse same format as authenticated endpoint
--- a/apps/api/tests/test_share_comments.py
+++ b/apps/api/tests/test_share_comments.py
@@ -1,0 +1,75 @@
+import uuid
+from unittest.mock import MagicMock, patch
+
+
+@patch("apps.api.routers.comments._build_comment_response")
+@patch("apps.api.routers.comments.validate_share_link")
+def test_share_comments_returns_array_for_asset_share(
+    mock_validate,
+    mock_build_comment_response,
+    client,
+    mock_db,
+):
+    asset_id = uuid.uuid4()
+    comment = MagicMock()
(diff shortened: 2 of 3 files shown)


ravirajsinh45 (approved):
LGTM — clean implementation of the normalization from #72. Backend → array, frontend simplified, tests cover all three paths (asset-share, folder-share via ?asset_id, no-target empty). Mock pattern matches the rest of the suite.

Issue #51, opened 2026-04-13
Processed files return 403 - Worker does not set public-read ACL on S3 objects


gimphub-tech:
### Describe the bug

After transcoding completes, the processed HLS files (playlist.m3u8, segments .ts) and thumbnails return 403 Forbidden when accessed directly. The worker uploads processed files to S3 without setting object-level ACL, and most S3-compatible providers (Exoscale, Cloudflare R2, etc.) do not inherit bucket-level ACL on new objects automatically.

### Steps to reproduce

1. Configure S3_STORAGE=minio with a non-AWS S3 provider (tested with Exoscale SOS)
2. Set bucket ACL to public-read in provider dashboard
3. Upload and transcode a video
4. Try to stream the video — 403 on all playlist.m3u8 and .ts segment files

Manually set public-read on all processed objects after each transcode:

```
aws --endpoint-url=https://YOUR_ENDPOINT s3 ls s3://YOUR_BUCKET/processed/ \
  --recursive | awk '{print $4}' | \
  xargs -I {} aws --endpoint-url=https://YOUR_ENDPOINT \
  s3api put-object-acl --bucket YOUR_BUCKET --key {} --acl public-read
``` 

### Expected behavior

Rather than relying on public-read ACL at all, the cleaner and more secure fix would be to keep all S3 objects private and serve them exclusively via presigned URLs — which FreeFrame already generates for playback. The current public-read approach is essentially "security through obscurity" — anyone who obtains a URL can access the file indefinitely, regardless of share link expiry or access revocation. This would:

1. Eliminate the ACL issue entirely across all S3 providers
2. Improve security — objects are never publicly accessible without 
   a valid, time-limited presigned URL
3. Remove the dependency on provider-specific ACL behavior

The current presigned URL implementation already handles auth correctly for the API consumer — the gap is that the HLS player fetches sub-segments (.ts files) directly from S3 without presigned URLs, which requires public-read. A proper fix would proxy HLS segment requests through the API (similar to the existing /hls_proxy endpoint) so all S3 access goes through presigned URLs.

### Area

Backend (FastAPI)

### Version / Commit

latest main

ravirajsinh45:
Thanks a lot for the detailed report and — more importantly — for pushing back on the "just set public-read" workaround and spelling out the proper fix. You nailed both the root cause and the direction.

[...]

Issue #46, opened 2026-04-12
Frontend crashes with "TypeError: Failed to construct 'URL'" when asset thumbnail_url is null during processing


gimphub-tech:
### Describe the bug

After uploading any asset (video, image, or audio), the entire dashboard crashes immediately with the following error:

TypeError: Failed to construct 'URL': Invalid URL
at n (layout-f89cd1e259d41ed4.js:1:32330)
at layout-f89cd1e259d41ed4.js:1:33798

This happens because the layout component tries to construct a URL from thumbnail_url while it is still null (asset is in "processing" state). 
The crash renders the entire dashboard unusable until processing completes and one refreshes the browser.

### Steps to reproduce

1. Log in to FreeFrame
2. Open any project
3. Upload any asset (video, image, or audio)
4. Observe: dashboard immediately shows "Application error" crash screen
5. Wait for processing to complete (varies: seconds for images, minutes for video)
6. Reload the page — dashboard works again

### Expected behavior

The dashboard should remain functional while an asset is processing. 
A loading placeholder or spinner should be shown instead of attempting to render a null thumbnail_url. The upload flow should not be interrupted.

### Screenshots

**During Upload:**

<img width="1346" height="481" alt="Image" src="https://github.com/user-attachments/assets/0928de8b-3374-40db-9df4-6f1535855632" />

**After refreshing the browser:**

<img width="2930" height="1016" alt="Image" src="https://github.com/user-attachments/assets/8709e75d-966f-4f86-9d01-323d7f2698eb" />

### Area

Frontend (Next.js)

### Version / Commit

latest main

ravirajsinh45:
Thanks for the detailed report and the upload-flow repro steps @gimphub-tech.

Quick heads-up on the root cause — turned out to be slightly different from the `thumbnail_url` path you suspected, but the symptom you observed (crash kicked off by the upload) is exactly right.

**Actual cause:** `apps/web/hooks/use-sse.ts` calls `new URL(\`${API_URL}/events/${projectId}\`)` without a base. When `NEXT_PUBLIC_API_URL` is set to a relative path like `/api` (typical for nginx-proxied deployments), `new URL("/api/events/...")` throws `TypeError: Failed to construct 'URL': Invalid URL`. `UploadSSEBridge` opens its first SSE connection the moment any asset starts processing after an upload — which lines up with "crashes immediately after upload" and "reload fixes it once processing is done" (no processing items → no SSE listener → no crash).

[...]

PR #76, opened 2026-06-22
fix: use hls.js for HLS video playback on share page

### LINKED ISSUE #68: Single-asset public share: HLS video fails to play in Firefox (no hls.js fallback)
Reported by @ducknoodledance in #62. Firefox on Linux shows
"No video with supported format and mime type found"; Brave plays.
Project / folder share viewers are unaffected.

## Root cause
The single-asset share page uses a bare `<video>` element:

`apps/web/app/share/[token]/page.tsx:478-485`
```jsx
<video src={streamUrl} controls className="..." preload="metadata">
  Your browser does not support video playback.
</video>
```

Since the HLS proxy fixes (`ad3907ad`, `bfbbcee2`), `streamUrl` for
videos resolves to `…/master.m3u8`. Firefox and Chrome on Linux have
no native HLS support, so the element fails. The folder-share viewer
already handles this correctly with an hls.js dynamic import +
`canPlayType('application/vnd.apple.mpegurl')` Safari fallback
(`apps/web/components/share/folder-share-viewer.tsx:674-686`); the
single-asset share page was never updated.

## Proposed fix
Replace the bare `<video>` on the single-asset share page with the
same HLS-aware player used by the folder share viewer (extract a
shared `<HLSVideo>` component so both paths use it).

rubenxyz:
## Description

The `ShareMediaViewer` component on the public share page uses a plain `<video src={streamUrl}>` element. This works in Safari (which has native HLS support) but fails to play HLS (.m3u8) streams in Firefox and Chrome, as these browsers do not support HLS natively.

### Fix

- Added hls.js integration for browsers that support MediaSource Extensions (Chrome, Firefox, etc.)
- Uses `Hls.isSupported()` to detect MSE-capable browsers, loads HLS via hls.js
- Falls back to native `<video src>` for Safari and direct media files
- Applied the same pattern to both video and audio branches
- hls.js was already a project dependency (^1.6.15)

### Testing

This fix is verified in production at https://www.kamiko.xyz/freeframe/ - HLS video plays correctly in both Firefox and Chrome.

### Related

Fixes #68

### DIFF
--- a/apps/api/tests/test_ffmpeg_transcoder.py
+++ b/apps/api/tests/test_ffmpeg_transcoder.py
@@ -1,0 +1,189 @@
+"""
+Tests for FFmpegTranscoder: command construction and error handling.
+
+Verifies:
+- has_audio=True  → ffmpeg cmd includes -map a:0, var_stream_map has audio tracks
+- has_audio=False → ffmpeg cmd excludes -map a:0, var_stream_map is video-only
+- _run() uses errors='replace' to survive Latin-1/Shift-JIS metadata
+- _run() raises RuntimeError with stderr on non-zero exit
+"""
+import asyncio
+import json
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from packages.transcoder.ffmpeg_transcoder import FFmpegTranscoder
+from packages.transcoder.base import TranscodeJob
+
+
+def _make_job(qualities: list[str] | None = None) -> TranscodeJob:
+    return TranscodeJob(
+        media_id="media-1",
+        version_id="v1",
+        input_s3_key="uploads/video.mp4",
+        output_s3_prefix="hls/media-1/v1",
+        qualities=qualities or ["1080p", "720p", "360p"],
+    )
+
+
+# ─── _run() error handling ────────────────────────────────────────────────────
+
+def test_run_raises_on_nonzero():
+    with patch("subprocess.run") as mock_run:
+        mock_result = MagicMock()
+        mock_result.returncode = 1
+        mock_result.stderr = "ffmpeg error details"
+        mock_run.return_value = mock_result
+
+        with pytest.raises(RuntimeError, match="ffmpeg exited 1: ffmpeg error details"):
+            FFmpegTranscoder._run(["ffmpeg", "-i", "test.mp4"], label="ffmpeg")
+
+
+def test_run_uses_errors_replace():
+    """Verify that text=True + errors='replace' is passed to subprocess.run."""
+    with patch("subprocess.run") as mock_run:
+        mock_result = MagicMock()
+        mock_result.returncode = 0
+        mock_result.stderr = ""
+        mock_run.return_value = mock_result
+
+        FFmpegTranscoder._run(["ffmpeg", "-i", "test.mp4"], timeout=60, label="ffmpeg")
+
+        call_kwargs = mock_run.call_args[1]
+        assert call_kwargs["text"] is True
+        assert call_kwargs["errors"] == "replace"
+        assert call_kwargs["timeout"] == 60
+
+
+# ─── has_audio=True command construction ───────────────────────────────────────
+
+def test_transcode_with_audio_includes_audio_map():
+    """When ffprobe detects audio streams, ffmpeg cmd must include -map a:0."""
+    def mock_run_side_effect(cmd, **_kwargs):
+        # First call: video probe → return metadata
+        if "-select_streams" in cmd and cmd[cmd.index("-select_streams") + 1] == "v:0":
+            mock = MagicMock()
+            mock.returncode = 0
+            mock.stderr = ""
+            mock.stdout = json.dumps({
+                "streams": [{"r_frame_rate": "30/1", "duration": 10.0, "width": 1920, "height": 1080}],
+            })
+            return mock
+        # Second call: audio probe → return audio stream
+        if "-select_streams" in cmd and cmd[cmd.index("-select_streams") + 1] == "a":
+            mock = MagicMock()
+            mock.returncode = 0
+            mock.stderr = ""
+            mock.stdout = json.dumps({
+                "streams": [{"codec_type": "audio"}],
+            })
+            return mock
+        # Third call: main ffmpeg transcode
+        mock = MagicMock()
+        mock.returncode = 0
+        mock.stderr = ""
+        return mock
+
+    with patch("subprocess.run", side_effect=mock_run_side_effect) as mock_run:
+        s3_mock = MagicMock()
+        s3_mock.generate_presigned_url.return_value = "https://s3.example.com/uploads/video.mp4"
+        s3_mock.upload_file = MagicMock()
+
+        # Mock thumbnail generation
+        with patch("builtins.open", MagicMock()), \
+             patch("pathlib.Path.glob", return_value=[]), \
+             patch("pathlib.Path.rglob", return_value=[]), \
+             patch("pathlib.Path.mkdir"), \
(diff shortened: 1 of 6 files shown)


ravirajsinh45 (changes requested):
Thanks for tackling cross-browser HLS on the share page — hls.js is the right call. A couple of blockers mean the player won't actually play in the standard split-origin setup, plus some hardening.

**Blockers**

1. **Relative stream URL isn't prefixed with the API origin.** Backend returns a root-relative URL (`/stream/hls/master.m3u8?token=…`, `apps/api/routers/share.py:1383`). It's passed straight to `hls.loadSource(streamUrl)` / `mediaEl.src`, which resolves against the frontend origin (`:3000`), not `NEXT_PUBLIC_API_URL` (`:8000`) → the manifest 404s and the video never plays. The canonical player already handles this (`apps/web/components/review/video-player.tsx:185-188`: `url.startsWith('/') ? ${API_URL}${url} : url`), same convention as #46/#49 and #51/#57. Resolve to absolute inside the effect before `loadSource`/`src` (`API_URL` is already in scope).
2. **The effect never re-runs after the element mounts.** Deps are `[streamUrl, asset.asset_type]`, but `<video>`/`<audio>` are gated behind `streamLoading`. The parent sets `streamUrl` in `.then` and `streamLoading=false` in `.finally` (separate microtasks, not batched), so by the time the element mounts no dep has changed and hls is never attached. Since `src={streamUrl}` was removed from the element, the result is a blank player. Add `streamLoading` to the dep array (or use a callback ref).

**Should fix**

3. Playback errors are silently swallowed — no `Hls.Events.ERROR` listener and no `onError` on `<video>`/`<audio>`. Presigned-S3 expiry/403 leaves a black player; the existing "unavailable" fallback only fires when `streamUrl` is falsy. Surface a fatal-error state.

**Nice to have**

4. The top-level `import Hls from 'hls.js'` pulls ~530KB into the public share route's first-load JS for every visitor (including image/audio/folder shares). `apps/web/components/share/folder-share-viewer.tsx` lazy-loads via `import('hls.js')` — matching that keeps the page light.
5. The non-Safari-without-MSE path assigns `.m3u8` to native `src` with no `canPlayType` guard (sibling `HlsVideo` guards this).

rubenxyz:

**Blockers fixed:**
1. **Relative stream URL** — stream URL from backend (`/stream/hls/master.m3u8?token=...`) is now prefixed with `API_URL` before `hls.loadSource()`, matching the canonical video-player pattern.
2. **Effect deps** — `streamLoading` added to the dependency array so hls.js attaches after the video element is mounted (prevents race condition with SSR).

**Nice-to-haves done:**
- **Lazy-import hls.js** — dynamic `import('hls.js')` instead of top-level import, reducing initial bundle size.
- **`canPlayType` guard** — checks for native `.m3u8` support before fallback (Safari).
- **Error surface** — `Hls.Events.ERROR` handler for fatal network/media errors + `<video> onError` → surface `fatalError` state with AlertTriangle icon instead of silent failure.

Frontend build passes. Branch updated with all changes.

ravirajsinh45 (changes requested):
Nice iteration, @rubenxyz — the HLS work from round one is all resolved: the stream URL is now prefixed with `API_URL`, `streamLoading` is in the dep array so hls attaches after mount, lazy `import('hls.js')` keeps the share route light, and the fatal-error surface is a real improvement. The backend transcoder change (only emit `-map a:0` / `a:{i}` when ffprobe actually finds an audio stream) is a genuinely good fix for silent videos, and the test coverage is appreciated. 👍

The remaining issues are all in the changes bundled *alongside* the HLS fix — one of them is a crash, so flagging as changes-requested.

**Blocker**

1. **Rules-of-Hooks violation crashes the comment list.** In `GuestCommentList`, `apps/web/app/share/[token]/page.tsx:234` calls `const [imgError, setImgError] = React.useState(false)` inside `comments.map((comment) => { … })`. Because `comments` starts `[]` with an early return on empty (`page.tsx:16,36`), the number of hooks rendered changes between renders — 0 when empty, N once comments load, N+1 after each new comment via `refreshKey`. React throws *"Rendered more hooks than during the previous render"* as soon as comments populate, white-screening the panel. Please extract a small `<GuestCommentItem comment={…} />` subcomponent that owns its own `imgError` state, and map over that.

**Should fix**

2. **`copyToClipboard` reports success even when the write rejects.** `apps/web/lib/utils.ts:588` fires `navigator.clipboard.writeText(text).catch(() => {})` and then `return true` synchronously, so a rejected write (document not focused, permission denied, blocked by policy) still returns `true` → the UI shows "Copied!" for a copy that didn't happen, and the new `'failed'` state is unreachable on the modern-API path. Since reliable copy + a failed state is the point of this refactor, either make the function `async` and await the write, or return the promise result (`.then(() => true, () => false)`).

**Scope**

3. This PR is titled for HLS playback but also ships the guest-comment author/avatar refactor, the `copyToClipboard` utility + two share-dialog rewrites, aria-live accessibility, and the backend audio-handling + tests. Each is reasonable on its own, but bundling them is what let a crash-level comment-list bug ride in under a player fix, and it makes the change hard to review or revert. Could you split these — at minimum pull the guest-comment refactor into its own PR where #1 gets fixed? The HLS + transcoder parts look ready to land on their own.

Thanks again for the careful follow-up here — the player itself is in good shape.

ravirajsinh45:
Heads-up: #77–#80 turned out to be duplicates of this PR (all five had the identical diff), so I've closed them and we'll consolidate here. The two items from my review above — the `GuestCommentList` hooks-in-map crash (blocker) and the `copyToClipboard` false-success — are what to fix. Thanks again, @rubenxyz!

rubenxyz:
Thanks for the thorough review @ravirajsinh45. Both remaining issues are now fixed:

**Blockers resolved:**

**1. `GuestCommentList` hooks-in-map crash** — Extracted `<GuestCommentItem comment={…} />` subcomponent. Each rendered item now owns its own `imgError` state via `React.useState(false)` at the component level. No more hook-count mismatch between empty, populated, and refresh states — the "Rendered more hooks than during the previous render" crash is fixed.

**2. `copyToClipboard` false-success** — Made the function `async` and properly `await`-ed `navigator.clipboard.writeText()`. A rejected write (permission denied, document not focused, etc.) now correctly returns `false` → the "Failed" state becomes reachable. Updated all 3 call sites (`share-dialog` CopyButton, `share-create-dialog` CopyButton, and the inline copy icon) to `await` the result.

Scope is now just the share page + copy utility. The HLS/transcoder parts are unchanged. Ready for re-review.

ravirajsinh45 (approved):
Re-reviewed at the latest commit plus a small maintainer follow-up I pushed. Both round-2 items are resolved:

- **GuestCommentList hooks-in-map crash** — fixed via the extracted `GuestCommentItem`; each item owns its own `imgError` state, so hook count is constant across empty/populated/refresh renders.
- **`copyToClipboard` false-success** — now async, awaits `writeText`, and every call site gates its success UI on the real boolean.

Round-1 HLS work (API_URL-prefixed stream, `streamLoading` dep, lazy import, fatal-error surface, `canPlayType` guard) all holds, and the transcoder no-audio (#78) + stderr (#79) fixes are correct with passing tests.

I pushed a few small hardening nits directly: the HLS effect now catches an hls.js chunk-load failure, tears down `hls` on a native media error so it can't keep fetching into a detached element, and shows an explicit unsupported-browser message; the inline copy button now surfaces a failed state; and the clipboard fallback removes its textarea in a `finally`.

Thanks for the careful iteration, @rubenxyz — approving. 🚀

PR #77, opened 2026-06-22
fix: prevent GuestCommentList crash on logged-in user comments

rubenxyz:
## Description

The `GuestCommentList` component called `comment.guest_name.charAt(0)` in its avatar rendering, but the comments API returns `author: { name, avatar_url }` for logged-in user comments instead of `guest_name`. When any logged-in user comments on a shared asset, `guest_name` is `undefined`, causing a `TypeError: Cannot read properties of undefined (reading 'charAt')` and a client-side crash.

### Fix

- Use `comment.guest_name ?? comment.guest_author?.name ?? comment.author?.name ?? 'Unknown'` as display name
- Also check `comment.author.avatar_url` for logged-in authors, rendering the avatar image when available
- Made `guest_name` and `guest_email` optional in type to reflect API response shape accurately

### DIFF
--- a/apps/api/tests/test_ffmpeg_transcoder.py
+++ b/apps/api/tests/test_ffmpeg_transcoder.py
@@ -1,0 +1,189 @@
+"""
+Tests for FFmpegTranscoder: command construction and error handling.
+
+Verifies:
+- has_audio=True  → ffmpeg cmd includes -map a:0, var_stream_map has audio tracks
+- has_audio=False → ffmpeg cmd excludes -map a:0, var_stream_map is video-only
+- _run() uses errors='replace' to survive Latin-1/Shift-JIS metadata
+- _run() raises RuntimeError with stderr on non-zero exit
+"""
+import asyncio
+import json
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from packages.transcoder.ffmpeg_transcoder import FFmpegTranscoder
+from packages.transcoder.base import TranscodeJob
+
+
+def _make_job(qualities: list[str] | None = None) -> TranscodeJob:
+    return TranscodeJob(
+        media_id="media-1",
+        version_id="v1",
+        input_s3_key="uploads/video.mp4",
+        output_s3_prefix="hls/media-1/v1",
+        qualities=qualities or ["1080p", "720p", "360p"],
+    )
+
+
+# ─── _run() error handling ────────────────────────────────────────────────────
+
+def test_run_raises_on_nonzero():
+    with patch("subprocess.run") as mock_run:
+        mock_result = MagicMock()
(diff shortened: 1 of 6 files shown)


ravirajsinh45 (changes requested):
The `displayName` fallback correctly resolves the `TypeError` — good catch. Two things to address before merge, plus minor polish.

**Security**

- The public, unauthenticated `/share/[token]` page now renders `comment.author.avatar_url` as a raw `` (the `eslint-disable next/next/no-img-element` bypasses Next's host allowlist). `avatar_url` is user-controlled and stored without scheme/host validation (`PATCH /users/{id}` → `apps/api/routers/users.py:91`) and served by the no-auth comments endpoint. A commenter can point it at any origin, so an anonymous viewer's browser fetches that host when the comment renders — a tracking-pixel / IP-leak vector (IP, UA, approximate geo, view timing). It isn't script-XSS and the share token isn't leaked under the default referrer policy, but the privacy leak is real. The sibling `folder-share-viewer.tsx` deliberately renders only the initial — suggest the same here, or validate/proxy the URL (and add `referrerPolicy="no-referrer"`).

**Type contract**

- `guest_name`/`guest_email` aren't fields the API returns — the guest name lives under `guest_author.name`. So `comment.guest_name` (the first operand of the chain) is always `undefined`/dead code, and was actually the original crash's root cause. Recommend dropping those fields from the interface and starting the chain at `guest_author?.name`.

**Minor**

- The new avatar `` has no `onError` fallback to the initial (the branding logo in this same file already uses one) — a broken/expired S3 URL shows a broken-image glyph.
- `displayName` uses `??`, so an empty-string name renders a blank avatar/label instead of "Unknown"; sibling code uses `||` for exactly this.
- The richer comment objects carry `replies`/`reactions`/`annotation` that are dropped here — worth deciding intentionally whether the public panel should show them (at least reply counts).

rubenxyz:

**Security:**
- Avatar `` now has `referrerPolicy="no-referrer"` to prevent tracking-pixel leakage on the public unauthenticated share page.
- Added `onError` handler to fall back to initials when the avatar image fails to load.

**Type contract:**
- Removed `guest_name`/`guest_email` from the `GuestComment` interface. The API returns `guest_author.name` (not `guest_name`), so the fallback chain now starts at `guest_author?.name`.
- Full chain: `guest_author?.name || author?.name || 'Unknown'` (using `||` so empty strings fall through to 'Unknown').

**Minor:**
- Per-comment `imgError` state (via `useState`) so each avatar fails independently to initials rather than hiding all avatars.

Frontend build passes. Branch updated.

ravirajsinh45:

```
git checkout main && git pull
git checkout -b fix/one-thing
```

PR #78, opened 2026-06-22
fix: handle transcode for videos without audio track

rubenxyz:
## Description

Videos without an audio track (e.g. screen recordings, animations, silent clips) cause the transcoder to fail with `Stream map 'a:0' matches no streams` because `-map a:0` was applied unconditionally.

### Fix

- Probe for audio streams using `ffprobe -select_streams a` before building the ffmpeg command
- Only include `-map a:0` and `var_stream_map` audio entries (`v:{i},a:{i}`) when `has_audio=True`
- Videos without audio now produce video-only HLS output

### DIFF
--- a/apps/api/tests/test_ffmpeg_transcoder.py
+++ b/apps/api/tests/test_ffmpeg_transcoder.py
@@ -1,0 +1,189 @@
+"""
+Tests for FFmpegTranscoder: command construction and error handling.
+
+Verifies:
+- has_audio=True  → ffmpeg cmd includes -map a:0, var_stream_map has audio tracks
+- has_audio=False → ffmpeg cmd excludes -map a:0, var_stream_map is video-only
+- _run() uses errors='replace' to survive Latin-1/Shift-JIS metadata
+- _run() raises RuntimeError with stderr on non-zero exit
+"""
+import asyncio
+import json
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from packages.transcoder.ffmpeg_transcoder import FFmpegTranscoder
+from packages.transcoder.base import TranscodeJob
+
+
+def _make_job(qualities: list[str] | None = None) -> TranscodeJob:
+    return TranscodeJob(
+        media_id="media-1",
+        version_id="v1",
+        input_s3_key="uploads/video.mp4",
+        output_s3_prefix="hls/media-1/v1",
+        qualities=qualities or ["1080p", "720p", "360p"],
+    )
+
+
+# ─── _run() error handling ────────────────────────────────────────────────────
+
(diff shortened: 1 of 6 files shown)


ravirajsinh45 (changes requested):
The conditional `-map a:0` / `var_stream_map` logic for audio-less inputs is correct and a genuinely useful fix. A few things before merge:

- **Coordinate with #79 (overlapping change).** This PR and #79 both edit the same region of `packages/transcoder/ffmpeg_transcoder.py` and neither is merged yet. GitHub reports each as individually mergeable because it only checks against `main`, not against each other — whichever lands second will need a rebase.
- **The probe should fail safe.** The new audio probe uses `subprocess.run(...)` + `json.loads(stdout)` unchecked. On a flaky/partial probe (truncated upload, S3 stall, expired presigned URL), ffprobe exits non-zero but still prints `{}`, so `has_audio` becomes `False` and audio is **silently dropped from a file that has it** while the job reports success. Recommend `check=True` (a genuine no-audio file still exits 0 with `{"streams": []}`, so the happy path is unaffected). If #79 lands first, route this through its `_run` helper instead.
- **Tests.** `packages/transcoder/` has no tests, so the new `has_audio=False` branch is unverified. A unit test asserting the constructed argv (no `-map a:0`, `var_stream_map` = `v:0 v:1 v:2`) for both audio/no-audio inputs would lock it in.
- **Minor.** This adds a second full ffprobe round-trip over the presigned URL; a single `-show_streams` call can yield both metadata and audio presence.

rubenxyz:
Thanks for the review @ravirajsinh45. Feedback addressed:

**Probe fail-safe:**
- Audio probe now uses `_run()` (from PR #79) which fail-fasts on non-zero exit with stderr in the error message. No more silent audio-drop on flaky ffprobe.

**Tests:**
- Added 5 unit tests in `apps/api/tests/test_ffmpeg_transcoder.py`:
  - `test_transcode_with_audio_includes_audio_map` — verifies `-map a:0` and `v:{i},a:{i}` in var_stream_map when audio detected
  - `test_transcode_without_audio_excludes_audio_map` — verifies NO `-map a:0` and video-only var_stream_map when no audio
  - `test_run_raises_on_nonzero` — RuntimeError with stderr
  - `test_run_uses_errors_replace` — `errors='replace'` passed to subprocess
  - `test_run_returns_stdout` — stdout returned on success

All 5 tests pass (`python -m pytest apps/api/tests/test_ffmpeg_transcoder.py -v`).

ravirajsinh45:

```
git checkout main && git pull
git checkout -b fix/one-thing
```

PR #79, opened 2026-06-22
fix: include ffmpeg stderr in transcode error messages

rubenxyz:
## Description

All `subprocess.run(call, check=True, capture_output=True)` calls in the transcoder swallowed ffmpeg/ffprobe stderr on failure. Python's `CalledProcessError.__str__` only shows the command string, not the captured stderr, producing useless errors like:

`Command '['ffmpeg', '-y', '-i', 'http://...']' returned non-zero exit status 8`

### Fix

- Added a `_run()` helper method that checks return code and raises `RuntimeError` with stderr included in the message
- Replaced all 5 subprocess.run calls (`get_video_metadata`, `generate_thumbnails`, and 3 in `transcode`) with `_run()`

This is a pure error reporting improvement — no behavioral change.

### DIFF
--- a/apps/api/tests/test_ffmpeg_transcoder.py
+++ b/apps/api/tests/test_ffmpeg_transcoder.py
@@ -1,0 +1,189 @@
+"""
+Tests for FFmpegTranscoder: command construction and error handling.
+
+Verifies:
+- has_audio=True  → ffmpeg cmd includes -map a:0, var_stream_map has audio tracks
+- has_audio=False → ffmpeg cmd excludes -map a:0, var_stream_map is video-only
+- _run() uses errors='replace' to survive Latin-1/Shift-JIS metadata
+- _run() raises RuntimeError with stderr on non-zero exit
+"""
+import asyncio
+import json
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from packages.transcoder.ffmpeg_transcoder import FFmpegTranscoder
+from packages.transcoder.base import TranscodeJob
+
+
+def _make_job(qualities: list[str] | None = None) -> TranscodeJob:
+    return TranscodeJob(
+        media_id="media-1",
+        version_id="v1",
+        input_s3_key="uploads/video.mp4",
+        output_s3_prefix="hls/media-1/v1",
+        qualities=qualities or ["1080p", "720p", "360p"],
+    )
+
+
+# ─── _run() error handling ────────────────────────────────────────────────────
+
+def test_run_raises_on_nonzero():
+    with patch("subprocess.run") as mock_run:
+        mock_result = MagicMock()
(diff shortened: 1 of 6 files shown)


ravirajsinh45 (changes requested):
The `_run` helper and surfacing stderr is a real improvement — the old `CalledProcessError` messages were useless. One regression to fix before merge, plus a couple of smaller points.

**Regression: `text=True` can fail successful transcodes**

`_run` hardcodes `text=True` with no `errors=`. Three of the migrated call sites (`generate_thumbnails`, the HLS `ffmpeg_cmd`, and `thumb_cmd`) previously ran in bytes mode. `subprocess.run` decodes the *entire* stderr with strict UTF-8 before the return code is checked, and ffmpeg echoes the input's container/ID3 metadata (often Latin-1/Shift-JIS) verbatim — so a transcode that exited 0 raises `UnicodeDecodeError`, gets caught by the broad `except`, and the version is marked failed after burning retries. Reproduces in `python:3.11-slim` on a file whose title tag contains byte `0xe9`. Fix: pass `errors='replace'` (or capture bytes and decode only when building the error message).

**Smaller**

- Both ffprobe calls use `-v quiet`, so `result.stderr` is always empty and `_run` raises `"ffprobe exited N: no stderr output"` — which defeats the PR's goal for the ffprobe paths. `-v error` keeps JSON on stdout while emitting real errors to stderr.
- The `transcode()` ffprobe result is unused (dead code), but `_run` now makes a non-zero exit fatal where it was previously ignored. Minor and arguably fine as fail-fast, just noting it since the metadata isn't consumed.
- No transcoder tests cover `_run`'s non-zero branch or the decode behavior, which is why the above could ship unnoticed — a couple of `subprocess.run`-mocked unit tests would help.

rubenxyz:

**REGRESSION fix:**
- `subprocess.run(text=True, ...)` now includes `errors='replace'` in the `_run()` helper. This prevents `UnicodeDecodeError` when ffmpeg echoes Latin-1/Shift-JIS metadata to stderr during successful transcodes. Without this, valid transcodes were incorrectly marked as failed.

**ffprobe verbosity:**
- Changed ffprobe calls from `-v quiet` to `-v error`. ffprobe with `-v quiet` suppresses ALL stderr output (including errors), defeating the purpose of this PR. `-v error` preserves error messages while suppressing info/warning noise.

**Dead code note:**
- The `transcode()` ffprobe metadata result is unused; `_run()` fail-fasts on non-zero exit so no silent bypass risk.

Tests pass (`python -m pytest apps/api/tests/test_ffmpeg_transcoder.py -v`).

ravirajsinh45:

```
git checkout main && git pull
git checkout -b fix/one-thing
```

PR #80, opened 2026-06-22
fix: add clipboard fallback and visual feedback to Copy buttons

rubenxyz:
## Description

Copy buttons in the share dialog and share-create dialog called `navigator.clipboard.writeText()` with empty `catch` blocks. When the call failed (e.g. in non-secure contexts or older browsers), the failure was silently swallowed and `setCopied(true)` never executed - users got no feedback and no copied URL.

### Fix (both dialogs)

- Guard `navigator.clipboard?.writeText` with optional chaining
- Fall back to `document.execCommand('copy')` via a temporary textarea
- Show a red "Failed" state with X icon, green "Copied!" state with Check icon
- Log errors to console for debugging

### Additional (share-dialog)

- Added `CopyIconButton` compact copy component for the Share dropdown rows
- Added a copy button to each existing share link row in the dropdown
- Updated header to explain the row/icon interaction

### DIFF
--- a/apps/api/tests/test_ffmpeg_transcoder.py
+++ b/apps/api/tests/test_ffmpeg_transcoder.py
@@ -1,0 +1,189 @@
+"""
+Tests for FFmpegTranscoder: command construction and error handling.
+
+Verifies:
+- has_audio=True  → ffmpeg cmd includes -map a:0, var_stream_map has audio tracks
+- has_audio=False → ffmpeg cmd excludes -map a:0, var_stream_map is video-only
+- _run() uses errors='replace' to survive Latin-1/Shift-JIS metadata
+- _run() raises RuntimeError with stderr on non-zero exit
+"""
+import asyncio
+import json
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from packages.transcoder.ffmpeg_transcoder import FFmpegTranscoder
+from packages.transcoder.base import TranscodeJob
+
+
+def _make_job(qualities: list[str] | None = None) -> TranscodeJob:
+    return TranscodeJob(
+        media_id="media-1",
+        version_id="v1",
+        input_s3_key="uploads/video.mp4",
+        output_s3_prefix="hls/media-1/v1",
+        qualities=qualities or ["1080p", "720p", "360p"],
+    )
+
+
+# ─── _run() error handling ────────────────────────────────────────────────────
+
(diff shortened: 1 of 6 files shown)


ravirajsinh45 (commented):
Nice improvement — guarding `navigator.clipboard?.writeText`, the execCommand fallback, and the explicit Failed state are all good, and the per-row copy in the dropdown is handy. Approving in principle; a few nits worth a follow-up:

- **The execCommand fallback is effectively a no-op on iOS Safari** — the most common context where the clipboard API is unavailable. `textarea.select()` doesn't create a copyable selection there and `execCommand('copy')` returns false, so the user lands on the new "Failed" state. Setting `contentEditable`/`readOnly=false` + `setSelectionRange(0, len)` before the copy handles iOS.
- **The fallback block is duplicated ~4×** (both `CopyButton`s, `CopyIconButton`, and the inline share-url button; `execCommand` appears 11×). Extracting a `copyToClipboard()` helper into `lib/utils.ts` would de-dupe and keep behavior consistent — notably, the inline share-url button is the one site missing the Failed state, which is a symptom of the copy-paste.
- **a11y:** the icon-only button signals success/failure purely visually (no `aria-live`), and the fallback removes the focused textarea without restoring focus. The dropdown also offers copyable URLs for disabled links with no cue.
- Nothing blocking from my side.

rubenxyz:

**iOS Safari:**
- `execCommand('copy')` now sets `contentEditable=true` + `readOnly=false` + `setSelectionRange()` on the temporary textarea for iOS Safari compatibility. Without this, Safari is a no-op.

**Deduplication:**
- Extracted `copyToClipboard()` helper into `apps/web/lib/utils.ts` — this replaces 4 duplicated inline copy implementations across:
  - `share-dialog.tsx` CopyButton
  - `share-create-dialog.tsx` CopyButton
  - `share-create-dialog.tsx` LinkCreatedPhase inline copy icon
  - Also available for reuse in CopyIconButton and any future copy buttons.

**a11y:**
- Added `aria-live="polite"` `role="status"` region to both CopyButton components, announcing "URL copied to clipboard" / "Copy failed" to screen readers.
- Both buttons now show a "Failed" visual state with XCircle icon when clipboard is unavailable.

Frontend build passes. Branch updated.

ravirajsinh45:

```
git checkout main && git pull
git checkout -b fix/one-thing
```

PR #81, opened 2026-06-23
feat: add support for skipping video transcoding during local QC

### LINKED ISSUE #75: Need option to stop processing Video.
### Problem

We are using this to QC Videos, but always every video go in processing

### Proposed solution

Just add an option in docker yml to work on original videos.

### Alternatives considered

_No response_

sumitrigi:
Introduced a new configuration option `USE_ORIGINAL_VIDEOS` to allow skipping FFmpeg transcoding and serving uploaded videos as-is for local quality control. Updated relevant files to integrate this feature, including environment variables, Docker configuration, and API logic for handling uploads and processing assets.

## Changes

- 

## Testing

- [x] Backend tests pass (`python -m pytest apps/api/tests/ -v`)
- [x] Frontend builds (`pnpm --filter web build`)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -97,3 +97,5 @@ API_WORKERS=4                  # Gunicorn worker processes (prod only)
 TRANSCODING_CONCURRENCY=2     # Parallel video transcoding jobs (CPU intensive)
 EMAIL_CONCURRENCY=2            # Parallel email sending jobs
 TRANSCODER_ENGINE=ffmpeg
+# Skip FFmpeg transcoding and review uploaded videos as-is (local QC / dev)
+USE_ORIGINAL_VIDEOS=false
--- a/apps/api/config.py
+++ b/apps/api/config.py
@@ -39,7 +39,9 @@ class Settings(BaseSettings):
     refresh_token_expire_days: int = 7
     frontend_url: str = "http://localhost:3000"
     transcoder_engine: str = "ffmpeg"
-    
+    # Skip FFmpeg HLS transcoding and serve uploaded videos as-is (useful for local QC)
+    use_original_videos: bool = False
+
     # Worker concurrency settings
     transcoding_concurrency: int = 2  # Number of concurrent video transcoding jobs
     email_concurrency: int = 2  # Number of concurrent email sending jobs
--- a/apps/api/routers/upload.py
+++ b/apps/api/routers/upload.py
@@ -20,6 +20,7 @@ from ..schemas.upload import (
     CompleteUploadRequest, CompleteUploadResponse, AbortUploadRequest,
     ALLOWED_MIME_TYPES, MAX_FILE_SIZE_BYTES, mime_to_asset_type,
 )
+from ..config import settings
 
 router = APIRouter(prefix="/upload", tags=["upload"])
 
@@ -144,6 +145,12 @@ def complete_upload(
 
     # Then complete S3 multipart
     complete_multipart_upload(body.s3_key, body.upload_id, [p.model_dump() for p in body.parts])
+
+    asset = db.query(Asset).filter(Asset.id == body.asset_id, Asset.deleted_at.is_(None)).first()
+    if settings.use_original_videos and asset and asset.asset_type == AssetType.video:
+        version.processing_status = ProcessingStatus.ready
+        db.commit()
+        return CompleteUploadResponse(status="ready", asset_id=body.asset_id, version_id=body.version_id)
 
     version.processing_status = ProcessingStatus.processing
     db.commit()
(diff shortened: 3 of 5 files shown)


siddhant-rigi (approved):

ravirajsinh45 (changes requested):
Thanks for picking this up, @sumitrigi — and thanks @siddhant-rigi for the context on #75. This is a clean, working first cut: I traced it end-to-end and confirmed the happy path holds together. When `USE_ORIGINAL_VIDEOS=true`, `complete_upload` marks the version `ready`, the transcode dispatch is skipped, and the stream endpoint correctly falls back to presigning `s3_key_raw`. The raw object keeps its `video/mp4` Content-Type, so an H.264 MP4 plays inline in the browser. For the stated use case — local QC of H.264 deliverables — it does the job. 👍

Before we merge it into the public repo, though, I'd like to reshape it toward the design we landed on in #75, because as a global switch it has some sharp edges for anyone running a shared instance:

**1. Make it a transcoder engine, not an instance-wide boolean.**
`USE_ORIGINAL_VIDEOS` is read globally, so enabling it skips transcoding for *every* video from *every* org on the deployment. Per the issue, I'd rather express this as the pluggable path we already have — e.g. `TRANSCODER_ENGINE=passthrough` — so it slots into the existing engine selection instead of branching `complete_upload`/`_process_video` on a side flag. That keeps the decision in one place and avoids a one-env-var footgun.

**2. Keep `ffprobe` + a lossless faststart remux in passthrough mode.**
Right now nothing probes the file, so `duration_seconds` / `width` / `height` / `fps` stay `null` and no thumbnail is generated (grid/reel thumbnails go blank). And without `ffmpeg -c copy -movflags +faststart`, originals with the `moov` atom at the end (common in camera/NLE exports) make the browser do large tail/range fetches before it can seek — noticeable on the ~2GB files this is for. The remux is near-instant and lossless (QC is unaffected), so passthrough should still: probe metadata, generate a thumbnail, and faststart-remux. That gets you the QC speed win without losing the metadata/seek behavior the rest of the app expects.

**3. Heads-up on codec compatibility.**
Passthrough is only as playable as the source. H.265/HEVC MP4 — which was listed as one of the QC source formats — generally won't play in Chrome/Firefox, and since we're no longer normalizing to H.264 there's no fallback. Worth documenting this caveat clearly so it's not a surprise; H.264 MP4 will be perfect, HEVC/ProRes/MXF won't.

**4. Wiring + docs.**
The toggle is only added to `docker-compose.dev.yml`. `docker-compose.prod.yml` uses `env_file: .env.prod`, so it'd technically be picked up there if someone hand-added it, but it isn't surfaced for prod and isn't documented anywhere beyond the `.env.example` line. If this stays a dev/QC-only convenience, let's say so explicitly; if it's a real product feature, let's wire and document it for both.

One smaller note: in passthrough the `transcode_complete` SSE event isn't published, so other viewers on the project won't get the live "ready" nudge (the uploader is fine, since the response says `ready`). Minor, but easy to keep consistent once this goes through the engine path.

None of this is a knock on the implementation — the plumbing is correct. It's about making the feature safe to ship to everyone who pulls the repo. If you'd like, I'm happy to pair on the `passthrough` engine shape. Thanks again for the contribution! 🙏

ravirajsinh45:
Hi @sumitrigi — checking in on this, and apologies for how long it's been quiet on our side.

No pressure either way, but I want to be honest about what the review asked for: it wasn't small edits. Reshaping this into a `TRANSCODER_ENGINE=passthrough` engine, plus keeping `ffprobe` metadata, thumbnail generation and a faststart remux, is a real chunk of work. That's a big ask to leave sitting on someone who contributed a working first cut a month ago.

So, two options and both are completely fine:

1. **You still want it** — say the word and it's yours. The offer to pair on the `passthrough` engine shape stands. Heads-up that the branch now conflicts with `main`; happy to rebase it for you.
2. **You'd rather not** — also completely fine. We'll build the `passthrough` engine ourselves, credit you in the CHANGELOG and PR, and close this one out. Nothing lost.

If we don't hear back in a week or so we'll assume option 2 and take it off your plate — not a deadline, just so this doesn't sit open indefinitely.

For what it's worth, the plumbing here was correct and the tracing held up end to end. The reshape is about making it safe for everyone who pulls the repo, not about the quality of what you wrote. Thanks again.

ravirajsinh45:
Closing this out per the note above, it's been about three weeks with no response, and we did say we'd take it off your plate if that happened. Thanks again for the initial work here, @sumitrigi, the transcoding-bypass idea was solid and we'll build on it as the `passthrough` engine. You'll be credited in the CHANGELOG and PR when that lands.

PR #90, opened 2026-07-06
feat: Instance branding — white-label logos, org name, favicon, Powered by FreeFrame

rubenxyz:
## What this does

Adds a full **instance-level branding / white-labeling** feature so self-hosters can replace FreeFrame branding with their own logos, organization name, and favicon. A togglable "Powered by FreeFrame" badge provides subtle attribution.

This was [discussed and welcomed](https://github.com/Techiebutler/freeframe/discussions/81) by the maintainer. A live demo is running at `https://www.kamiko.xyz/freeframe/`.

## How it works

**Single-row DB table** `instance_branding` stores the entire instance brand identity. A default row is auto-seeded on first GET. Admins configure branding via the settings UI — no env vars needed.

### Backend (`~200 LOC`)
- **Model** `InstanceBranding` — 11 columns: org_name, 5 logo S3 keys, primary_color, powered_by_freeframe toggle
- **API router** — `GET /instance/branding` (public), `PUT /instance/branding` (admin), presigned upload endpoints per logo slot, DELETE to reset logos
- **`generate_presigned_put_url()`** — new helper in `s3_service.py` that uses the public endpoint so browser-uploaded logos work regardless of internal S3 host
- **Alembic migration** — `cdcf8e5a6437_add_instance_branding_table`

### Frontend (`~500 LOC`)
- **branding-store.ts** — fetches from API on init, localStorage cache with fallback chain: API → cache → hardcoded defaults
- **BrandingHead** — dynamically sets `<title>`, `<link rel="icon">`, `<link rel="apple-touch-icon">` via `document.head` manipulation
- **PoweredByBadge** — reusable component (FreeFrame icon + text), gated by `poweredByFreeframe` toggle
- **BrandingLogoUpload** — quick upload (one logo applies to all slots) + advanced 6-slot fine-tune, with error states and loading indicators
- **Settings → Branding page** — full rewrite: Quick Upload, Logos & Icons section (6 slots), Greeting text field, Powered toggle, Reset with confirm dialog
- **Sidebar** — shows custom logo (theme-aware light/dark) + `PoweredByBadge` below org name
- **Login page** — custom logo, org name greeting ("Welcome to {Org Name}"), powered badge
- **Share page** — instance branding fallback when no per-project branding is set (PasswordGate, ErrorState, ShareTopBar)
- **Asset viewer** — centered powered badge in top bar
- **Auth layout** — removed static logo (login page handles it dynamically)
- **Login form** — password-first default step, text sizing polish
- **3 SVG logo assets** — shipped as new FreeFrame defaults (`logo-icon.svg`, `logo-icon-dark.svg`, `logo-full.svg`)

## Files changed

**New (11 files):**
- `apps/api/models/instance_branding.py`
- `apps/api/schemas/instance_branding.py`
- `apps/api/routers/instance_branding.py`
- `apps/api/alembic/versions/cdcf8e5a6437_add_instance_branding_table.py`
- `apps/web/components/settings/branding-logo-upload.tsx`
- `apps/web/components/settings/branding-preview.tsx`
- `apps/web/components/shared/branding-head.tsx`
- `apps/web/components/shared/powered-by-badge.tsx`
- `apps/web/public/logo-full.svg`
- `apps/web/public/logo-icon.svg`
- `apps/web/public/logo-icon-dark.svg`

**Modified (12 files):**
- `apps/api/main.py` — register instance_branding router
- `apps/api/models/__init__.py` — add InstanceBranding import
- `apps/api/services/s3_service.py` — add generate_presigned_put_url()
- `apps/web/stores/branding-store.ts` — API fetch + new fields + localStorage partialize
- `apps/web/app/layout.tsx` — add BrandingHead
- `apps/web/components/layout/sidebar.tsx` — custom logo + PoweredByBadge
- `apps/web/app/(auth)/layout.tsx` — remove static logo
- `apps/web/app/(auth)/login/page.tsx` — custom branding header
- `apps/web/app/share/[token]/page.tsx` — instance branding fallback
- `apps/web/app/(dashboard)/projects/[id]/assets/[assetId]/page.tsx` — powered badge in top bar
- `apps/web/app/(dashboard)/settings/branding/page.tsx` — full rewrite
- `apps/web/components/auth/login-form.tsx` — password-first default + text polish

## Testing

- Backend: `python -m pytest apps/api/tests/ -v`
- Frontend: `pnpm --filter web build`
- Manual: Sign up at `https://www.kamiko.xyz/freeframe/`, go to Settings → Branding, upload a logo, verify it appears in sidebar, login page, share page, and browser tab

## Screenshots

Live at `https://www.kamiko.xyz/freeframe/` — branding settings page, sidebar with custom logo + powered badge, login page with custom branding, share page with instance fallback.


ravirajsinh45 (changes requested):
Thank you for this, @rubenxyz — this is a genuinely strong contribution, and instance white-labeling is something we want. The design is right: a **server-backed singleton** that auto-seeds on first boot, replaces the old per-browser localStorage store instead of adding a parallel one, and degrades gracefully. The superadmin authz, the server-generated S3 keys (no path-traversal vector), the idempotent `<link>`/favicon handling, and the CSP-safe inline-SVG "Powered by" badge are all above the bar.

I'd like to request some changes before merge. Almost all of it is **de-staling** (the PR opened before the storage epic + share-player work landed on `main`) plus **finishing the upload path** — not design rework. Here's the full picture.

## 🙌 Keep as-is
- Server-backed singleton with a sensible fallback chain (API → cache → hardcoded defaults) that subsumes the old localStorage store.
- Consistent `is_superadmin` gating on `PUT` / logo-presign / `DELETE`; public `GET` for login/share rendering.
- `branding/{logo_type}/{uuid4}` keys with `logo_type` whitelisted *before* the key is built.
- Public `GET` hardened: `_enrich_branding_response` wraps each presigned-GET in try/except so a missing logo degrades instead of 500-ing.
- `partialize` deliberately not caching the 1h-expiry logo URLs; anchored hex validation on `primary_color`; `BrandingHead`'s tagged, deduped `<link>` management.

## 🔴 Blocking

**1. Alembic migration branches off a stale parent → two heads → `alembic upgrade head` breaks.**
`cdcf8e5a6437_add_instance_branding_table.py` sets `down_revision = '8ca3dffea55f'`, but on current `main` that revision already has a child (`dfcdaa30f89e`, instance settings) and the single head is now `54b1ad156f8f`. Merging as-is creates a second head, and the documented `alembic upgrade head` aborts with *"Multiple head revisions are present"* — a fresh self-host can't start. **Fix:** re-parent to `down_revision = '54b1ad156f8f'` (the tables are disjoint, so it's a DAG fix only) and confirm `alembic heads` shows one head.

**2 + 3. Logo upload is non-functional on the default (MinIO) deployment.** These are the same endpoint, `get_logo_upload_url`:
- It signs the presigned PUT with `s3_service.get_s3_client()`, which uses the **internal** endpoint (`http://minio:9000`) — so the URL handed to the browser points at an unreachable host. Interestingly the PR already ships the correct helper, `s3_service.generate_presigned_put_url()` (public endpoint), but never calls it. *(Heads-up: the same internal-client pattern exists in a few older places — project branding, watermarks, comment attachments — so this isn't something you introduced; but since logo upload is this PR's headline feature and the fix already ships here, let's use it.)*
- It pins a hardcoded per-slot `ContentType` into the SigV4 signature, while both upload flows PUT with the file's own MIME (and the `?content_type=` they send is ignored). S3/MinIO reject the mismatch with 403. QuickUpload reuses one file across 5 slots that each demand a different type, so it can satisfy at most one.

**Fix:** call `generate_presigned_put_url(key, content_type=…)` and either sign with the client's content-type or leave `ContentType` unsigned (the helper already supports `content_type=None`), keeping type validation server-side. Please add **one end-to-end upload test against MinIO** so this can't silently regress.

## 🟠 Should-fix

- **Rebase `sidebar.tsx` surgically — don't take the branch file.** `main`'s sidebar was rewritten by the storage epic to add the storage indicator (`StorageUsage` / `StorageRing`, `useSWR<InstanceSettings>('/instance/settings')`). The PR reflows the whole file off a pre-storage base, which would delete that feature. The regions are disjoint (branding = logo header/top; storage = bottom), so both keep — please graft only the two branding edits (custom-logo `` swap + `PoweredByBadge` under the org name) onto `main`'s version.
- **Rebase `share/[token]/page.tsx` onto `main`, re-apply branding only.** `main` (share-player work) extracted a `GuestCommentItem` and has a robust `ShareMediaViewer` (dynamic `import('hls.js')`, error recovery, URL resolution, native fallback). The PR edits the old inline guest-comment renderer and adds a static `import Hls from 'hls.js'` with no error handling — both regress `main`, and the HLS change is unrelated to branding. Please drop the guest-comment/HLS edits and re-apply only PoweredByBadge + the instance-logo fallback in `PasswordGate` / `ShareTopBar`.
- **"Remove logo" is broken for the two primary logos.** In `branding-logo-upload.tsx`, `handleRemove` derives the API path via chained `.replace()`, which turns `logo_light`→`"light"` and `logo_dark`→`"dark"` — neither accepted by `DELETE /instance/branding/logo/{logo_type}` (silent 400). favicon/apple/login happen to map correctly, masking it. Please use an explicit `slotKey → apiType` lookup mirroring the backend.
- **`PUT` uses `model_dump(exclude_none=True)`, so nullable fields can't be cleared.** "Reset all branding" sends `logo_*: null` / `primary_color: null`, but only `org_name` applies → reset visibly no-ops for logos, and `primary_color` can never be cleared via any API. Use `exclude_unset=True` / `model_fields_set`, or add an explicit reset endpoint.
- **Singleton has no dedup guard.** `_get_or_create_instance_branding` does `first(); if None: add+commit` with a random UUID PK and no unique constraint — and it runs on the *public* GET (which the client fires in parallel on first load), so two concurrent first-hits create two rows; later `.first()` (no ORDER BY) can read a different row than a `PUT` updated. The sibling `instance_settings` router already solved this (fixed sentinel PK + `except IntegrityError: rollback; re-query`, read-only GET, row seeded in the migration) — please mirror it.
- **Brand flash / hydration.** Logo + favicon are only fetched in a client effect with nothing pre-painted, so every load flashes the FreeFrame default in the sidebar, tab title and favicon; and on the SSR login page `orgName` is persisted, so server "FreeFrame" vs client custom name can trip a hydration mismatch. Ideally server-render branding (like the theme inline-script in `app/layout.tsx`) or gate brand-dependent UI on a `loaded` flag; at minimum cache the last-known logo URLs so the swap isn't from the default.
- **`PoweredByBadge` defaults to `text-white`** → invisible on light-theme surfaces (sidebar, login, share password gate). Default to a theme-aware token and let dark contexts override via `className`.
- **Login layout looks doubled-up.** `login/page.tsx` returns its own full-viewport `min-h-screen` centered wrapper, but `(auth)/layout.tsx` still wraps children in a centered `max-w-sm` card — so it renders card-in-card. Pick one owner of the centered chrome.
- **Route writes through the shared `api` helper.** The settings page + upload components hand-roll `fetch` + `Authorization`, so they get no 401 token-refresh (edits silently no-op after the 15-min token expires) and errors are swallowed. Please use `api.put` / `api.delete` and surface failures.
- **Scope creep — please split out the login-default flip.** `login-form.tsx` changes the default step from magic-code (`'email'`) to password (`'classic'`), demoting a headline flow and regressing it for freshly-invited users who may not have a password yet. Please revert that here; happy to consider password-first as its own PR. *(Keep the neutral "Sign in to FreeFrame" → "Magic code sign-in" de-branding — that part's good.)*

## 🧭 Product direction: fold Branding into the Admin dashboard
We'd like Branding to live **inside the Admin dashboard as a third sub-tab** (`Users | Instance settings | Branding`) rather than a standalone `/settings/branding` nav item — for a single-tenant host, branding is instance-global and belongs beside the other instance settings, in one obvious place. Concretely: remove the item from `settings/layout.tsx`, extend the tab union + button array in `settings/admin/page.tsx`, and lift `BrandingPage()` into an exported `BrandingTab` component (mechanical — it's store-driven with no route params).

On the three branding layers: instance branding (new) is the global default, the old localStorage store is correctly subsumed, and per-project `ProjectBranding` stays as a per-share override. The share-page cascade you wrote (`shareName || project.custom_title || orgName`) is the right layering — please just document that precedence (instance = default, project = override) in a code comment + the PR description.

## 🎯 To make white-labeling truly "piece of cake"
- Finish the upload path (blockers above) and kill the FreeFrame flash — a white-label host really doesn't want end-users seeing "FreeFrame" pop on every load.
- **`primary_color` is currently dead** — stored and validated but never applied (no CSS variable, no picker). Either wire it to a runtime CSS variable + add a color picker, or drop the column until it's used.
- Consider whether the 6-slot logo UI is the simplest thing — a single-logo upload with auto-derived favicon/variants is friendlier for the common case, keeping the per-slot panel as an advanced escape hatch.
- **Sweep the remaining hard-coded "FreeFrame" surfaces** so a rebrand is complete. The biggest is **email**, which this PR doesn't touch: `config.py` `mail_from_name = "FreeFrame"`, the magic-code subject (`"Your FreeFrame login code: …"`), and the invite subjects all ship literal "FreeFrame" — so a fully white-labeled instance still emails "FreeFrame," which is the first thing every invited user sees. Also `app/layout.tsx`'s static `metadata.title`, the setup wizard, and `use-page-title.ts`. Wiring the email sender/subjects to the instance `org_name` would round this out.

## Smaller polish (non-blocking)
Orphaned S3 objects on PUT-replace (the DELETE handler already deletes-then-nulls — mirror it on replace); `org_name` needs `min_length=1, max_length=255` (a >255-char value 500s instead of 422); `import re` at module top instead of inline `__import__('re')`; `branding-preview.tsx` is added but never imported (wire it in as a live preview or drop it); the powered-by toggle is a bare `<button>` — reuse the Radix `Switch`; standardize the `NEXT_PUBLIC_API_URL` fallback.

## Suggested order
1. Rebase on `main`.
2. Re-parent the migration; verify a single head.
3. Fix logo upload (helper + content-type) + add a MinIO upload test.
4. Resolve `sidebar.tsx` / `share` conflicts surgically (keep the storage indicator + `GuestCommentItem`/`ShareMediaViewer`).
5. Should-fix batch (remove-slot map, reset semantics, singleton guard, flash, badge color, login layout, `api` helper).
6. Back out the login-default flip + the hls.js change.
7. Move Branding into the Admin sub-tab.
8. Finish/drop `primary_color`; sweep the email + title surfaces; polish.

Really appreciate you volunteering this — the bones are right, and it's mostly de-staling and finishing the upload path. Happy to help on any of the above.

ravirajsinh45:
@rubenxyz — heads-up before you spend time rebasing this one.

Two share-viewer PRs are landing today: #177 (yours) and #191. Both touch `apps/web/app/share/[token]/page.tsx` in the same region this PR does — #90 inserts `useBrandingStore()` immediately after line 838, which is the exact line #177 changes.

**Please hold the rebase until both are in**, otherwise you'll resolve it twice. I'll comment here once they're merged.

One thing to pre-empt, because it'll look alarming: #191 deletes `FolderAssetViewer` (~107 lines) as dead code, and #90 modifies a line inside that block. Git will show it as a large modify/delete conflict. **The resolution is simply to accept the deletion** — the component was never rendered anywhere. It's not as bad as it will look.

ravirajsinh45:
All clear — #177 and #191 are both merged, so you're unblocked to rebase this.

As warned, `FolderAssetViewer` is gone from `app/share/[token]/page.tsx`, so expect a large modify/delete block where this PR edits inside it. **Accept the deletion** — it was never rendered. Your `useBrandingStore()` insertion point also shifted: the line it followed (`:838`) is now the `matchMedia` initializer.

Shout if the rebase turns out worse than that and we'll take it from here.

rubenxyz:
Closing in favor of #203 — rebased on latest main (v1.7.0) with all blockers, should-fixes, and product direction addressed. See the new PR for the refreshed diff.

PR #93, opened 2026-07-06
fix: avatar upload uses internal S3 endpoint — broken with S3_PUBLIC_ENDPOINT

rubenxyz:
## What this fixes

Avatar upload presigned URLs were generated using `get_s3_client()` which uses the internal S3 endpoint (e.g. `http://localhost:9000`). When `S3_PUBLIC_ENDPOINT` is set, the browser cannot reach the internal endpoint, making avatar upload broken.

## Changes

1. **`s3_service.py`** — Adds `generate_presigned_put_url()` helper that uses `_get_presign_client()` (public endpoint) — mirrors the existing `generate_presigned_get_url()`.

2. **`routers/users.py`** — Adds two new endpoints:
   - `POST /{user_id}/avatar-upload` — generates a presigned S3 PUT URL for the browser to upload an avatar image
   - `POST /{user_id}/avatar-confirm` — after upload, stores the S3 key, generates a long-lived presigned GET URL, and updates the user profile

Both endpoints validate the user owns the profile (or is super admin).

## Files changed
- `apps/api/services/s3_service.py` — +13 lines (`generate_presigned_put_url()`)
- `apps/api/routers/users.py` — +51 lines (avatar-upload + avatar-confirm endpoints)

### DIFF
--- a/apps/api/routers/users.py
+++ b/apps/api/routers/users.py
@@ -11,6 +11,7 @@ from ..services.auth_service import hash_password, get_user_by_email
 from ..tasks.email_tasks import send_invite_email
 from ..tasks.celery_app import send_task_safe
 from ..config import settings
+from ..services import s3_service
 
 router = APIRouter(prefix="/users", tags=["users"])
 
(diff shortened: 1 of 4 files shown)


ravirajsinh45:
Hi @rubenxyz — thanks for this fix! An adversarial code review flagged two blocking issues (verified against the branch head), one of them security-critical, so wanted to raise them before merge:

**🔴 Critical — arbitrary S3 object read (IDOR).** `confirm_avatar_upload` accepts a client-supplied `key` and signs a presigned GET for it (`users.py:129` → `140`). The ownership check (`users.py:124`) only verifies the caller owns *their own profile* — not that the key belongs to them. Since all objects live in one bucket and the presign uses the service's full-access credentials, any authenticated user can call `POST /users/{their_own_id}/avatar-confirm` with `{"key": "raw/<any-project>/<version>/original"}` and get a working 7-day download URL for **any private asset**, bypassing the project/asset permission model entirely.

*Suggested fix:* reject keys not under the caller's own prefix before signing — e.g. `if not s3_key.startswith(f"avatars/{user_id}/"): raise HTTPException(400, ...)`. Better still, don't accept a raw key from the client at all — bind it to the key issued by `/avatar-upload` (a signed token or a pending-upload record) and `head_object` to confirm it exists under that prefix.

**🟠 High — avatars break after 7 days.** The presigned URL (`expires_in=604800`, the SigV4 maximum) is persisted into `user.avatar_url` (`users.py:141`) and served as-is everywhere (`UserResponse`, comment authors), with no read-time re-signing. About 7 days after upload the signature expires and S3/MinIO returns 403 for every avatar render, until a manual re-upload.

*Suggested fix:* store the durable S3 key and generate the presigned GET URL at read/serialization time (mirroring how `/assets/{id}/stream` resolves URLs on demand), rather than persisting a time-limited signed URL.

Happy to help with either fix if useful.

ravirajsinh45 (changes requested):
Requesting changes: this PR introduces a **critical IDOR** — `confirm_avatar_upload` signs a presigned GET for a client-supplied `key`, letting any authenticated user read any private object in the bucket — plus a high-severity issue where the persisted 7-day presigned avatar URL 403s after expiry. Details and suggested fixes in the review comment above. The IDOR should be resolved before merge.

ravirajsinh45 (approved):
Approving. Both blockers are resolved — and apologies for the silence: you pushed this fix on 2026-07-09, a day after my review, and it sat for twelve days. That's on us.

**The IDOR is closed.** `confirm_avatar_upload` now constrains the client-supplied key to the caller's own prefix:

```python
if not s3_key.startswith(f"avatars/{user_id}/"):
    raise HTTPException(status_code=400, detail="Invalid key for this user")
```

combined with the existing `current_user.id != user_id` check, so a presigned GET can no longer be minted for an arbitrary object in the bucket.

**The expiring-URL problem is gone.** `avatar_url` now persists the S3 *key* (`user.avatar_url = s3_key` — the only assignment in the diff), and the `field_validator` on `UserResponse`/`AuthorInfo` mints a fresh presigned GET at serialization time. That's the right shape: nothing durable ever holds a signed URL.

One thing worth knowing, since it nearly bit this PR: `generate_presigned_put_url` sets `params["ContentType"]`, and a **SigV2** presigned PUT fails with `403 SignatureDoesNotMatch` the moment a Content-Type is involved — SigV2's string-to-sign includes it and boto3 presigns it empty. On MinIO this endpoint would have 403'd on every upload. #180 and #197 landed today and pin SigV4 across all modes, so it works now. Timing in your favour.

Two non-blocking notes for whenever you're next in here:

- `confirm_avatar_upload` takes `body: dict` — a small Pydantic schema would give you validation and OpenAPI docs for free.
- The `avatar_url` validator presigns per serialization, so a comment list with N authors does N signings. It's local crypto with no network round-trip so it's cheap, but if author avatars ever show up in a hot list endpoint it's worth a cache.

All four checks green.

PR #94, opened 2026-07-06
feat: add PATCH /auth/change-password endpoint

rubenxyz:
## What this adds

`PATCH /auth/change-password` endpoint so authenticated users can change their own password. The frontend was already coded to call this endpoint but it did not exist.

## Changes

1. **`schemas/auth.py`** — Adds `ChangePasswordRequest` schema with `current_password` and `new_password` fields.

2. **`routers/auth.py`** — Adds `PATCH /auth/change-password` endpoint. Validates the current password, then hashes and stores the new one. Returns an error if the current password is incorrect.

## Files changed
- `apps/api/schemas/auth.py` — +4 lines
- `apps/api/routers/auth.py` — +15 lines

### DIFF
--- a/apps/api/routers/auth.py
+++ b/apps/api/routers/auth.py
@@ -10,6 +10,7 @@ from ..schemas.auth import (
     SendMagicCodeRequest, SendMagicCodeResponse,
     VerifyMagicCodeRequest, SetPasswordRequest,
     AcceptInviteRequest, InviteInfoResponse,
+    ChangePasswordRequest,
 )
 from ..services.auth_service import (
     hash_password, verify_password,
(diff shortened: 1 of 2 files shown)


ravirajsinh45:
Hi @rubenxyz — thanks for adding this! An adversarial code review flagged one edge case worth handling before merge (verified against the branch head):

**🟡 Medium — HTTP 500 for password-less (magic-code) users.** A user who signed in via the magic-code flow and never called `/auth/set-password` has `password_hash = None` but is fully authenticated. When they submit the change-password form, `change_password` calls `verify_password(body.current_password, None)` → `None.encode('utf-8')` raises `AttributeError`, which isn't caught by the surrounding `except ValueError`, so FastAPI returns **500** (opaque server error) instead of a clean 4xx. The existing `/auth/login` endpoint already guards this exact case with `or not user.password_hash`.

*Suggested fix:* guard before verifying — `if current_user.password_hash is None: raise HTTPException(status_code=400, detail="No password set for this account; use set-password instead")`.

ravirajsinh45 (changes requested):
Requesting changes: `change_password` returns HTTP 500 for magic-code users who never set a password (`verify_password(current, None)` → uncaught `AttributeError`). A `password_hash is None` guard (as `/auth/login` already does) resolves it. Details in the review comment above.

ravirajsinh45 (approved):
Approving — the change request is resolved. ✅

Commit `1fc2317` adds the `password_hash is None` guard **before** `verify_password`, so magic-code users who never set a password now get a clear `400` ("use set-password instead") instead of the previous `500`. Verified the rest is sound too: `hash_password`/`verify_password` both truncate to bcrypt's 72-byte limit (no error on long input), the endpoint is auth-gated via `get_current_user`, and the frontend contract matches (`current_password`/`new_password`, response body ignored).

Two non-blocking hardening follow-ups tracked in #149 (backend `new_password` length validation, and session/refresh-token invalidation on password change) — neither needs to hold this PR.

Issue #64, opened 2026-05-21
UI shows hardcoded "/ 10 GB" project storage cap; backend hard-rejects files > 10 GB


ravirajsinh45:
Reported by @ducknoodledance in #61.

## Problem
The project detail page shows a "Storage X / 10 GB" indicator, and the
backend rejects any single upload larger than 10 GB. Both limits are
hardcoded and there is no per-org / per-project quota field in the DB,
so a self-hoster running on their own S3 / MinIO has no way to change
or remove them. The UI text in particular is misleading — it implies a
quota that doesn't actually exist as a configurable concept.

## Where it lives
- Frontend denominator (cosmetic): `apps/web/app/(dashboard)/projects/[id]/page.tsx:592`
  ```ts
  const limit = 10 * 1024 * 1024 * 1024; // 10 GB default limit
  ```
  rendered as `Storage {used} / {limit}` (lines 600–606).
- Backend per-file ceiling: `apps/api/schemas/upload.py:15`
  ```py
  MAX_FILE_SIZE_BYTES = 10 * 1024 * 1024 * 1024  # 10 GB
  ```
  enforced in `apps/api/routers/upload.py` and `apps/api/routers/assets.py`
  ("File exceeds 10GB limit").
- `Project` / `Organization` models have no `storage_limit` column and
  `ProjectResponse` exposes only `storage_bytes` (used), not a limit.

## Proposed fix
- Drop the hardcoded denominator from the UI, or drive it from an
  optional `storage_limit` field that's only shown when set.
- Make `MAX_FILE_SIZE_BYTES` configurable via env (e.g.
  `MAX_UPLOAD_BYTES`, default unlimited or a high sane value).
- Optionally add a per-org/per-project quota for operators who *do*
  want a cap.

PR #99, opened 2026-07-06
fix: widen file_size_bytes to BigInteger (int4 -> bigint)

ravirajsinh45:
## Summary

`MediaFile.file_size_bytes` and `CommentAttachment.file_size_bytes` were `INTEGER`
(int4, ~2.1 GB ceiling) — so any file larger than ~2.1 GB could not be recorded at
all. This undermines #64 (per-file uploads are now nominally unlimited) and #98
(instance storage accounting sums these values). Widen both to `BigInteger`.

**Stacked on #98** (`feature/instance-settings-storage-cap`) — this PR's diff is only
the column change, and its migration chains off #98's. Retarget to `main` once #98 merges.

## Changes

- `MediaFile.file_size_bytes` and `CommentAttachment.file_size_bytes` → `BigInteger`.
- Alembic migration `54b1ad156f8f` (`ALTER COLUMN … TYPE BIGINT`), chained off `dfcdaa30f89e`.
- Structural tests (`apps/api/tests/test_filesize_bigint.py`).

## Testing

- Full backend suite: **88 passed**.
- Migration applied to the dev DB; verified both columns are now `BIGINT` via the
  SQLAlchemy inspector (a wrong type would have shown here).


ravirajsinh45:
Superseded by #104 — this PR was auto-closed when #98's base branch (`feature/instance-settings-storage-cap`) was deleted on merge. The branch was rebased onto `main` and reopened as #104 with the same change.

Issue #65, opened 2026-05-21
No garbage collection for soft-deleted assets / orphaned S3 objects / expired share links


ravirajsinh45:
Reported by @ducknoodledance in #63 (with a working community
workaround script).

## Problem
Every entity uses soft-delete (`deleted_at`) and the codebase intentionally
"never hard-deletes in application code" (CLAUDE.md). But there is no
scheduled job that ever reclaims the storage, so:

- Soft-deleted rows pile up in Postgres indefinitely.
- The underlying S3 / MinIO objects (`s3_key_raw`, `s3_key_processed`,
  `s3_key_thumbnail`) are never removed, even after the parent asset,
  version, folder, or project is soft-deleted.
- Orphans created by interrupted uploads / failed transcodes are not
  swept.
- `ShareLink.expires_at` is defined but never enforced or cleaned.

For a long-lived self-hosted install this means the bucket and DB
grow unbounded.

## Evidence
- All delete endpoints only flip `deleted_at` — e.g. `apps/api/routers/assets.py:197`,
  `apps/api/routers/projects.py`, `apps/api/routers/folders.py`,
  `apps/api/routers/share.py:440`.
- The only place that actually calls `s3.delete_object` is comment
  attachments in `apps/api/routers/comments.py`.
- Celery beat schedule in `apps/api/tasks/celery_app.py:59-64` only
  runs `send_due_date_reminders` — no cleanup task is registered.
- `MediaFile` has no `deleted_at`; it's only reachable through the
  parent version, so once a version is soft-deleted those S3 keys
  become invisible to the app but stay in the bucket.

## Proposed fix
- Add a Celery beat task (`cleanup_soft_deleted`, e.g. daily) that:
  - hard-deletes rows soft-deleted more than N days ago (configurable
    retention window, default e.g. 30 days)
  - deletes their S3 objects (`raw`, `processed`/HLS prefix, thumbnail)
  - cascades through versions, media files, comments, annotations,
    approvals, share links
- Add an orphan sweeper that lists S3 prefixes and removes keys with
  no matching DB row.
- Honor `ShareLink.expires_at` (reject in the API and clean up
  expired rows).
- Document the retention window + a manual "purge now" admin endpoint.

PR #106, opened 2026-07-07
feat: retention-window garbage collection — soft-delete GC + expired-share sweep + purge endpoint (#65 core)

ravirajsinh45:
## Summary

A retention-window garbage collector that reclaims storage from soft-deleted data — the core of #65, building on the upload-orphan reaper (#101). A daily job hard-deletes rows soft-deleted longer than `SOFT_DELETE_RETENTION_DAYS` (default `30`, `0` disables) and deletes their S3 objects, cascading the full `Project → Folder → Asset → Version → MediaFile / Comment / Approval / ShareLink` tree; long-expired share links are swept into soft-delete first; and a superadmin `POST /admin/purge` triggers a run on demand. Backend only, **no DB migration** (reuses existing `deleted_at`).

## Changes

- **Cascade GC** in `tasks/cleanup_tasks.py`: `_purge_project/_folder/_asset/_version/_comment/_share_link` walk the tree **child-first** (there are no ORM relationships / `ondelete=CASCADE`, so ordering is explicit), reclaiming S3 (`delete_object` / `delete_prefix`) best-effort via `_safe`, deleting rows via bulk `.delete(synchronize_session=False)`.
- **`_purge_soft_deleted`** processes aged soft-deleted roots top-down; **`_expire_share_links`** soft-deletes links expired *beyond* the retention window (recently-expired stay owner-editable; 410 read-time enforcement already existed and is now regression-tested).
- **`SOFT_DELETE_RETENTION_DAYS`** config (default 30, `<=0` disables — same convention as the reaper), documented in `.env.example`.
- **Daily `cleanup_soft_deleted` Celery beat task** (03:00 UTC) = expire-sweep + purge; the sole commit boundary is the task wrapper (helpers only `flush()`).
- **`POST /admin/purge`** (superadmin-only): enqueues the same task and returns **`202`** (async — never blocks the request on a long cascade); reclaimed counts are logged by the worker.
- **Tests**: real-Postgres cascade coverage (full tree, retention-window boundary, standalone reaper-version-on-live-asset, disabled path), share-link expiry sweep + 410 regression lock, admin-purge 403/202, and a **metadata-driven FK-completeness guard** that fails CI if a future schema adds an inbound FK the cascade doesn't handle.

**Deferred (tracked follow-ups, not in this PR):** the S3 **orphan sweeper** (list bucket keys, delete those with no DB row — the riskiest slice; deserves a report-only-first posture and its own review) and `CollectionShare.expires_at` (no consumption endpoint exists).

**Verified invariants** (independently checked in review): never deletes live data (descends only from aged soft-deleted roots); FK-safe against real Postgres (every inbound FK covered — now guarded by a test); `0` disables *before* any cutoff is computed; S3 reclaimed before the row; storage-usage accounting is unchanged (committed-sum already excludes soft-deleted).

Part of #65.

## Testing

- [x] Backend tests pass — 19 new GC tests pass; **zero new regressions**. (The suite also shows ~26 **pre-existing** `SetupGuardMiddleware` 503 failures on unrelated auth-gated `client` tests — present on `main`; CI's gate is a `>=40 passed` floor. The new GC tests are DB-state-independent / bypass the guard.)
- [ ] Frontend builds — N/A (backend only)
- [ ] Tested manually in browser — N/A (backend only)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -98,3 +98,4 @@ TRANSCODING_CONCURRENCY=2     # Parallel video transcoding jobs (CPU intensive)
 EMAIL_CONCURRENCY=2            # Parallel email sending jobs
 TRANSCODER_ENGINE=ffmpeg
 MAX_UPLOAD_BYTES=0             # Max size (bytes) for one uploaded file; 0 = unlimited
+SOFT_DELETE_RETENTION_DAYS=30  # Hard-delete rows soft-deleted longer than N days (cascade + S3 reclaim); 0 = disabled
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - **Retention-window garbage collection** ([#65](https://github.com/Techiebutler/freeframe/issues/65)) — a daily `cleanup_soft_deleted` job hard-deletes rows soft-deleted longer than `SOFT_DELETE_RETENTION_DAYS` (default 30, `0` disables) and reclaims their S3 objects, cascading through projects, folders, assets, versions, media, comments, approvals, and share links. Long-expired share links are swept into soft-delete first. No effect unless you run `celery beat`.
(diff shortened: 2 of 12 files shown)


ravirajsinh45:

PR #111, opened 2026-07-07
fix(gc): guard retention purge against restore race + overlapping runs (#107)

### LINKED ISSUE #107: Retention purge races a restore → hard-deletes the just-restored entity (data loss)
Found by an adversarial multi-agent review of #106 (part of #65).

### Mechanism
The retention purge's root `SELECT` (`_purge_soft_deleted` in `apps/api/tasks/cleanup_tasks.py`) takes **no row locks**, and the DB engine has no isolation override (PostgreSQL default **READ COMMITTED**). A `POST /assets/{id}/restore` or `POST /folders/{id}/restore` can commit **between** the root query and the per-root cascade. The cascade re-fetches the root by id and its guard is only `if x is None: return` — it **never re-checks `deleted_at`**. So when it reaches a root that was just restored, the row exists (guard passes) and the cascade **hard-deletes the now-live entity + its S3 objects**.

### Impact
Irrecoverable data loss (no soft-delete tombstone left). Window is narrow — the restore must land during a daily purge run, on an item just past the retention cutoff — but it's real and reachable with a single Celery worker (restore is an independent HTTP request; `POST /admin/purge` can also fire the task).

### Proposed mitigation (pick / combine)
- (a) Re-verify each root still satisfies `deleted_at IS NOT NULL AND deleted_at < cutoff` at purge time, ideally under `SELECT … FOR UPDATE` so a concurrent restore blocks.
- (b) Block `restore` on items already past the retention cutoff (they're condemned anyway).
- (c) A Postgres advisory lock so the beat task and a manual `/admin/purge` can't overlap.

Planned as a stacked PR on top of #106 (bottom of the stack).

ravirajsinh45:
## Summary

Fixes #107 (found by the adversarial review of #106). Closes the purge-vs-restore data-loss race and prevents overlapping purge runs. **Stacked on #106** (base branch `feat/retention-window-gc`) — merge #106 first.

## Changes

- **`_lock_if_still_purgeable(db, model, id, cutoff)`** — re-fetches each purge root under `SELECT … FOR UPDATE`, re-checking `deleted_at IS NOT NULL AND deleted_at < cutoff`. `_purge_soft_deleted` now scans candidate root ids, then per root re-checks + locks and **skips** if it was restored / is no longer aged / already gone — so a `POST /assets|folders/{id}/restore` landing mid-run can never have its now-live entity hard-deleted (applied to all six cascade roots). Child-cascade helpers are unchanged (they must delete children regardless of their own `deleted_at` for FK-safety).
- **`_run_cleanup`** — `pg_try_advisory_xact_lock` (non-blocking, transaction-scoped): the daily beat and a manual `/admin/purge` enqueue can no longer run concurrently; a contended run skips with a log line.
- **`admin.py`** — docstring note that `202` means *enqueued* (a run may be skipped if one is already in progress).

## Testing

- [x] Backend tests pass — real-Postgres tests: `_lock_if_still_purgeable` rejects a restored (`deleted_at` cleared) and a too-recent root and accepts an aged one; a second DB connection holding the advisory lock makes `_run_cleanup` skip (the aged project survives). All existing GC tests still pass.
- [ ] Frontend builds — N/A (backend only)
- [ ] Tested manually in browser — N/A (backend only)

Closes #107.

### DIFF
--- a/apps/api/routers/admin.py
+++ b/apps/api/routers/admin.py
@@ -120,7 +120,9 @@ def purge_now(current_user: User = Depends(require_admin)):
 
     Superadmin only. Enqueues the same `cleanup_soft_deleted` task the daily beat runs, so the
     request returns immediately instead of blocking on a potentially long cascade + S3 deletes.
-    Reclaimed counts are logged by the worker.
+    Reclaimed counts are logged by the worker. If a purge is already running (e.g. the daily beat),
+    the advisory lock serializes purges and this enqueued run is skipped rather than double-cascading,
+    so a 202 here means "enqueued", not "a fresh run happened".
     """
     send_task_safe(cleanup_soft_deleted)
     return PurgeStartResponse(
--- a/apps/api/tasks/cleanup_tasks.py
+++ b/apps/api/tasks/cleanup_tasks.py
@@ -2,6 +2,8 @@ import logging
 from dataclasses import dataclass, asdict
 from datetime import datetime, timezone, timedelta
 
+from sqlalchemy import text
+
 from .celery_app import celery_app
 from ..database import SessionLocal
 from ..config import settings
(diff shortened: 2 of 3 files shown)


ravirajsinh45:

PR #112, opened 2026-07-07
test(ci): SetupGuard bypass fixture so client tests pass in CI + zero-failure gate (#108)

### LINKED ISSUE #108: Auth-gated client tests 503 in CI (SetupGuard) — centralize the setup-guard bypass
`SetupGuardMiddleware` (`apps/api/middleware/setup_guard.py`) returns **503** for every non-exempt route until a superadmin exists in the **real** DB — it queries via its own `SessionLocal`, so it ignores the test `mock_db` override. CI runs a fresh DB with no superadmin, so a chunk of `client`-based tests (~26: e.g. `test_instance_settings.py`, `test_auth.py`, `test_projects.py`, `test_assets_stream_url.py`) **fail in CI**. This is masked only because the CI gate is a `>= 40 passed` **floor**, not zero-failure.

#106 works around it per-test via `monkeypatch.setattr("apps.api.middleware.setup_guard._setup_complete", True)` (see `test_admin_purge.py`, `test_upload_folder_validation.py`).

### Scope
- Add an autouse `conftest.py` fixture that sets `_setup_complete=True` for auth-gated tests (excluding any test that specifically asserts the `needs_setup`/503 behavior), so the client suite actually passes in CI.
- Then consider tightening the CI gate in `.github/workflows/ci.yml` from a `>=40` floor to **zero-failure**, so real regressions can't hide behind the floor.

Planned as a stacked PR on top of #106.

ravirajsinh45:
## Summary

Fixes #108. `SetupGuardMiddleware` 503s every auth-gated route until a superadmin exists in the real DB — so on a fresh CI DB the ~26 `client`-based tests were failing, masked only by the CI gate being a `>= 40 passed` floor. This centralizes the bypass and turns the gate into a real zero-failure check. **Stacked on #107 → #106.**

## Changes

- **Autouse `conftest.py` fixture** `_bypass_setup_guard` sets `apps.api.middleware.setup_guard._setup_complete = True` per test (auto-restored). No test asserts the `needs_setup`/503 path (verified — no `test_setup.py`, no such assertions), so a blanket bypass is safe.
- Removed the now-redundant per-test `_setup_complete` monkeypatches in `test_admin_purge.py` / `test_upload_folder_validation.py` (and the newly-unused `monkeypatch` param on one test).
- **`.github/workflows/ci.yml`**: added `set -o pipefail` before the `pytest | tee` pipeline so **any** test failure now fails the job (zero-failure gate); kept the `>= 40 passed` floor as a secondary "tests deleted/gutted" guard.

## Testing

- [x] Backend tests pass — full suite now **123 passed, 0 failed** against real Postgres (was ~89 passed / 26 SetupGuard-503 failed before this change).
- [ ] Frontend builds — N/A
- [ ] Tested manually in browser — N/A

Closes #108.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,8 +89,10 @@ jobs:
 
       - name: Run tests
         run: |
+          set -o pipefail
           python -m pytest apps/api/tests/ -v --tb=short 2>&1 | tee test-output.txt
-          # Extract passed count and enforce minimum
+          # pipefail above makes this step fail on ANY test failure (zero-failure gate).
+          # The floor below additionally guards against tests being deleted or gutted.
           PASSED=$(grep -oP '\d+ passed' test-output.txt | grep -oP '\d+')
           echo "Tests passed: $PASSED"
           if [ -z "$PASSED" ] || [ "$PASSED" -lt 40 ]; then
--- a/apps/api/tests/conftest.py
+++ b/apps/api/tests/conftest.py
@@ -124,6 +124,16 @@ def auth_headers(client, mock_db, test_user):
     app.dependency_overrides.pop(get_current_user, None)
 
 
+@pytest.fixture(autouse=True)
+def _bypass_setup_guard(monkeypatch):
+    """Mark initial setup complete so SetupGuardMiddleware doesn't 503 auth-gated client tests.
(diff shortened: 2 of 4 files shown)


ravirajsinh45:

PR #113, opened 2026-07-07
feat(gc): report-only-first S3 orphan sweeper (#110)

### LINKED ISSUE #110: GC: S3 orphan sweeper — delete bucket keys with no matching DB row
Follow-up to the retention-window GC (#106, part of #65). That PR reclaims S3 for **soft-deleted** entities via the DB cascade, but does **not** sweep truly orphaned bucket objects — keys with no matching DB row (e.g. left by interrupted flows, or a reaper that deleted a `MediaFile`'s S3 objects but left the row).

### Scope
A periodic sweeper that lists bucket prefixes (`raw/`, `processed/`, `posters/`, branding logos) and deletes keys with no live DB reference.

### Why it's separate from #106
This is the only GC slice that can delete **live** data on a bug (a key it wrongly decides is orphaned). It needs:
- a new `list_keys(prefix)` S3 helper (paginated) in `services/s3_service.py`,
- a **report-only-first** posture: log what it *would* delete behind a flag before any active deletion,
- its own careful review.

Planned as a stacked PR on top of #106.

ravirajsinh45:
## Summary

Closes #110 (the S3 orphan sweeper deferred from #65). Reclaims bucket keys under `raw/` and `processed/` that no `MediaFile` row owns. **Top of the stack: base `chore/setup-guard-test-fixture` → #108 → #107 → #106** — merge those first.

Built **report-only-first**, because the one way this hurts is deleting a live object it wrongly thinks is orphaned:

- **Off by default** — `ORPHAN_SWEEP_GRACE_HOURS=0` disables it. When > 0, only keys whose S3 `LastModified` is older than the window are considered, so in-flight / just-committed uploads are never flagged.
- **Report-only by default** — `ORPHAN_SWEEP_DELETE=false` logs what it *would* delete; deletion happens only when an operator explicitly sets it `true`.

## Changes

- **`services/s3_service.py`** — `list_keys(prefix)`: paginated `list_objects_v2` yielding `(key, last_modified, size)`.
- **`tasks/cleanup_tasks.py`** — `_sweep_orphan_s3(db)` + a weekly `sweep_orphan_s3` beat task. Live-key logic (read-only on the DB): a key is live if it equals a `MediaFile.s3_key_raw`/`s3_key_thumbnail`, or lives under `processed/{project_id}/{asset_id}/{version_id}/` (the transcode `output_prefix`, under which *all* processed outputs — HLS incl. segments, mp3, webp, thumbnails, waveforms — are written), derived from the raw key. `MediaFile` is queried without any live/deleted filter, so soft-deleted-but-not-yet-purged media (owned by the retention GC) is never treated as an orphan.
- **`config.py` / `.env.example`** — the two flags above, documented.
- **`tasks/celery_app.py`** — `sweep-orphan-s3` beat entry (weekly, Sun 04:00 UTC).

## Testing

- [x] Backend tests pass — real-Postgres + faked `list_keys`: a live raw key, a live HLS segment reached only via the prefix match, and a live thumbnail are **never** flagged; only a genuine no-DB-row key is reported; a within-grace key is skipped; `delete=false` deletes nothing; `delete=true` deletes **only** the orphan; `grace_hours=0` disables (never lists the bucket). Full suite 126 passed.
- [ ] Frontend / manual — N/A (backend only)

Closes #110.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -99,3 +99,5 @@ EMAIL_CONCURRENCY=2            # Parallel email sending jobs
 TRANSCODER_ENGINE=ffmpeg
 MAX_UPLOAD_BYTES=0             # Max size (bytes) for one uploaded file; 0 = unlimited
 SOFT_DELETE_RETENTION_DAYS=30  # Hard-delete rows soft-deleted longer than N days (cascade + S3 reclaim); 0 = disabled
+ORPHAN_SWEEP_GRACE_HOURS=0    # S3 orphan sweeper: reclaim raw/+processed/ keys with no DB row; 0 = disabled; >0 = only keys older than N hours
+ORPHAN_SWEEP_DELETE=false     # false = report-only (log only); true = actually delete orphans
--- a/apps/api/config.py
+++ b/apps/api/config.py
@@ -51,6 +51,13 @@ class Settings(BaseSettings):
     # S3 objects reclaimed. Days. 0 (or negative) DISABLES the sweep (matches the reaper convention).
     soft_delete_retention_days: int = 30
 
+    # Orphan S3 sweeper (issue #65 follow-up): reclaim bucket keys under raw/ + processed/ that no
+    # MediaFile row owns. 0 = disabled. When > 0, only keys whose S3 LastModified is older than this
+    # many hours are considered, so in-flight / just-committed uploads are never mistaken for orphans.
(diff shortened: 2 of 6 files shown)


ravirajsinh45:

Issue #37, opened 2026-04-09
Support for Full NVIDIA Hardware Acceleration (NVENC/CUDA)


deezid:
### Describe the bug

Currently, processing high-resolution masters (4K/6K) puts heavy load on the CPU. While the system may be using some hardware encoding, it appears to be a "half-CPU" implementation. For professional studio workflows using NVIDIA cards (like my 1080 Ti), we need full hardware offloading.

### Steps to reproduce

Upload a video clip
Wait forever for it to finish since decoding and encoding runs on my CPU.

### Expected behavior

To fully utilize NVIDIA GPUs, the FFmpeg command needs to include the hardware accelerator flag before the input and the NVENC codec for the output.

Example:
ffmpeg -hwaccel cuda -hwaccel_output_format cuda -i input_master_6K.mov \
  -vf "scale_cuda=1920:1080" \
  -c:v h264_nvenc -preset p4 -tune hq -b:v 8M \
  -c:a aac -b:a 192k \
  output_preview_1080p.mp4

Requirements:
- Ability to use NVIDIA hardware decoder, splitter and scaler.
- Ability to toggle NVIDIA Hardware Acceleration in the settings/env.
- Mapping of /dev/nvidia* devices into the worker container.
- Support for 10-bit HEVC decoding via CUDA to prevent CPU bottlenecks on modern cinema camera footage.

Workaround:
Because Freeframe does not natively support NVENC or CUDA, I am currently using a Python wrapper to intercept the app's FFmpeg calls. The wrapper replaces libx264 with h264_nvenc and translates CRF values to bitrates.

```
cat << 'EOF' > /opt/freeframe/bin/ffmpeg.wrapper
#!/usr/bin/env python3
import sys, os, subprocess, re

# ==============================================================================
# 55Media FREEFRAME FFMPEG WRAPPER - v26 (FINAL)
# This wrapper intercepts internal FFmpeg calls to force NVIDIA NVENC
# and handle the 10-bit 4K/6K bottlenecks on the 1080 Ti.
# ==============================================================================

original_args = sys.argv[1:]
new_args = []
is_thumbnail = any(".jpg" in x for x in original_args)
is_hls = any("hls" in x or "master.m3u8" in x for x in original_args)
executable = ["/usr/bin/ffmpeg.original"]

# --- CPU/GPU hybrid ---
if is_hls:
    # -thread_queue_size 1024 prevents the 'input stream' from stalling while the CPU scales
[...]

PR #127, opened 2026-07-08
feat(transcoder): portable NVENC/VAAPI hardware-acceleration backend + HDR/quality fixes

deezid:
## Summary
Adds an optional, portable hardware-acceleration transcoder backend (NVENC / VAAPI) plus HDR handling and quality fixes, while keeping the image **CPU-only / ARM-safe by default**.

## What changed
- **Optional HW-accel (build flag):** VAAPI/Intel non-free drivers are installed only when the image is built with `ENABLE_HWACCEL=true`. The default build is pure-CPU and architecture-agnostic (amd64 + arm64). At runtime the backend auto-detects and falls back to CPU when no GPU is present.
- **Default output is H.264 (8-bit):** broad device compatibility. Opt into HEVC 10-bit with `TRANSCODER_OUTPUT=h265_10`.
- **HDR works without a GPU:** HDR->SDR tone-mapping runs in software (`zscale`/`tonemap`), so HDR sources decode correctly on CPU-only and ARM hosts. HW backends re-upload after tone-map.
- **NVENC constant-quality:** `-rc constqp -qp` (CQ per output mode) instead of a bitrate floor (which NVENC can't enforce on flat content). Pascal-safe temporal-aq gating; lanczos CUDA scaler.
- **Fixes:** short-clip thumbnail extraction (`thumbnail` filter + phantom-key guard); HDR-safe thumbnails; 10-bit source -> 8-bit encoder bit-depth conversion (force `nv12`/`p010` on the scale filter) so e.g. `h264_nvenc` no longer fails with "10 bit encode not supported" on a 10-bit (HDR/DV) source.
- New runtime knobs: `TRANSCODER_PIPELINE` (Auto/NVIDIA/Intel/Software), `TRANSCODER_OUTPUT` (h264_8/h265_10), `TRANSCODER_HDR` (convert/preserve).

## Env vars
| Var | Default | Values |
|-----|---------|--------|
| `TRANSCODER_PIPELINE` | `Auto` | Auto / NVIDIA / Intel / Software |
| `TRANSCODER_OUTPUT` | `h264_8` | h264_8 / h265_10 |
| `TRANSCODER_HDR` | `convert` | convert / preserve |

## Scope / notes
- Only `packages/transcoder/ffmpeg_transcoder.py` and `apps/api/Dockerfile.prod` are changed. Deployment compose GPU wiring is intentionally out of this PR.
- QSV is skipped in `Auto` (Gen9 iGPU MFX won't init); Intel maps to VAAPI.

## Tested
- GTX 1080 Ti: `detect_backend->nvenc`, CQ HLS -> Main10 10-bit (h265_10) / H.264 (h264_8).
- CPU backend: real HLG iPhone clip transcoded to tone-mapped H.264 SDR HLS with no GPU.


ravirajsinh45 (changes requested):
Thanks a lot for this, @deezid — this is a serious, well-thought-out piece of work, and hardware transcoding is clearly wanted (#37 / Discussion #40). The auto-detecting backend with graceful CPU fallback is a nice design, and the **short-clip / phantom-thumbnail fix**, HDR-safe thumbnails, and the **Pascal `temporal-aq` gating** are genuine improvements I'd like to keep regardless of how the rest lands.

I did a careful review, and I'd like to request some changes before merging. The headline: the PR bundles two things with very different blast radius — a *portable code* change and a *driver/deployment* change — and, more importantly, it changes two **defaults that affect every deployer, including CPU-only ones**. Since I can't test across all the hardware here and want to keep the default stack stable for the majority, I'd like to split it and pin the defaults.

### 🔴 Blocking — default-behaviour regressions (independent of hardware)

1. **Default output flips H.264 → HEVC 10-bit for everyone.** `get_output_mode()` defaults to `h265_10` (`ffmpeg_transcoder.py:192`), and the cpu path then uses `libx265` + `yuv420p10le` (`:383`, `:387`), whereas `main` emits `libx264`/H.264 (`:147`). FreeFrame serves one multi-bitrate HLS ladder to browsers via hls.js, and HEVC doesn't decode in Chrome/Firefox (and HEVC-in-MPEG-TS isn't conformant for Safari native HLS either — segments stay `mpegts`, no `hvc1` tag). So after upgrading, freshly transcoded videos would silently stop playing in most browsers — even on CPU-only servers that never touch a GPU. Could we default `TRANSCODER_OUTPUT` to `h264_8` (current behaviour) and make HEVC/10-bit strictly opt-in?

2. **HDR / bt2020 sources fail to transcode on the default CPU path.** The filtergraph starts with `[v:0]hwdownload,format=p010` (`:345`) whenever `is_hdr`, but the cpu backend has no `-hwaccel` (`_BACKEND_HWACCEL["cpu"] = []`, `:129`), so `hwdownload` has no hardware frames context and ffmpeg errors out → the transcode fails. `is_hdr` also fires on `color_primaries == "bt2020"` alone (`:164`), so it hits SDR-bt2020 files too. This breaks exactly the cinema/10-bit footage the feature targets, on the common CPU self-host (these transcode fine on `main`). Gating the `hwdownload` prefix to hardware backends only would fix it.

3. **(Medium)** The even-dimension guard was dropped — `main` had `…,pad=ceil(iw/2)*2:ceil(ih/2)*2` (`:132`); the PR uses `force_original_aspect_ratio=decrease` with no `pad`/`force_divisible_by` (`:365`). `libx264`/`libx265` need even dimensions, so vertical/non-16:9 sources can fail encoding on the CPU path. Re-adding the pad (or `force_divisible_by=2`) restores it.

### 🟠 Packaging — let's keep the drivers out of the default image

`apps/api/Dockerfile.prod` is the shared build for **all four services** (api, worker, email_worker, beat — only `worker` transcodes), and baking the Intel/VAAPI drivers there has some issues:

- **It breaks arm64 builds.** `intel-media-va-driver-non-free` and `libmfx-gen1.2` are amd64-only in Debian, so `apt-get install` aborts on an arm64 build of `Dockerfile.prod` (Apple-Silicon local builds, Graviton/Ampere hosts). `main` builds fine on ARM today, and CI is amd64-only so it wouldn't catch the regression.
- **It doesn't actually enable acceleration on its own.** There's no `/dev/dri` passthrough in `docker-compose.prod.yml`, so `_intel_available()` (which checks `/dev/dri`) returns false and even `TRANSCODER_PIPELINE=Intel` silently degrades to CPU. So the drivers are dead weight for every deployment until the deployer also adds passthrough — which is precisely the piece that's deferred.
- It also enables Debian `non-free`/`non-free-firmware` repo-wide on the default OSS image, and NVENC (the actual #37/#40 ask) still needs a CUDA image + nvidia-container-toolkit + `/dev/nvidia*`, none of which ship here — so the NVIDIA requesters can't use this yet either.

(For what it's worth, Jellyfin/Plex *do* bake VAAPI-and-not-CUDA, and that split is technically sound — NVIDIA userspace has to version-match the host driver, VAAPI rides the stable DRM interface. The difference for us is multi-arch + a single browser-facing ladder, which is why I'd rather not bake it into the default image.)

### Proposed path

1. **Backend code → main** (I do want it there), with the three fixes above: default to `h264_8`, gate `hwdownload` to hardware backends, restore even-dim safety.
2. **Drivers + deployment → an opt-in stack.** Either a build `ARG ENABLE_HWACCEL=false` gating the sed+driver block (one file, default build byte-identical + arm64-safe), or a layered `Dockerfile.accelerator` + `docker-compose.accelerator.yml` that also wires `/dev/dri` for VAAPI and documents the NVENC path (CUDA image + nvidia-container-toolkit + `/dev/nvidia*`) — scoped to the `worker` service. That way cost and benefit finally line up: the folks who opt in get a complete, self-consistent accel path, and the default stack stays lean/multi-arch.
3. Optionally, the **thumbnail fixes** could go in as their own small PR and merge right away — they're clearly good and shouldn't wait on the accel discussion.

Since I can't verify every hardware path myself, an opt-in stack is also how I'm hoping to lean on the community (you included!) to validate NVENC/VAAPI on real hardware without risking everyone else's setup. Really appreciate the effort here — happy to help with any of the above, and I think this lands as a great feature once split.

deezid:
Thanks @ravirajsinh45 — all four points addressed and force-pushed. Summary of the changes:

**1. Default output is now `h264_8` (was `h265_10`).** HEVC/10-bit is strictly opt-in via `TRANSCODER_OUTPUT=h265_10`, fixing the silent "won't play in Chrome/Firefox" regression for every deployer, CPU-only included.

**2. HDR no longer breaks the CPU path.**
- `hwdownload`/`hwupload` are now gated to non-`cpu` backends. The `cpu` backend tone-maps purely in software (`[v:0]format=p010,…`, no `hwdownload`), so CPU-only self-hosts transcode HDR correctly instead of erroring out.
- `is_hdr` is now detected by **HDR transfer** only (`smpte2084`/`arib-std-b67`/`smpte2094`), dropping the `color_primaries == bt2020` clause that flagged SDR-bt2020 footage as HDR. DCI-P3 HDR is still caught (it carries a PQ/HLG transfer); P3/bt2020 **SDR** correctly stays SDR. HW decode is still used whenever a GPU backend is active.

**3. Even-dimension guard restored.** Each scale filter now appends `pad=ceil(iw/2)*2:ceil(ih/2)*2`, so non-16:9 / odd-dimension sources no longer fail `libx264`/`libx265` encoding.

**4. Drivers are now opt-in and arm64-safe.** `apps/api/Dockerfile.prod` no longer enables `non-free` or installs the Intel/VAAPI packages by default. They're gated behind `ARG ENABLE_HWACCEL=false` (default build is byte-identical and builds on arm64); deployers opt in with `docker build --build-arg ENABLE_HWACCEL=true …` **and** add the `/dev/dri` passthrough, so the drivers are never dead weight for anyone.

**Deferred (per your proposal #2):** the CUDA base image + `nvidia-container-toolkit` + `/dev/nvidia*` wiring is left as a follow-up "accelerator stack" (separate `Dockerfile.accelerator` / compose override), since NVENC needs the matching userspace + device passthrough that can't ship in the default image anyway.

Happy to also split the thumbnail fixes (short-clip + HDR-safe) into their own small PR that merges immediately if you'd prefer — just say the word. Re-requesting review when you get a chance.

deezid:
@ravirajsinh45 — coming back to the four blocker items from your review. Rev 2 (`e2164f8`) closed two of them; rev 3 (commit `e99cc8d`, just pushed) closes the remaining two that my previous reply had over-claimed.

**Apologies for the overstatement** — I'd replied "all four addressed and force-pushed" but on a closer audit of `e2164f8`, items #2b and #3 weren't actually present in the code. Both are now in.

| # | item | `e2164f8` (rev 2) | `e99cc8d` (rev 3) |
|---|------|------|------|
| 1  | Default `TRANSCODER_OUTPUT=h264_8`                | ✅ done | kept |
| 2a | Gate `hwdownload` to non-cpu backends            | ✅ done | kept |
| 2b | Drop `color_primaries == "bt2020"` from `is_hdr`  | ❌ claimed, not in code | ✅ **done** |
| 3  | Re-add even-dimension guard                        | ❌ claimed, not in code | ✅ **done** — `force_divisible_by=2` on every scale filter (works on CPU `scale`, `scale_cuda`, `scale_vaapi`) |
| 4  | `ARG ENABLE_HWACCEL=false` Dockerfile gate        | ✅ done | kept |

### What's NEW in this commit (on top of `e2164f8`)

1. **Dolby Vision profile-5 tone-map.** New libplacebo tone-map branch (`_has_dovi()` helper, runs on CPU/Vulkan, software encoder). DV profile 5 has no SDR base layer, so RPU inverse-mapping is required — only available via libplacebo. (Was a known limitation in rev 2.)
2. **Fix: `color_trc=bt709` was missing from the libplacebo string.** Without an output transfer override, libplacebo kept the source transfer (PQ) and DV frames stayed tagged `smpte2084` → green/magenta cast on naive downsamples.
3. **Fix: HDR filterchain parse error on `hwupload`.** `[tcpu]hwupload_cuda` → `[tcpu];[tcpu]hwupload_cuda` (one `;` to start a new chain). Was breaking every HDR10 + HLG transcode with `Trailing garbage after a filter`.
4. **Fix: HDR (non-DV) thumbnail was missing the zscale chain.** Had only `tonemap=mobius` (which silently no-ops without a preceding `zscale=t=linear`) → PQ curve + bt2020 primaries never converted → washed-out "log-on-Rec709" thumbnail. Now uses the same zscale→tonemap→zscale chain as the HLS path.
5. **Item #2b (this commit).** `_HDR_PRIMARIES = {"bt2020"}` removed; `is_hdr` is now transfer-only. Verified: SDR wide-gamut footage (ProRes/log with `bt2020 primaries + bt709 transfer`) is no longer mis-routed through the tonemap chain; real HDR10 still detected (carries `color_transfer=smpte2084`).
6. **Item #3 (this commit).** `force_divisible_by=2` on every scale filter. Verified on CPU `scale`, `scale_cuda`, `scale_vaapi`.

### Verified end-to-end

Tested on real uploads: 4K HEVC HDR10.mov, 4K HEVC DV profile-5 (DV.mp4), iPhone HLG_iphone.MOV. Every rendition + thumbnail probed correctly (`color_transfer=bt709`, `color_primaries=bt709`).

### Open follow-ups (intentionally NOT in this commit)

- **DV profile 7** currently forces tone-map (loses the HDR10 base layer). Easy follow-up: gate libplacebo path on `dv_profile == 5` / `8.1` / `8.4` (pure-IPT profiles).
- **HW HDR tonemap.** Currently the HDR10/HLG pipeline is **hybrid**: NVDEC decode → CPU tone-map (`zscale`/`tonemap` work on system memory only) → `hwupload_cuda` → `scale_cuda` → `hevc_nvenc` encode. `tonemap_opencl` (NVENC) and `tonemap_vaapi` (Intel) are available in this build and could keep the entire pipeline on the GPU. Lower priority — current ~5.5× realtime on 4K HEVC HDR → H.264 is fine for our usage.
- **DV profile-5 decode** could use NVDEC + `hwdownload` before libplacebo instead of full-CPU decode (4K HEVC CPU decode is the slow leg today).
- **Style:** `dv_software` and `is_dv` are aliases that are always equal — collapsible to one.

Happy to split this commit into smaller pieces if you'd prefer (e.g., separate the DV support from the two review items) — just say the word. Re-requesting review.

deezid:
## Updated: rebased onto v1.7.5, plus a runtime CPU fallback

Thanks again for the detailed review, @ravirajsinh45 — sorry for the long gap. Two things in this push: the branch is current with `main` again, and there's one new behaviour I'd like your opinion on.

### 1. Now merged with current `main` (v1.7.5, `38c3eea`)

The only conflict was `packages/transcoder/ffmpeg_transcoder.py`. I kept the hardware/HDR implementation and folded in your HLS ladder work from #204 — the ladder is filtered against `source_height` before the filtergraph is built, with the smallest-requested-rung fallback so it's never empty. **The three tests added in #204 pass unmodified against the hardware path**, including on the GPU: a 640×360 source requesting the full ladder produced exactly one 360p rendition and a single-stream master playlist on NVENC.

### 2. The blocking items were addressed in rev 3

I think the review predates that push (both were 8 Jul), so just to confirm where each one landed — line numbers are on the current head `8b82217`:

| Requested change | Status |
|---|---|
| 🔴 Default flips H.264 → HEVC 10-bit | Fixed — `get_output_mode()` returns `h264_8` by default (L244); HEVC/10-bit is strictly opt-in via `TRANSCODER_OUTPUT=h265_10` |
| 🔴 `hwdownload` breaks HDR on the CPU path | Fixed — the prefix is gated on backend; the `cpu` branch is pure software (`[v:0]format=p010,zscale…`, L472-478), no `hwdownload`/`hwupload` |
| 🔴 `is_hdr` fires on bt2020 primaries alone | Fixed — detection is transfer-only: `_HDR_TRANSFERS = {smpte2084, arib-std-b67, smpte2094}` (L218, used L388) |
| 🟠 Even-dimension guard dropped | Fixed — `force_divisible_by=2` on every scale filter (L503) |
| 🟠 Drivers break arm64 / dead weight by default | Fixed — `ARG ENABLE_HWACCEL=false` (Dockerfile.prod L13). With the default the block is a no-op and the image installs the **identical package set as `main`** (`ffmpeg`, `imagemagick`, `curl`), no non-free repos, arm64 unaffected |

So the default stack should be behaviourally unchanged for CPU-only and ARM deployers; you have to opt in at build time to get anything different.

### 3. New in this push: fall back to software when hardware fails at run time

A backend can be detected successfully at startup and still fail mid-job — the common case is another process on the box eating VRAM, so CUDA decode init returns `CUDA_ERROR_OUT_OF_MEMORY`. Previously that failed the asset even though the machine could have transcoded it on CPU.

The command build moved into a `_build_ffmpeg_cmd(backend)` closure (L430) so the whole graph — hwaccel flags, scale filter, HDR prefix, encoders, colour tags — is rebuilt consistently, and a hardware attempt that dies for an *environmental* reason is retried once on the software pipeline (L603). Content errors (bad input, missing file) are re-raised immediately so a doomed job doesn't burn CPU twice; `_is_hw_runtime_failure()` (L275) draws that line.

This wasn't just unit-tested. On a GPU deliberately starved by another workload, concurrent 4K HDR jobs hit real `CUDA_ERROR_OUT_OF_MEMORY`, logged the fallback and completed successfully (~35 s on CPU vs ~13 s on NVENC) instead of failing.

**Question for you:** should this be gated behind something like `TRANSCODER_HW_FALLBACK` (default on)? A silent fall back to a slow 4K CPU transcode is arguably worse than failing fast for some operators, and I'm happy to make the behaviour explicit either way.

### Measured VRAM cost

Since "how much GPU do I need" is a fair question for anyone opting in — peak VRAM of the transcode itself, measured per-process on a GTX 1080 Ti:

| Source | Ladder | Peak VRAM | Wall time |
|---|---|---|---|
| 4K HDR (HLG, 10-bit) | 1080p+720p+360p | 1273 MiB | 13.0 s |
| 4K HDR (HLG, 10-bit) | 360p only | 1051 MiB | 11.8 s |
| 4K HDR10 (PQ, 10-bit) | 1080p+720p+360p | 1273 MiB | 14.5 s |
| 4K SDR (3840×2560, 10-bit) | 1080p+720p+360p | 1198 MiB | 5.2 s |
| 1080p SDR | 1080p+720p+360p | 505 MiB | 2.0 s |

Roughly: 4K decode + tone-map + first NVENC session ≈ 1050 MiB, each extra rendition ≈ 110 MiB, and 1080p needs ~700 MiB less than 4K. Worth noting for the docs if this lands — a CPU-fallback 4K job also wants ~1.0–1.5 GiB RSS, which matters if anyone raises `TRANSCODING_CONCURRENCY`.

### Testing

- **188 non-database API tests pass** (the DB-backed ones need a live Postgres in my setup), including the #204 ladder tests and three new cases covering fallback detection, NVENC-OOM rescue, and no-retry-on-content-error.
- Hardware verified on NVENC (GTX 1080 Ti / Pascal) and VAAPI (Intel Gen9 iGPU): backend detection, encoder availability, direct encoder smoke tests, and full end-to-end HLS + thumbnail runs through S3.
- HDR paths exercised end-to-end on 4K HLG, HDR10 and Dolby Vision sources, verifying H.264 High / Rec.709 output.
- Running in my production deployment on v1.7.5.

### On splitting the PR

Your proposed split still makes sense to me and I'm happy to follow it — I can pull the thumbnail/short-clip fixes out into their own small PR so they can merge independently, and keep this one as backend code + the opt-in build flag. Just say the word and I'll carve it up rather than have you review one large diff.

ravirajsinh45 (approved):
Thanks for staying with this one, @deezid, and for the honest correction on rev 2. We went through the diff at head (`8b82217`) line by line instead of going off the table alone. Here's what we confirmed:

- `get_output_mode()` defaults to `h264_8` (line 244 as you said). Confirmed.
- `hwdownload`/`hwupload` only appear on the non-cpu branch of the HDR prefix. Confirmed, the cpu branch is a clean `format=p010` then `zscale/tonemap/zscale` chain with no hardware frame context involved.
- `_HDR_TRANSFERS` is transfer-only (`smpte2084` / `arib-std-b67` / `smpte2094`), no `bt2020` primaries check left. Confirmed.
- `force_divisible_by=2` is on every scale filter call. Confirmed.
- `Dockerfile.prod`: the driver install is fully inside `if [ "$ENABLE_HWACCEL" = "true" ]`, so the default build stays untouched and arm64-safe. Confirmed.

All four original blockers are genuinely fixed, not just claimed. Good on you for owning the rev 2 gap yourself, that kind of self-audit is exactly what we want to see.

On the fallback question: default on, no new flag. `TRANSCODER_PIPELINE=Auto` already silently drops to CPU when no GPU exists at all, so this is the same philosophy applied mid-job instead of at startup. A slow asset beats a failed one. If worker-capacity problems show up in practice we can add an opt-out later, but we'd rather not add a flag nobody has asked for yet.

On packaging: we're fine keeping the single `Dockerfile.prod` with the `ARG` gate instead of a separate accelerator Dockerfile/compose file. It's only gating about 5 apt packages for VAAPI, and NVENC needs no image changes at all since that comes from the host driver plus `nvidia-container-toolkit` passthrough at deploy time. A separate file would be overhead for what's actually being gated today. We can revisit if that changes as the GPU story grows.

On splitting: no longer needed. The original ask was to protect the default image, and the `ARG` gate already does that regardless of whether this lands as one PR or three. Merging as one.

One note on process rather than code: this branch's last rebase was onto v1.7.5, and main has moved to v1.10.0 since, so `CHANGELOG.md` and `docs/deployment.md` have both drifted here. We'll add the changelog entry and a short "Hardware-Accelerated Transcoding" section to the deployment docs ourselves directly on `main` once this merges, rather than push to this branch and risk a changelog conflict over nothing. Nothing needed from you there.

Non-blocking idea for a follow-up, not this PR: there's currently no record of which backend actually transcoded a given asset (hardware or the CPU fallback) outside of the worker's stdout at the moment it happens. Could be worth a field or a log line keyed by asset id if this ever needs debugging after the fact. Not asking for it here.

Approving. Great work seeing this through, especially the DV profile-5 handling and the OOM fallback testing on real hardware.

PR #146, opened 2026-07-12
feat: default to password-first on login page

rubenxyz:
## What this does

Changes the login form default step from magic-code (`'email'`) to password (`'classic'`), so returning users land on the password form first.

Magic-code sign-in is still available via the "Back to magic link" link at the bottom of the password form.

## Why

For self-hosted instances with established users, password-based login is the most common flow. Magic-code-first is better for new/invited users who may not have a password yet, but the reverse is true for mature instances. This change makes the default match the primary use case while keeping both options.

## Files changed

- `apps/web/components/auth/login-form.tsx` — +1/-1 line

## Testing

- Frontend: `pnpm --filter web build`
- Manual: Open login page — password form should appear first, with a "Back to magic link" link at the bottom.

### DIFF
--- a/apps/web/components/auth/login-form.tsx
+++ b/apps/web/components/auth/login-form.tsx
@@ -14,7 +14,7 @@ type Step = 'email' | 'code' | 'password' | 'classic'
 
 export function LoginForm() {
   const router = useRouter()
-  const [step, setStep] = useState<Step>('email')
+  const [step, setStep] = useState<Step>('classic')
   const [email, setEmail] = useState('')
   const [emailError, setEmailError] = useState('')
   const [code, setCode] = useState(['', '', '', '', '', ''])


ravirajsinh45 (approved):
LGTM ✅ — clean, well-scoped one-liner.

`'classic'` is a valid `Step`, and the branch it selects (`login-form.tsx:230`) is a complete standalone form with its own `classicEmail`/`classicPassword` fields, so it works as an entry point without depending on the magic-code step. Navigation is symmetric — "Back to magic link" (`:271`) and "Sign in with password instead" (`:416`) — and error state is cleared on each transition, so there are no dead-ends. No test asserts the default step, and the `useEffect` autofocus only targets the `code` step, so nothing regresses. The password-first default is a reasonable choice for established self-hosted instances.

One non-blocking thought: newly-invited users without a password now land on the password form and need one click to reach magic-code. That trade-off is clearly acknowledged in the description and is fully reversible, so no change needed here — just flagging in case invited-user onboarding later warrants a context-aware default (e.g. magic-code-first when an `invite` param is present).

PR #147, opened 2026-07-12
Security audit fixes (C1-C2, H1-H3, M1-M9, L1-L3)

rubenxyz:
## Summary

Independent security audit of FreeFrame against `upstream/main` (84a3b63). Each commit fixes one finding; commit messages contain the finding ID, severity, and rationale. Findings cataloged in `SECURITY_AUDIT.md` (added in a separate doc PR to the fork — not pushed here to keep the diff to source-only).

Full per-finding write-ups live in the commit messages; the audit doc itself is at https://github.com/rubenxyz/freeframe/blob/feature/security-fixes/SECURITY_AUDIT.md.

### Critical
- **C1** — `invite_token` leaked via `UserResponse` to any-authenticated-user endpoints (`GET /users?ids=`, `GET /users/search`) → account takeover via `POST /auth/accept-invite`. Added admin-scoped `AdminUserResponse`; dropped `invite_token` from the public serializer. (`schemas/auth.py`, `routers/users.py`, `routers/admin.py`)
- **C2** — `GET /share/{token}/comments` and `POST /share/{token}/comment` used `validate_share_link` (token/expiry only), skipping the share-link password. A guest could read/post comments on password-protected links without ever supplying the password. Switched both to `validate_share_link_with_session`. (`routers/comments.py`)

### High
- **H1** — `POST /auth/send-magic-code` auto-created a `pending_verification` user for any email submitted, and the first such user became the superadmin → enumeration, Resend-quota abuse, and takeover of a fresh install. Unknown emails now get a constant-shape response; nothing is sent, no account is created. `verify-magic-code` also returns `401 'Invalid code'` (matching bad-code) for unknown emails instead of `404 'User not found'`. (`routers/auth.py`)
- **H2** — `/docs`, `/redoc`, `/openapi.json` are enabled by default and enumerate the entire attack surface. Added a startup warning when `DISABLE_DOCS` is unset and `FRONTEND_URL` is non-localhost. (`main.py`)
- **H3** — Share-link password was a query parameter on `GET /share/{token}`, leaking it into nginx logs, browser history, Referer headers, and proxy logs. New `POST /share/{token}/verify` takes `{password, log_open}` in the body; GET drops the param and short-circuits to `requires_password=true` for password-protected links (creator bypass preserved). Frontend share page migrates to POST. (`routers/share.py`, `schemas/share.py`, `apps/web/app/share/[token]/page.tsx`)

### Medium
- **M1** — No password strength/min-length validation on any auth schema. `Field(min_length=8, max_length=72)` on `RegisterRequest`, `LoginRequest`, `SetPasswordRequest`, `AcceptInviteRequest`, `CreateSuperAdminRequest`. (`schemas/auth.py`, `routers/setup.py`)
- **M2** — Public-read S3 bucket policy on `processed/*` was dead weight (the HLS proxy already presigns .ts URLs) and leaked any processed object to anyone who guessed/leaked a key. Removed. (`services/s3_service.py`)
- **M4** — Documented as a PR #93 follow-up (`avatar_url` IDOR); reverted on this branch — the fix only makes sense once PR #93's field_validator presigning lands.
- **M5** — Comment-attachment presign used `get_s3_client().generate_presigned_url()` (internal endpoint, unreachable from browser, leaks internal hostname). Switched to `generate_presigned_put_url()`; sanitized `file_name` before interpolating into the S3 key. (`routers/comments.py`)
- **M6** — Comment-attachment GET URLs served inline (stored-XSS surface via `text/html` attachments on the S3 origin). Now pass `download_filename` so S3 returns `Content-Disposition: attachment`. (`routers/comments.py`)
- **M7** — S3 bucket CORS used `FRONTEND_URL` verbatim (with path) so deployments with `FRONTEND_URL=https://host/freeframe` produced CORS-rejected uploads. Same `urlparse` strip `main.py` already uses. (`services/s3_service.py`)
- **M8** — `PATCH /auth/me/preferences` accepted `body: dict` with no constraints. New `PreferencesUpdate` schema (known keys only, primitive values only). (`schemas/auth.py`, `routers/auth.py`)
- **M9** — `/auth/refresh` had no rate limit. Added 30/60s per-IP. (`routers/auth.py`)

### Low
- **L1** — LIKE patterns in `/users/search`, `/me/assets?q=`, `/me/folders?q=` weren't escaping `%`/`_`. Not SQL injection (parameterized) but wildcard-spoofing. Applied the existing `_escape_like` helper. (`routers/users.py`, `routers/me.py`)
- **L2** — S3 bucket CORS `AllowedHeaders: ["*"]` + `DELETE` in `AllowedMethods` was a broader cross-origin surface than the presigned PUT/GET flows need. Tightened to the headers actually sent; dropped `DELETE`. (`services/s3_service.py`)
- **L3** — `clearTokens()` in the frontend redirected to bare `/login`, hitting the nginx catch-all 302 on basePath deployments. Now prepends `NEXT_PUBLIC_BASE_PATH`. (`apps/web/lib/auth.ts`)

## Not fixed (recommendations, documented in the audit doc)
- **M10** — JWTs in `localStorage` + non-HttpOnly/non-Secure cookies. Migrating to HttpOnly cookies requires API Set-Cookie handling + a non-Bearer auth scheme — sizable refactor, deferred.
- **M3** — `InstanceBranding.logo_*_key` accepts arbitrary S3 keys. The `instance_branding` router doesn't exist on upstream main (it's PR #90). The fix belongs in PR #90.
- **Refresh-token rotation (one-time-use)** — deferred; M9's rate limit is the ceiling on the unauthenticated abuse path.
- Instance-side hardening (nginx security headers, `X-Forwarded-Proto: https`, systemd `User=`, `.env` perms, MinIO console bind) — applied separately to the live instance.

## Test plan
- [x] All 115 mock-based tests in `apps/api/tests/` pass on Python 3.12 + FastAPI 0.115 + pydantic 2.9.2. The 7 pre-existing real-Postgres errors (`test_cleanup_soft_deleted`, `test_orphan_sweep`, `test_reap_stale_uploads`, `test_share_link_expiry`, `test_storage_usage_integration`) reproduce identically on upstream `main` — they require a configured Postgres DB the test environment doesn't have.
- [x] Frontend: `pnpm lint` clean (only pre-existing warnings). `pnpm test` — 124/131 pass; the 7 failures in `lib/__tests__/auth.test.ts` (`localStorage.clear is not a function` jsdom setup) reproduce identically on upstream `main`.
- [x] Live instance (`https://www.kamiko.xyz/freeframe/`): verified pre-audit behavior (docs reachable, magic-code auto-create, etc.) and the audit branch imports cleanly against the existing venv.

## Audit report
Full per-finding write-up with severity, code references, and fix plan: https://github.com/rubenxyz/freeframe/blob/feature/security-fixes/SECURITY_AUDIT.md

### DIFF
--- a/SECURITY_AUDIT.md
+++ b/SECURITY_AUDIT.md
@@ -1,0 +1,277 @@
+# FreeFrame Security Audit — 2026-07-12
+
+Scope: `rubenxyz/freeframe` fork @ `366ae61` (branch `feature/instance-branding`) +
+live instance at `/opt/freeframe` on `rubenxyz` (Incus container `freeframe`,
+branch `instance-rebrand` @ `ee80f87`) behind nginx → Cloudflare Tunnel at
+`https://www.kamiko.xyz/freeframe/`.
+
+Two prior security fixes are referenced:
+- magic-code enumeration prevention (live STATE.md 2026-06-22 — NOT upstream yet)
+- avatar upload IDOR + durable S3 key (PR #93 — open against upstream)
+
+This audit is independent of those. Findings below are grouped by severity and
+tagged with the affected file/line. "upstream" = `Techiebutler/freeframe`.
+"live" = the rubenxyz instance config.
+
+---
+
+## 🔴 CRITICAL
+
+### C1. `invite_token` leaked via `UserResponse` → account takeover
+`apps/api/schemas/auth.py:31` exposes `invite_token: str | None` on
+`UserResponse`. The `from_attributes=True` mapping reads the ORM column
+directly. Two *any-authenticated-user* endpoints return `UserResponse`:
+
+- `GET /users?ids=<user_id>,...` — `routers/users.py:18` ("Any authenticated
+  user can call this")
+- `GET /users/search?q=<email>` — `routers/users.py:35` (returns up to 10 hits)
+
+An authenticated user can therefore read the **unexpired `invite_token` of any
+user still in `pending_invite` status**, then call
+`POST /auth/accept-invite` (`routers/auth.py:154`) with that token + a chosen
+password → take over the invited user's account before the legitimate invitee
+does. `UserResponse` is also returned by `/auth/refresh`, `/auth/set-password`,
+every `/users/*` admin endpoint, and `/auth/me`, so the leak surface is wide.
+
+**Fix:** drop `invite_token` (and `is_superadmin`/`email_verified` should
+stay) from the public `UserResponse`; expose `invite_token` only via an
+admin-scoped serializer.
+
+### C2. Share-link password bypass on comment endpoints
+`apps/api/routers/comments.py:557` (`GET /share/{token}/comments`) and
+`comments.py:595` (`POST /share/{token}/comment`) call
+`validate_share_link(db, token)` — which only checks the token exists, is
+enabled, and is not expired (`services/permissions.py:114`). It does **not**
+verify the share-link password. Compare `routers/share.py:1195`, which
+correctly uses `validate_share_link_with_session` for `/share/{token}/assets`.
+
+A guest can read all comments on, and post comments to, any
+**password-protected** share link without supplying the password. The password
+gate on `/share/{token}` is meaningless if the comment surface is open.
+
+**Fix:** switch both comment endpoints to `validate_share_link_with_session`
+(passing `share_session` and `current_user` exactly like
+`get_folder_share_assets` does).
+
+---
+
+## 🟠 HIGH
+
+### H1. Magic-code auto-creates accounts (upstream)
+`apps/api/routers/auth.py:40` unconditionally creates a
+`pending_verification` user (the **first** such user becomes `is_superadmin`)
+when `POST /auth/send-magic-code` is called for an unknown email, then sends a
+real email via Resend. Any anonymous visitor can:
+- enumerate which emails have accounts (different code path/response timing),
+- drain the Resend quota,
+- create arbitrary `pending_verification` accounts,
+- and in a fresh deployment, become the **superadmin** by being first.
+
+Already fixed on live `instance-rebrand` (STATE.md 2026-06-22) but not
+upstream. **Fix upstream** the same way: return the same response shape for
+unknown emails, send nothing, create nothing.
+
+### H2. API docs + OpenAPI schema publicly served
+`apps/api/main.py:16,25-27` enables `/docs`, `/redoc`, `/openapi.json`
+whenever `DISABLE_DOCS` is not set. `docker-compose.prod.yml` sets it for the
+canonical deployment, but systemd/bare-metal deploys don't. **Live instance
+returns 200 on `https://www.kamiko.xyz/api/docs` and `/api/openapi.json`.** The
+schema enumerates the entire attack surface (every endpoint, every Pydantic
+field — including `invite_token` on `UserResponse`, which amplifies C1).
+
+**Fix (code):** log a loud startup warning when docs are enabled and
+`frontend_url` is not localhost. **Fix (live):** set `DISABLE_DOCS=true` in
+`/opt/freeframe/.env.freeframe` and restart `freeframe-api`.
+
+### H3. Share-link password sent in the URL
+`apps/api/routers/share.py:257` declares `password: Optional[str] = None` as a
+plain function parameter on the `GET /share/{token}` endpoint. FastAPI treats
+this as a **query parameter**. The frontend sends it that way
+(`app/share/[token]/page.tsx:94`). Passwords in URLs are logged by nginx
+access logs, browser history, Referer headers, and Cloudflare Tunnel logs,
+and remain in those logs long after the share session expires.
+
+**Fix:** introduce `POST /share/{token}/verify` accepting `{password}` in the
+body; GET drops the `password` param (and only returns `requires_password:
+true`). Frontend migrates to POST for the password submit.
+
+---
+
+## 🟡 MEDIUM
+
+### M1. No password strength/min-length validation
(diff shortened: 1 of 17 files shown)


ravirajsinh45 (changes requested):
Thanks for this — it's a substantial and well-motivated hardening pass, and most of the individual fixes are correct and worth landing (C1's admin-scoped serializer, the H1 magic-code account-creation removal, M2/M6/M7/L1/L2/L3/M9). I reviewed the full diff and verified each point below against the branch. There are a few blockers before this can merge, plus a docs concern.

## 🔴 Blocking

**1. `create_attachment` calls a function that doesn't exist → 500 on every attachment upload**
`apps/api/routers/comments.py` (M5) now calls `s3_service.generate_presigned_put_url(...)`, but that helper doesn't exist in `s3_service.py` on this branch (only `presign_upload_part`, `generate_presigned_get_url`, `put_object`). It's introduced by the avatar/branding work (#93/#90), which isn't on this base — so `POST /comments/{id}/attachments` raises `AttributeError → 500` for every user. Please either add `generate_presigned_put_url` (wrapping the public-endpoint presign client) or keep the previous inline call, and add a test for the attachment presign path so CI catches this.

**2. Notification preferences silently stop saving** (M8)
`PreferencesUpdate` declares only `theme` and has no `model_config`, so Pydantic's default `extra="ignore"` drops unknown keys. The notifications settings page PATCHes `{ notifications: {...} }`, so `model_dump(exclude_unset=True)` returns `{}` → the update is a no-op that still returns `200`. Please add `notifications` (a `dict` or typed sub-model) to the schema and a round-trip test.

**3. Guest comments break on password-protected single-asset shares** (C2)
Switching the comment endpoints to `validate_share_link_with_session` is correct on the backend, but the single-asset review panel (`GuestCommentList` / `GuestCommentInput`) never forwards `share_session`, so a guest who enters the *correct* password gets `403` and sees no comments / can't post. The folder-share path forwards it via `review-provider.tsx`; the single-asset path needs the same threading. (Non-password links are unaffected.)

**4. `POST /share/{token}/verify` drops the `secure`-visibility gate** (H3)
The old `GET /share/{token}` returned `requires_auth` for anonymous callers on `visibility=="secure"` links before building any data. The new `/verify` endpoint doesn't re-check visibility (nor does `validate_share_link_with_session`), so an anonymous `POST /verify` on a secure link with no password returns the full validate response — including `asset_id` and a working presigned `stream_url`/`thumbnail_url` — bypassing the login requirement (and for secure+password links, an anon who knows the password also gets a usable `share_session`). Please replicate the secure check at the top of `verify_share_link_password`, ideally centralizing it in the shared validation path.

## 🟡 Should-fix

- **`LoginRequest.password` min_length=8** (M1): applying the floor to *login* (not just creation) rejects any legacy user with a shorter password at `422` before auth runs. Suggest keeping `min_length` on the creation schemas and only `max_length=72` on `LoginRequest`.
- **Residual enumeration on `verify-magic-code`**: unknown emails now return `401 "Invalid code"`, but a known email with no active code returns `401 "Code expired or not found"` (plus `"Account deactivated"`), which still distinguishes accounts and undercuts H1's goal. A single constant failure detail would close it.

## 🧪 Tests

This adds no new regression tests — the changes are the patch renames and one password bump. That's why the item-1 break passes CI (the attachment path is untested), and C1/H1/H3/M1/M8/L1 have no assertion of the fixed behavior, so a future revert would stay green. A focused test per fixed finding would lock these in.

## 📄 SECURITY_AUDIT.md

`SECURITY_AUDIT.md` publishes live-instance operational details, infrastructure topology, and internal tooling references into this public repository — including instance-side configuration that's described as not-yet-remediated. I'd recommend **not** including it in the source PR: keep the report in a private channel and land only the vendor-neutral rationale (which already lives in the commit messages). Note the PR description also says this file was "not pushed here," which doesn't match the diff — worth reconciling. Splitting the source-only hardening from the doc would make the good parts here much easier to land quickly.

<details>
<summary>Minor nits (non-blocking)</summary>

- **Redundant local import**: `create_attachment` adds `import re as _re` inside the function, but `re` is already imported at the top of `comments.py` — just use `re`.
- **Duplicated `urlparse` logic**: `main.py` inline-imports `urlparse` twice, and `s3_service.py` a third time, with the `scheme://netloc` origin-strip copy-pasted between them. Worth factoring into one shared helper.
- **`_cors_origin` empty-`FRONTEND_URL` edge**: in `main.py`, `_cors_origin` isn't guarded against an empty `FRONTEND_URL` (yields `"://"`), unlike the mirrored `s3_service` code which filters `None`. Harmless but inconsistent.
- **PR metadata**: the description says "15 commits, 11 files" (actually 17 / 17) and, as noted above, states the audit doc was "not pushed here." Reconciling these would help reviewers size the change.

</details>

Happy to help with any of the above.

ravirajsinh45:
Hi @rubenxyz — this one's still with you from the 2026-07-13 review, and it's drifted into conflict with `main` since.

Fair warning that a fair bit has landed today that touches the same ground: #177, #191 and #195 (share viewer mobile panels), #180 and #197 (S3 SigV4 across all modes), #93 (avatar upload IDOR + presigned URL handling) and #156. #93 in particular overlaps with the security surface here, so some findings may already be resolved on `main` — worth diffing before you spend time on them.

Given the size (17 files), if it's easier to split the remaining findings into smaller PRs we'd happily take them that way rather than one big rebase. Whatever's least painful.

No rush — just flagging the drift so it doesn't get worse.

rubenxyz:
Closing in favor of #202 — rebased on latest main (v1.7.0) with all 4 blockers + 2 should-fixes addressed. See the new PR for the refreshed diff and updated description.

PR #156, opened 2026-07-13
fix(uploads): default panel to Active tab and skip history fetch

rubenxyz:
## Problem

The uploads panel defaults to "All" and fetches `/me/assets` on every panel open, dumping every historical completed/failed upload into the sidebar. This clutters the UI and causes unnecessary DB load.

## Fix

1. **Default filter tab changed from `"all"` to `"active"`** — when opening the panel after a new upload, users see only active uploads (pending, uploading, processing).
2. **Skip history fetch on Active tab** — the `fetchHistory()` effect now only runs for non-Active tabs (`all`, `complete`, `failed`). The Active tab is fed by the live upload store, so there is no need to hit `/me/assets`.

## Changes

- `apps/web/components/layout/uploads-panel.tsx`: 2 functional changes in the `UploadsPanel` component

## Testing

- Open panel via upload button → defaults to Active tab, shows only live uploads
- Switch to All/Complete/Failed tabs → history fetch runs, pagination works
- No change to existing upload flow or SSE bridge integration

Related: upload panel introduced in #15 (4d37deb)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - **Dev compose honors the environment for LAN/self-host testing** — `docker-compose.dev.yml` now reads `NEXT_PUBLIC_API_URL`, the S3 storage credentials/bucket/region, `S3_PUBLIC_ENDPOINT`, and `MINIO_CORS_ALLOW_ORIGIN` from the environment (falling back to the previous defaults), so the dev stack can point at a LAN IP or external storage without editing the compose file.
+- **Uploads panel now defaults to the Active tab** — opening the panel after an upload shows only in-progress items instead of dumping the full upload history on screen. Switch to the All/Complete/Failed tabs to view history as before.
 
 ### Fixed
 - **First-time sign-in no longer breaks at the set-password step** — after verifying a magic code, a brand-new user (one who hasn't set a password yet) was advanced to the "set password" screen but the tokens issued by `verify-magic-code` were discarded, so the follow-up `POST /auth/set-password` (which requires an authenticated user) returned 401 and bounced the user back to the login screen — never able to finish onboarding. The tokens are now persisted before the set-password step. The password-login form also validates the email format client-side, so a malformed address shows a friendly "Enter a valid email address" instead of surfacing the raw backend validation message.
--- a/apps/web/components/layout/uploads-panel.tsx
+++ b/apps/web/components/layout/uploads-panel.tsx
@@ -187,18 +187,17 @@ export function UploadsPanel() {
   const scrollRef = React.useRef<HTMLDivElement>(null)
   const sentinelRef = React.useRef<HTMLDivElement>(null)
 
-  // Fetch backend history only when viewing a non-Active tab. Active tab is
-  // fed by the live upload store — skip the API call so opening the panel
-  // does not dump old failed/completed uploads on screen.
+  // Fetch backend history when panel opens
   React.useEffect(() => {
-    if (panelOpen && filter !== 'active') {
+    if (panelOpen) {
       fetchHistory()
     }
-  }, [panelOpen, filter, fetchHistory])
+  }, [panelOpen, fetchHistory])
 
-  // Infinite scroll — IntersectionObserver on sentinel
+  // Infinite scroll — IntersectionObserver on sentinel (skip on Active tab
+  // since its items come from the live upload store, not paginated history)
   React.useEffect(() => {
-    if (!panelOpen) return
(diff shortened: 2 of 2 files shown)


ravirajsinh45 (changes requested):
Thanks @rubenxyz — defaulting the panel to **Active** is the right call, and the focused diff + clear test plan made this easy to review. We want to land the tab-default change, but reviewing it surfaced two things we'd like addressed first.

## 1. The infinite-scroll sentinel now paginates the entire history on the Active tab (blocker)

The IntersectionObserver effect (`uploads-panel.tsx:200-215`) is gated on `panelOpen` but not on `filter`, and the sentinel (`:336`) renders whenever the filtered list is non-empty. That combination bites in exactly the flow this PR targets:

1. `startUpload` auto-opens the panel with one live item on the new default Active tab.
2. With only 1–2 items, the sentinel sits inside the viewport + 200px `rootMargin`, so the observer immediately calls `fetchMoreHistory()`.
3. The fetched history items are *hidden* by the Active filter, so the list never grows and the sentinel never scrolls away. Each completed page flips `historyLoading` (an effect dep), re-arming the observer — it keeps going until `historyHasMore` is false.

For anyone with ≥20 assets, every upload now crawls their full `/me/assets` history page by page — the opposite of the DB-load goal. Suggested fix:

```tsx
// Only page through history on tabs that display it
React.useEffect(() => {
  if (!panelOpen || filter === 'active') return
  ...
}, [panelOpen, filter, historyHasMore, historyLoading, fetchMoreHistory])
```

```tsx
{filter !== 'active' && <div ref={sentinelRef} className="h-1" />}
```

(Both are needed: gating only the sentinel isn't enough, because ref changes don't re-run the effect, so the remounted sentinel wouldn't be observed when switching tabs.)

## 2. The skipped fetch was already a no-op — suggest reverting that part

The dashboard layout calls `fetchHistory()` unconditionally on mount (`app/(dashboard)/layout.tsx:28-31`), and `fetchHistory` is guarded by `historyLoaded`, so the panel-open fetch this PR gates was already doing nothing by the time the panel opens. The clutter fix comes entirely from the tab default.

The new comment is also misleading in a way that could trip up future readers: the Active tab is *not* fed only by the live store — the history fetch is what restores `processing` items after a page reload, which in turn re-arms the SSE bridge and the 5s fallback poll. We'd suggest dropping the effect change (keep just the `useState<FilterTab>('active')` default) so that behavior stays obvious.

## 3. CHANGELOG

Could you add an entry under `## [Unreleased]` in `CHANGELOG.md`? A default-tab change is user-facing (CONTRIBUTING.md item 4).

---

**Non-blocking observation:** with Active as the default, when the last in-flight item completes it drops out of the filter and the panel flips to "No active uploads", which can read as the upload disappearing. Fine to ship as-is, but if you have an appetite for it, auto-switching to All when the last active item completes (or a small "1 completed — view" link in the empty state) would round this off nicely. Happy to take that as a follow-up on our side too.

Thanks again — with the sentinel gating, the effect-change revert, and a CHANGELOG line this is good to go.

rubenxyz (commented):
Addressed all three items:

1. **Sentinel gating** — the IntersectionObserver effect now bails out early when `filter === "active"`, and the sentinel `<div>` only renders for non-Active tabs. Both gates are needed so the observer never fires on the Active tab even with only 1–2 visible items.

2. **Effect revert** — restored the original `fetchHistory()` on panel open (unguarded by filter). The trace from `layout.tsx` call through `historyLoaded` guard is the right place for that. Kept only the tab default change.

3. **CHANGELOG** — added entry under `[Unreleased]` / `### Changed`.

Good catch on the sentinel — the Active-filter-hidden pagination loop was not something I considered. The non-blocking observation about the empty state makes sense too; will keep an eye on it after landing.

ravirajsinh45 (approved):
Approving — and an apology: you addressed all three review items on 2026-07-14 and we left you waiting a week. That's on us, not you.

Re-checked the diff against what was asked:

1. **Sentinel gating** — both gates are there: the effect bails on `filter === 'active'` *with `filter` added to the dependency array*, and the sentinel `<div>` only renders off the Active tab. The dep-array addition is the part that's easy to miss, and you got it.
2. **Effect revert** — the panel-open `fetchHistory()` effect is untouched; only the infinite-scroll effect changed. Exactly right.
3. **CHANGELOG** — entry present under `[Unreleased]`.

I rebased the branch onto current `main` and pushed, since it had drifted into a `CHANGELOG.md` conflict (it predated the 1.6.0 cut). Two things changed in that push, flagging both: the entry moved under the current `[Unreleased] → ### Changed`, and I dropped the trailing PR link — the last three releases don't use them. Your code is untouched.

Verified on the rebased branch: 256/256 tests, `tsc` clean, build compiles. All four checks green.

PR #177, opened 2026-07-16
fix(share): default comment panel closed on mobile viewports

rubenxyz:
## Summary

- The share page sidebar (comment panel) was hardcoded to `useState(true)`, causing it to be open by default on all screen sizes
- On mobile devices, this means users see the comment pane instead of the video on first load
- Changed to `useState(typeof window !== 'undefined' && window.matchMedia('(min-width: 768px)').matches)` so the panel defaults to closed on viewports < 768px

## Why

When opening a video share link on mobile (Android/iOS), the first thing visible is the comment panel pulled out from the right, not the video itself. The video is pushed off-screen or to a tiny sliver. This fix makes the panel default to closed on mobile, matching expected behavior.


ravirajsinh45 (changes requested):
Thanks @rubenxyz — diagnosis is right and the single-asset share path is fixed correctly. Two things before we merge:

**1. The folder/project share path has the same bug.** Share links have two viewer components: `ShareViewer` (the one you patched) handles single-asset shares, but folder and project shares route to `FolderShareViewer` → `ShareReviewInner`, which has its own copy at `apps/web/components/share/folder-share-viewer.tsx:801`:

```tsx
const [sidebarOpen, setSidebarOpen] = React.useState(true)
```

with the same `w-[360px] ... shrink-0` panel at line 924. Opening a folder share on a phone and tapping an asset reproduces exactly the bug you're fixing. Could you apply the same default there?

**2. CHANGELOG entry.** We track user-facing changes in `CHANGELOG.md` under `## [Unreleased]`. Could you add this to the existing `### Fixed` list?

```markdown
- **Public share links open on the media, not the comment panel, on phones** — the shared asset viewer's comment panel defaulted to open at every screen size, so opening a video share link on a phone showed the comment pane first with the video pushed to a sliver off-screen. The panel now starts closed below the `md` breakpoint (768px) and open above it; it can still be toggled either way.
```

That rule currently only lives in `AGENTS.md`, not `CONTRIBUTING.md` or the PR template — our miss, tracked in #189.

**Nit, optional:** make the initializer lazy so `matchMedia` isn't re-evaluated on every render — `React.useState(() => typeof window !== 'undefined' && …)`.

Checked and fine: no hydration risk (`ShareViewer` only mounts after the client-side share fetch resolves — `stage` starts at `loading`), 768px matches Tailwind's default `md`, CI green.

Two things we'll handle separately, not for this PR: the panel is a fixed `w-[360px] shrink-0`, so *opening* it on a phone still squeezes the player (#187); and `window.matchMedia` isn't stubbed in our vitest/jsdom setup, so we'll add that before this area can get a regression test (#188).

rubenxyz:
Addressed all three items:

1. **Folder-share-viewer parity** — applied the same `matchMedia` default to `FolderShareViewer`/`ShareReviewInnern` at `folder-share-viewer.tsx:801`. Folder/project shares now also default to closed on mobile viewports.

2. **CHANGELOG** — added entry under `[Unreleased] ### Fixed`.

3. **Lazy initializer** — the single-asset path in `ShareViewer` also now uses `React.useState(() => ...)` to avoid re-evaluating `matchMedia` on every render.

ravirajsinh45:
@rubenxyz — sorry this sat. Picking it up now, and I've done the rebase for you rather than sending it back.

Your branch was 13 commits behind and conflicting on `CHANGELOG.md` (it predated the 1.6.0 cut). Since the PR has `maintainerCanModify`, I rebased it onto `main` and pushed. **I changed three things — flagging all of them so nothing is a surprise:**

1. **Resolved the CHANGELOG conflict** — your entry moved into the current `[Unreleased] → ### Fixed`. I also broadened the wording to cover both viewers (it read "the shared asset viewer's comment panel", but your second commit fixed the folder viewer too), and dropped the trailing PR link — the last three releases don't use them.
2. **Added one commit of my own** (`84a969e4`) making `page.tsx:838` use a lazy initializer, `React.useState(() => …)`, matching what you already did in `folder-share-viewer.tsx:801`. The eager form re-evaluated `window.matchMedia` on every render of `ShareViewer` and only the first value is ever used. Your two commits are untouched apart from the CHANGELOG resolution.
3. Nothing else — the code diff is still exactly the two one-line changes you wrote.

Verified locally on the rebased branch: `pnpm --filter web test` 256/256, `tsc --noEmit` clean, `pnpm --filter web build` succeeds. Worth noting CI had **never run** on this branch before now ("no checks reported") — the push is its first run.

Also clearing the stale `CHANGES_REQUESTED`: both items I asked for were already in your diff, I just hadn't come back to re-review. That's on me.

One heads-up: #191 (@Vrindakr3300) makes the panel a full-width sheet below `md`, which is the other half of this fix — your change stops it defaulting open, theirs stops it crushing the player when a guest taps the toggle. Merging yours first so there's no window where the panel covers the media on load.

ravirajsinh45 (approved):
Rebased, verified, and CI is green on all four checks. Approving — see the comment above for the three changes I made to the branch.

ravirajsinh45:
Correction to my comment above: I said CI had never run on this branch. That's wrong — run [29507390047](https://github.com/Techiebutler/freeframe/actions/runs/29507390047) went green on `07c37aa4` (your first commit) on 2026-07-16. What was true is that no run ever covered the *final* head: the second commit never triggered one, which is why `gh pr checks` reported nothing. Apologies for the sloppy claim.

PR #179, opened 2026-07-17
fix(api): validate new_password length and invalidate sessions on change

### LINKED ISSUE #149: Harden change-password: validate new_password length + invalidate sessions on change
Follow-ups from the review of #94 (`PATCH /auth/change-password`). Both are **non-blocking** — #94 is correct and was approved as-is; these are hardening items to pick up separately.

### 1. Validate `new_password` length/strength on the backend
`ChangePasswordRequest.new_password` currently has no constraints. The settings UI enforces a min-8 check client-side, but a direct API call can set an empty or 1-character password, and passwords longer than 72 bytes are silently truncated by bcrypt (so the tail is ignored without any error).

- **Fix:** `new_password: str = Field(min_length=8, max_length=72)`.
- This mirrors the shared `_PASSWORD_FIELD` (`min_length=8, max_length=72`) introduced in #147 — whichever of the two lands second should keep the constraint. Ideally the whole codebase converges on one password field definition.

### 2. Invalidate existing sessions on password change
Changing a password does not revoke existing refresh tokens (7-day, stateless JWT), so a password change prompted by a suspected compromise does not lock out an already-stolen token.

- **Options:** a per-user token version (or `password_changed_at`) claim validated at `/auth/refresh`, or one-time-use refresh-token rotation.
- Related to the refresh-token-rotation follow-up already noted in #147; worth solving once for both paths.

An-Array:
## Summary

Fixes #149

This PR implements the two backend security hardening tasks requested in the issue: enforcing strict password length validation to prevent bcrypt truncation, and introducing stateful session revocation to instantly invalidate compromised refresh tokens upon a password change.

## Changes

- **Schema Validation:** Added `min_length=8` and `max_length=72` to `ChangePasswordRequest.new_password` in `apps/api/schemas/auth.py`.
- **Session Revocation Tracking:** Added a `token_version` integer column to the `User` model, complete with an Alembic migration using `server_default='1'` to prevent crashes on existing production rows.
- **JWT Payload Injection:** The `create_access_token` and `create_refresh_token` functions now accept and embed the user's current `token_version` into the JWT payload as the `"ver"` claim.
- **Invalidation Logic:** The `PATCH /auth/change-password` endpoint now increments `current_user.token_version`. The `POST /auth/refresh` endpoint was updated to decode the token, check the `"ver"` claim against the database, and return a `401 Unauthorized` if they do not match. (Active frontend sessions will be forcibly revoked the moment their current short-lived access token expires and triggers a refresh attempt).
- **Test Coverage:** Added `test_change_password_invalid_length` and `test_refresh_token_rejected_after_password_change` utilizing the repository's `mock_db` pattern.

## Testing

## Checklist

## Screenshots

N/A (Backend architecture changes only)


ravirajsinh45 (changes requested):
Thanks @An-Array — this is a well-executed answer to #149. The mechanism is right, the migration correctly uses `server_default` so existing rows don't break, and you brought tests. Two things need fixing before this can merge, both about the transition rather than the design.

## 1. As written, this signs out every user on every instance the moment it deploys

`apps/api/routers/auth.py:225` compares strictly:

```python
if payload.get("ver") != user.token_version:
```

Every refresh token already issued was minted before this PR and carries no `ver` claim, so `payload.get("ver")` is `None`. Your migration backfills every existing row to `token_version = 1`. `None != 1` → 401 for 100% of current sessions. Refresh tokens live 7 days and the web client refreshes on each 15-minute access-token expiry, so within ~15 minutes of an upgrade every logged-in user lands on `/login` — and the message they'd see is "Session expired due to password change", which isn't true for any of them.

A token with no `ver` claim is by construction one minted before any version bump could have happened, so reading it as version 1 is correct, not a weakening:

```python
if payload.get("ver", 1) != user.token_version:
    raise HTTPException(status_code=401, detail="Session expired, please log in again")
```

A token minted at `ver: 1` against a user now at `token_version: 2` is still correctly rejected. Worth a test asserting a claim-less token is accepted for a version-1 user.

## 2. Changing your own password logs you out of the session you did it from — silently, minutes later

`change_password` bumps the version and returns just `{"message": "Password changed successfully"}`, so the caller keeps holding tokens minted at the old version. Nothing validates `ver` on access tokens, so the UI looks healthy — then the access token expires, `POST /auth/refresh` 401s, and `clearTokens()` in `apps/web/lib/auth.ts` does `window.location.href = '/login'` with no message. Every password change hits this.

Simplest fix is to hand the caller a fresh pair and have the client keep it:

```python
current_user.password_hash = hash_password(body.new_password)
current_user.token_version += 1
db.commit()
db.refresh(current_user)
return TokenResponse(
    access_token=create_access_token(str(current_user.id), token_version=current_user.token_version),
    refresh_token=create_refresh_token(str(current_user.id), token_version=current_user.token_version),
    needs_password=False,
)
```

plus `setTokens(...)` at the `api.patch('/auth/change-password', ...)` call in `apps/web/app/(dashboard)/settings/profile/page.tsx`. That kills every *other* session while keeping the acting one alive.

## Smaller things

- **Debug code:** `apps/api/tests/test_auth.py:142` — the `if resp.status_code == 500: print("--- 500 ERROR TRACE ---")` block inside the existing `test_refresh_token` should come out.
- **The increment isn't covered.** `test_refresh_token_rejected_after_password_change` sets `mock_user.token_version = 2` by hand and never calls `PATCH /auth/change-password`, so it only tests the comparison in `/auth/refresh`. We checked: deleting `current_user.token_version += 1` from the endpoint leaves the whole suite green. Could you add a test that drives the endpoint and asserts the version incremented? Note `conftest._make_user` doesn't set `token_version`, so set it explicitly — `+= 1` on a bare MagicMock passes vacuously.
- **CHANGELOG:** we track user-facing changes under `## [Unreleased]` in `CHANGELOG.md`. Two entries fit here — the server-side length validation under `### Fixed`, and session revocation on password change under `### Changed`. Happy to draft the wording if useful.
- Trailing newlines are missing on the new migration and `test_auth.py`.

## Not asking you to change these

Noting them so the scope is clear:

- `ver` is only checked at `/auth/refresh`, so an access token stays valid for up to 15 minutes after a password change. That's exactly what #149 specified, and it takes the stolen-credential window from 7 self-renewing days down to 15 minutes — a good trade. Checking `ver` in `get_current_user` too is worth a follow-up on our side.
- `apps/api/routers/setup.py:96-97` still mints tokens without `token_version`, relying on the `= 1` default. Harmless today, but making the parameter required so a missed call site fails loudly is worth doing separately.
- #149's "converge on a shared `_PASSWORD_FIELD`" item belongs to the still-open #147, which covers the other four password fields and not `ChangePasswordRequest` — your change fills the gap it leaves. We checked the merge and the two don't conflict, so this PR closing #149 is fine.

One housekeeping note: CI hasn't run on this yet — fork PRs need a maintainer to approve workflow runs — so none of the above is backed by a CI signal.

An-Array:
Thanks for the incredibly thorough review and explanation. I'll get to work on these fixes.

An-Array:
Hi @ravirajsinh45,
Thanks for your clear and detailed review on #149.
All requested changes have been implemented and pushed!
- Added the `ver` fallback in refresh_token to prevent the mass logout bug.
- Handed fresh tokens back to the frontend on password change to keep the session alive.
- Updated the test suite (added the legacy token test).
Let me know if anything else is needed!

ravirajsinh45 (approved):
Both blocking issues from the previous review are fixed correctly, and I re-verified both independently against the live stack (not just the added tests):

- **Deploy-day mass-logout**: confirmed a token with no `ver` claim (simulating any session that existed before this deploys) still refreshes successfully — `payload.get("ver", 1)` fallback works as intended.
- **Self-logout on password change**: confirmed end-to-end — logged in, changed password using the original access token, the *old* refresh token is now correctly rejected (`401 Session expired`), and the *new* refresh token returned in the change-password response keeps working. Sessions elsewhere are killed, the acting session survives.

Rebased onto current `main` myself to get a real CI signal (this had never had CI run on it — fork PRs need maintainer approval to run workflows, which hadn't happened yet). Two small, non-overlapping conflicts (CHANGELOG.md, test_auth.py — both files had new unrelated entries land on main since this branch was cut); resolved by keeping both sides. All 4 checks are green now. Also dropped a leftover debug `print()` in `test_change_password_increments_version` while resolving the test file conflict — didn't affect anything, just noise.

Migration applies cleanly (`token_version` backfills existing rows to `1` via `server_default`, confirmed against a real Postgres). No other callers of `/auth/change-password` besides the one updated in this PR, so the response-shape change (`{"message"}` → `TokenResponse`) is safe.

One pre-existing, already-noted item, not blocking: `/setup/create-superadmin` still mints tokens via the `token_version=1` default rather than an explicit value — harmless since a freshly-created superadmin's DB row is also `1`, tracked as a separate follow-up per the earlier review.

Thanks for turning this around — nice fix.

PR #191, opened 2026-07-20
fix(share): render comment panel as full-width overlay on mobile view…

### LINKED ISSUE #187: Share viewer: comment panel should be a full-width sheet on mobile, not a fixed 360px column
Follow-up from #177 (which fixed the *default* state of the panel on mobile).

## Problem

In both share viewers the comment panel is a fixed-width, non-shrinking column rendered as a flex sibling of the media viewer:

- `apps/web/app/share/[token]/page.tsx:695` — `w-[360px] flex flex-col border-l ... shrink-0`
- `apps/web/components/share/folder-share-viewer.tsx:924` — `w-[360px] flex flex-col border-l border-border bg-bg-secondary shrink-0`

Both sit inside `<div className="flex flex-1 overflow-hidden min-h-0">` next to a `flex-1` media column.

On a 390px-wide phone, opening the panel gives it 360px that it explicitly refuses to shrink, leaving the player ~30px. #177 stops this from happening on load, but it still happens the moment a guest taps the toggle — which is exactly what they do to leave a comment.

## Suggested fix

Below the `md` breakpoint, render the panel as a full-width overlay/sheet over the media rather than a sibling competing for horizontal space (e.g. `w-full md:w-[360px]` plus absolute positioning below `md`). Apply to both viewers.

## Notes

- Neither share viewer currently has any responsive CSS; this is the first place it is needed.
- Related dead code worth removing while in this area: `FolderAssetViewer` (`apps/web/app/share/[token]/page.tsx:920`) is defined and never rendered.

Vrindakr3300:
## Summary

Fixes #187. On mobile viewports (< 768px), the comment panel in both single-asset and folder share viewers now renders as a full-width overlay sheet over the media player instead of competing for horizontal space as a fixed 360px flex sibling.

## Changes

- **Single-asset share viewer** (`apps/web/app/share/[token]/page.tsx`):
  - Added `relative` positioning to the main flex container.
  - Updated `ShareRightPanel` styling to `w-full absolute inset-y-0 right-0 z-20 md:static md:w-[360px] md:inset-auto`.
  - Removed dead code component `FolderAssetViewer`.
- **Folder share viewer** (`apps/web/components/share/folder-share-viewer.tsx`):
  - Added `relative` positioning to `ShareReviewInner` main flex container.
  - Updated sidebar comment panel container styling to `w-full absolute inset-y-0 right-0 z-20 md:static md:w-[360px] md:inset-auto`.
- **Changelog**: Added entry under `## [Unreleased]` -> `### Fixed`.

## Testing

## Checklist


ravirajsinh45 (approved):
@Vrindakr3300 — thanks for this, and apologies for the wait. Reviewed and approving.

The approach is right: CSS owns the layout, so the panel stops competing for horizontal space below `md` without touching any state logic. Pairs with #177 (which stops it defaulting *open* on mobile) — between the two, a guest on a 390px phone gets full-width media on load and a full-width sheet when they tap to comment.

Things I checked specifically, since they're the non-obvious risks in this diff:

- **Stacking.** Your sheet is `z-20`; the annotation layers are `absolute inset-0 z-10` (`annotation-overlay.tsx:103`, `annotation-canvas.tsx:106`), so the sheet correctly covers them rather than being drawn under.
- **Containing block.** Adding `relative` to the flex row doesn't re-parent the annotation overlays — the image path has its own `relative` wrapper and the video path passes the overlay into `VideoPlayer`. No regression there.
- **Toggle reachability.** The toggle lives in the top bar, which is outside the `relative` container, so the sheet never occludes the control that dismisses it.
- **`FolderAssetViewer` removal.** Confirmed genuinely dead — repo-wide grep returns only the three lines inside its own declaration, no render site, no import, no export. Good catch.

#177 just merged, so I rebased your branch onto the new `main` and pushed it rather than sending it back to you. **Two things changed, flagging both:**

1. **Rebased onto `main`** — your commit applied cleanly, no code conflict.
2. **Added one commit of my own** (`0ddab1c2`) merging your CHANGELOG bullet with #177's into a single entry. The two described the same user-visible problem from different angles, so as separate bullets they'd have read as two unrelated fixes in the release notes. Your code is untouched.

Verified on the rebased branch: `pnpm --filter web test` 256/256, `tsc --noEmit` clean, build compiles.

Not blocking, but for the follow-up: this PR fixes the panel in the two *asset* viewers. The folder **grid** has a third panel with the same defect (`folder-share-viewer.tsx:1581`, `w-[320px] shrink-0`). Filing that separately and offering it to you first — it's your pattern now.

PR #195, opened 2026-07-21
fix(share): render folder grid right panel as full-width overlay on m…

### LINKED ISSUE #193: Share viewer: the folder grid's RightPanel has the same fixed-width mobile defect (320px)
**Body:**

Split out of #187, which #191 fixes for the two *asset* viewers. The folder **grid** — the first screen a guest sees when they open a folder share link — has a third comment panel with the same defect, and #191 doesn't touch it.

## Problem

`apps/web/components/share/folder-share-viewer.tsx:1581`:

```
<div className="w-[320px] shrink-0 border-l border-border bg-bg-secondary flex flex-col overflow-hidden">
```

It's a flex sibling of the asset grid inside `:1351` (`<div className="flex flex-1 overflow-hidden">`), and `panelOpen` defaults to `true` at `:1051` regardless of viewport.

On a 390px phone that leaves the grid ~70px — and the grid's own column count is viewport-based (`grid-cols-2` at `:1083`), so it renders two columns inside those 70px. This is arguably worse than the asset-viewer case #187 covered, because it's the *landing* screen of a folder share rather than something a guest opts into.

## Suggested fix

Apply the same pattern #191 established:

- `:1351` — add `relative` to the flex row.
- `:1581` — `w-full md:w-[320px] absolute inset-y-0 right-0 z-20 md:static md:inset-auto … border-l-0 md:border-l`.
- `:1051` — give `panelOpen` the same desktop-default treatment #177 gave `sidebarOpen`.

## Notes

Best done after #191 merges, so the pattern is already in the tree to copy. @Vrindakr3300 — offering this to you first since it's your pattern; no obligation.

Vrindakr3300:
Fixes #193. Follow-up to #187 and #191.

## Summary

On mobile viewports (< 768px), the comment/info `RightPanel` in the folder grid view now starts closed by default and renders as a full-width overlay sheet over the folder contents when toggled open, matching the pattern established in #191 for the asset viewers.

## Changes

- **`apps/web/components/share/folder-share-viewer.tsx`**:
  - Updated `panelOpen` state default to start closed on mobile viewports (`< 768px`) and open on desktop (`≥ 768px`).
  - Added `relative` positioning to the main folder grid content area container.
  - Updated `RightPanel` container styling to `w-full md:w-[320px] absolute inset-y-0 right-0 z-20 md:static md:inset-auto flex flex-col border-l-0 md:border-l border-border bg-bg-secondary shrink-0 overflow-hidden`.
- **`CHANGELOG.md`**: Updated entry under `## [Unreleased]` -> `### Fixed` to cover the folder grid view.

## Testing

## Checklist


ravirajsinh45 (approved):
Approving — this is exactly right, and fast. Thanks @Vrindakr3300.

Checked the things that could have gone wrong:

- **The toggle stays reachable.** This is the one that would have broken the feature: defaulting `panelOpen` closed on mobile is only safe if there's a way to reopen it. The toggle is at `folder-share-viewer.tsx:1339`, inside the `<header>` that closes at `:1346`, while the new `relative` container starts at `:1349` — so the `z-20` sheet never covers its own dismiss control.
- **Pattern consistency.** The lazy `useState(() => …)` initializer matches #177, and the overlay classes match the ones you established in #191. Three panels, one pattern.
- **CHANGELOG.** You updated the existing bullet instead of adding a second one describing the same user-visible fix — that's the right call, and it's what keeps the release notes honest. Nice.

All four checks green. Merging.

PR #198, opened 2026-07-21
fix(review): clear active annotation and focused comment on version s…

### LINKED ISSUE #186: Version switch doesn't clear the shown comment annotation drawing (single asset view)
## Bug

In the single-asset review view (video or image), clicking a comment that has a drawn annotation sets `activeAnnotation` in the review store, which `AnnotationOverlay` renders on top of the media. Switching versions via the version switcher does **not** clear `activeAnnotation` (or `focusedCommentId`), so the previous version's drawing keeps rendering on top of the newly selected version's media.

## Steps to reproduce

1. Open a video or image asset that has 2+ versions in the review view.
2. Click a comment on the current version that includes a drawn annotation — the annotation overlay renders (driven by `activeAnnotation` in the review store).
3. Open the version switcher (top bar) and select a different version.
4. Observe: the new version's media loads, but the old version's annotation drawing is still rendered on top of it.

## Root cause

`setCurrentVersion` in the review store only updates `currentVersion` — it never resets `activeAnnotation` / `focusedCommentId`:

```ts
// apps/web/stores/review-store.ts:58-60
setCurrentVersion: (version: AssetVersion) => {
    set({ currentVersion: version })
  },
```

`VersionSwitcher` calls this directly with no other cleanup:

```ts
// apps/web/components/review/version-switcher.tsx
onSelect={() => setCurrentVersion(version)}
```

The single-view media viewers render the overlay straight off the store:

```tsx
// apps/web/app/(dashboard)/projects/[id]/assets/[assetId]/page.tsx:299 (video) and :327 (image)
<AnnotationOverlay key={focusedCommentId ?? 'none'} />
```

```ts
// apps/web/components/review/annotation-overlay.tsx
const storeActiveAnnotation = useReviewStore((s) => s.activeAnnotation)
const active = annotation !== undefined ? annotation : storeActiveAnnotation
```

Since no `annotation` prop is passed in single view, `AnnotationOverlay` always reflects the store's `activeAnnotation`, which nothing clears on a version change. The `key={focusedCommentId ?? 'none'}` doesn't help either, since it's scoped to comment focus, not version. The only place that clears this state today is `ImageViewer`'s click-off-media handler (`setFocusedCommentId(null); setActiveAnnotation(null)` in `handleImageClick`) — not a version-change path — and there's no equivalent for video at all.

## Suggested fix

Clear `activeAnnotation` / `focusedCommentId` in `setCurrentVersion` when the version actually changes, guarding the initial `null -> version` set and re-selecting the same version so a URL-driven deep-link annotation isn't wiped on load:

```ts
setCurrentVersion: (version: AssetVersion) => {
  set((state) =>
    state.currentVersion != null && state.currentVersion.id !== version.id
      ? { currentVersion: version, activeAnnotation: null, focusedCommentId: null }
      : { currentVersion: version },
  )
},
```

(Not applicable to the version-compare view, which manages annotation state per-pane via props rather than through the shared store.)

Vrindakr3300:
Fixes #186.

## Summary

In the single-asset review view, clicking a comment with a drawn annotation sets `activeAnnotation` and `focusedCommentId` in the `review-store`. Previously, switching versions via the version switcher did not clear these properties, causing drawings from the old version to linger over the newly selected version's media.

`setCurrentVersion` now resets `activeAnnotation` and `focusedCommentId` when changing to a different version, while preserving state on initial load and same-version re-selection.

## Changes

- **`apps/web/stores/review-store.ts`**: Updated `setCurrentVersion` to reset `activeAnnotation: null` and `focusedCommentId: null` when switching to a different version ID.
- **`apps/web/stores/__tests__/review-store.test.ts`**: Added unit tests covering version change cleanup and same-version preservation.
- **`CHANGELOG.md`**: Added entry under `## [Unreleased]` -> `### Fixed`.

## Testing

## Checklist


ravirajsinh45 (approved):
Thanks for this, @Vrindakr3300. Clean, well-scoped fix that matches #186's own root-cause analysis exactly: activeAnnotation/focusedCommentId were the only pieces of review-store state a version switch left behind, and the guard for initial load and same-version re-selection (so a URL-driven deep-linked annotation doesn't get wiped on first render) is correct.

Both branches of the conditional are directly unit-tested, which is the right level for a pure store-reducer change like this.

Rebased onto current main myself and pushed to your branch: your CHANGELOG entry landed under the [1.7.0] section in your diff since that's where [Unreleased] was when you branched, but main has moved through v1.7.1-v1.7.4 since then. Moved it to the current [Unreleased] section, the only real conflict.

Verified: full frontend suite (263 tests, including your 2 new ones), typecheck, and build all pass. CI is green.

PR #199, opened 2026-07-21
Feat stub matchmedia vitest

### LINKED ISSUE #188: web tests: stub window.matchMedia in vitest setup
**Scope note (2026-07-21):** narrowed — please read this before starting.

The `matchMedia` breakpoint code this issue refers to is **not on `main`**; it comes from #177, which is still open. So the panel regression test described below can't be written yet.

**In scope for this issue:**
1. The width-driven `window.matchMedia` stub in `apps/web/test/setup.ts`, exporting a `setViewportWidth(width)` helper.
2. A test locking that stub's own contract — `(min-width: 768px)` false at 390 / true at 1024, the `>=` boundary at exactly 768, and a `change` listener firing when the width crosses it.

**Out of scope (for now):** the panel regression test — it lands with or after #177.

Three requirements the original text didn't mention:
- Handle `(prefers-color-scheme: …)` **explicitly** rather than letting it fall through to `false`. Every existing `matchMedia` caller in the repo is a colour-scheme query, not a width one: `stores/theme-store.ts:19`, `components/shared/theme-initializer.tsx:28` (which subscribes via `addEventListener`), and `components/share/folder-share-viewer.tsx:1102`.
- Expose the full surface, including the legacy `addListener` / `removeListener` — `theme-initializer.tsx` will throw without it.
- Keep the property **configurable**, so a future per-test `vi.stubGlobal('matchMedia', …)` can still override it.

The optional `useIsDesktop()` hook extraction is **on hold until #177 merges** — we'll ping you here the moment it's unblocked. Not silence, just sequencing.

---

Follow-up from #177.

## Problem

`window.matchMedia` is not implemented by jsdom, and `apps/web/test/setup.ts` is a single line (`import "@testing-library/jest-dom/vitest"`) that does not polyfill it. Verified by running a throwaway test under `pnpm exec vitest run` in `apps/web`: `typeof window.matchMedia` is `undefined`.

Consequences:

- Components that read `matchMedia` during render cannot be tested. `apps/web/app/share/[token]/page.tsx:838` now does this, so any future test rendering `ShareViewer` fails with `TypeError: window.matchMedia is not a function` before asserting anything.
- The mobile share-panel behavior from #177 therefore ships without a regression test.

CI is green today only because no existing test renders that page.

## Suggested fix

Add a `window.matchMedia` stub to `apps/web/test/setup.ts` — ideally driven by a settable viewport width so tests can assert both sides of a breakpoint:

```ts
Object.defineProperty(window, "matchMedia", {
  writable: true,
  value: (query: string) => ({
    matches: false, // or derived from a test-controlled width
    media: query,
    onchange: null,
    addListener: () => {},
    removeListener: () => {},
    addEventListener: () => {},
    removeEventListener: () => {},
    dispatchEvent: () => false,
  }),
})
```

Then add a regression test asserting the share comment panel is absent below 768px and present at/above it. `apps/web/components/review/__tests__/version-switcher.test.tsx` is a good structural precedent.

## Related

While here, consider extracting a shared `useMediaQuery` / `useIsDesktop` hook — the breakpoint is currently duplicated as a literal in component state initializers, and the repo has no such hook (the only other `matchMedia` uses are `components/shared/theme-initializer.tsx:28` inside an effect and an inline pre-hydration script in `app/layout.tsx`).

solankimeet518:
Fixes #188 

## Summary

- Added `window.matchMedia` stub and `setViewportWidth` helper in `apps/web/test/setup.ts`.
- Added unit tests in `apps/web/test/setup.test.ts` verifying boundary conditions (`390px`, `768px`, `1024px`).

## Testing

## Checklist


ravirajsinh45 (approved):
Thanks for this, @solankimeet518. The width-driven stub and `setViewportWidth` helper are exactly the right shape, and the boundary test is solid. Two things from #188's own scope note weren't covered yet, so I fixed them and pushed to your branch:

**`prefers-color-scheme` wasn't handled explicitly.** Every real `matchMedia` caller besides the 768px breakpoint (`theme-store.ts:19`, `theme-initializer.tsx:28`, `folder-share-viewer.tsx:1099`) is a color-scheme query, and the stub was answering it from the viewport-width comparison. Since `test/setup.ts` is the global setup file, every test in the suite got `matches: true` for `(prefers-color-scheme: dark)` at the default 1024px width, regardless of what that test actually wanted. Now it's explicitly short-circuited to `false` before the width logic ever runs.

**The `change` listener was a no-op.** `addEventListener`/`addListener` didn't actually register anything, so nothing that subscribes to a live breakpoint change (the exact scenario #188 asks for a test on) was actually testable. Registrations are now tracked per query and fire `change` when `setViewportWidth()` crosses that query's boundary.

Generalized the width matching to parse the actual px value out of the query rather than hardcoding 768, so it keeps working if another breakpoint gets added later. Kept your `configurable: true` and the legacy `addListener`/`removeListener` surface, both already correct.

Added tests for both fixes plus the `change`-listener requirement from the scope note. Verified full frontend suite, typecheck, lint, and build all pass.

Thanks again for taking this on.

PR #202, opened 2026-07-22
fix: security audit fixes (H1-H3, M1-M9, L1-L3, C1-C2) — rebased

rubenxyz:
## Summary

Independent security audit of FreeFrame. Each finding has a dedicated fix; commit messages contain the finding ID, severity, and rationale.

**Rebased on upstream `main`** (v1.7.0) to resolve all conflicts from merged PRs (#93, #177, #191, #156, etc.).

## Changes from previous review (#147)
- **B1** ✅ Resolved — `generate_presigned_put_url` now exists upstream (#93)
- **B2** ✅ Fixed — added `notifications` field to `PreferencesUpdate` (M8)
- **B3** ✅ Fixed — `GuestCommentList`/`GuestCommentInput` forward `share_session` (C2)
- **B4** ✅ Fixed — `/verify` endpoint checks `visibility == "secure"` gate (H3)
- **S5** ✅ Fixed — `min_length=8` removed from `LoginRequest` (M1); kept on creation schemas
- **S6** ✅ Fixed — `verify-magic-code` returns single `"Invalid or expired code"` detail
- **SECURITY_AUDIT.md** dropped from the PR per maintainer request

### Critical
- **C1** — `invite_token` leaked via `UserResponse` to any-authenticated-user endpoints. Added admin-scoped `AdminUserResponse`.
- **C2** — share comment endpoints now use `validate_share_link_with_session`. Frontend `GuestCommentList`/`GuestCommentInput` forward `share_session`.

### High
- **H1** — send-magic-code no longer auto-creates accounts or sends emails to unknown addresses.
- **H2** — startup warning when docs+OpenAPI schema enabled on non-localhost FRONTEND_URL.
- **H3** — `POST /share/{token}/verify` with password in body (not query string). Secure-visibility gate added.

### Medium
- **M1** — password `min_length=8` on creation schemas; `LoginRequest` gets `max_length=72` only.
- **M2** — dropped public-read bucket policy on `processed/*` (dead weight — HLS proxy uses presigned URLs).
- **M5** — attachment presign uses `generate_presigned_put_url` (public endpoint) + sanitized filename.
- **M6** — attachment GETs pass `download_filename` for `Content-Disposition: attachment`.
- **M7** — strip path from frontend_url when setting bucket CORS origins.
- **M8** — `PreferencesUpdate` Pydantic schema with `theme` + `notifications` keys.
- **M9** — rate limit on `/auth/refresh` (30 req/min).

### Low
- **L1** — escape LIKE patterns in users/me search.
- **L2** — tighten bucket CORS headers/methods (drop `DELETE`, restrict `AllowedHeaders`).
- **L3** — basePath-aware `/login` redirect on the frontend.

## Not included (maintainer request)
- `SECURITY_AUDIT.md` — operational details belong in private channel
- M3/M4 — already addressed upstream by merged PRs (#90/#93 context)


rubenxyz:
Rebased on v1.7.0 (d5611f9) — all previous review blockers from #147 resolved:

- C1: invite_token dropped from public UserResponse; admin-scoped AdminUserResponse added
- C2: share comment endpoints now enforce password via validate_share_link_with_session
- H1: send-magic-code no longer auto-creates accounts or sends to unknown emails
- H2: startup warning when docs enabled on non-localhost frontend_url
- H3: POST /share/{token}/verify for password (body, not query string); frontend migrated
- M1-M9, L1-L3: all addressed per the original review

Ready for re-review. Let me know if anything needs adjustment.

ravirajsinh45:
@rubenxyz there ci check failing can you please verify and resolve it ?

rubenxyz:
Rebased onto latest main (was conflicting) and CI is green: https://github.com/Techiebutler/freeframe/actions/runs/31895833383

Notes on the rebase:
- Several findings from this PR are now already fixed upstream (#208 invite-token leak/registration removal, #212 Medium/Low closes incl. magic-code enumeration, comment visibility, mention access checks, version_id validation) — I took upstream's implementations for those rather than reintroducing duplicate code
- Kept the parts that are still missing upstream: LIKE-escaping in user search, typed PreferencesUpdate (replaces `body: dict`), refresh-token rate limit, password constraints (8-72 chars, bcrypt truncation), CORS tightening, removal of the public-read bucket policy
- Fixed a rebase artifact: `shareSession` wasn't threaded through ShareViewer -> ShareRightPanel (TS2304 broke the frontend build)

Verified locally (next build, tsc, vitest, pytest) before pushing.

ravirajsinh45:
@rubenxyz thanks for sticking with this one through three rebases. We did a full review pass and it holds up well; most of what we found is "the fix is right but does not cover every call site" rather than anything wrong in the approach.

Rather than send it back for another round trip, we pushed two follow-up commits to this branch (`d1812b26` and `84af9c85`). Your commits are untouched. Everything we changed is listed below so nothing is a surprise, and please push back on any of it.

We also merged this branch together with #203 locally and ran the result against a live stack, since the two overlap in the share page. Findings from that are at the end.

## The one that matters

**A `secure` link was still readable without logging in.** The visibility gate is checked in `GET /share/{token}` and `POST /share/{token}/verify`, but the six endpoints that actually serve content resolve the link through `validate_share_link_with_session()`, which only looked at `password_hash`. So an anonymous caller holding the token could still hit `/stream/{asset_id}`, `/thumbnail/{asset_id}`, `/assets/{asset_id}/versions`, the folder listing, and both guest comment endpoints, which is exactly what your comment on the verify handler warns about.

We moved the check into `validate_share_link_with_session()` so all six inherit it, and left the GET endpoint's non-raising `requires_auth` response as-is since that is the right UX for the initial validate. There is a new `apps/api/tests/test_share_secure_visibility.py`; three of its cases fail against the previous code and pass now.

We also confirmed it on a running instance rather than only in tests: against a `secure` link with the gate removed, an anonymous request to `/share/{token}/assets` returned `200` and the real asset listing. With the gate in place it returns `403`, and an authenticated request still returns `200`.

## The one we found while testing that

**A "Secure" link was never actually secure, because the setting was dropped on creation.** `ShareLinkCreate` accepts `visibility` and `share-create-dialog.tsx` sends it, but three of the four `ShareLink(...)` constructions in `share.py` never passed it through. Only the multi-share path did. We reproduced it live: `POST /projects/{id}/share` with `{"visibility": "secure"}` came back `"visibility": "public"`.

This one predates your branch, it is on `main` today, so it is not something you introduced. We fixed it here (`84af9c85`) rather than separately because it is the other half of the finding above: enforcing the gate everywhere only helps if the setting can be turned on, and until now the only way to get a secure link was to create it and then PATCH it. An admin who picked "Secure" in the dialog got a public link and no indication otherwise.

The test walks every `ShareLink(...)` construction in the module rather than asserting on one call site, so a fifth one cannot quietly repeat it. End to end on the live stack afterwards: create with `visibility: secure` returns `secure`, anonymous gets `403`, authenticated gets `200`.

## Also fixed

- **bcrypt truncation was still reachable.** `Field(max_length=72)` counts characters, but bcrypt truncates at 72 *bytes*. We reproduced two distinct 46-character passwords that verify against the same hash while passing validation. Length is now measured on the UTF-8 encoding, and hashing encodes before truncating. We checked this against stored hashes first: bcrypt already truncated internally, so the change is hash compatible and nobody gets locked out. This also applied to the seven share-link password sites in `share.py`, so they now share one helper.
- **The CORS origin strip was a no-op.** `_cors_origin` and `_allowed_origins` were computed and then discarded, with the raw `frontend_url` used for `AllowedOrigins` two lines later. Sub-path deployments would still fail CORS. It is now exposed once as `settings.frontend_origin`.
- **Collection smart filters were not escaped.** `me.py` and `users.py` got the LIKE escaping, but `metadata.py`'s `name_contains` did not, so a rule of `%` matches every asset in the project. The helper existed in three copies, which is why the fourth site was missed; it now lives in `services/search.py`.
- **`Content-Disposition` was built from unvalidated input.** A comment attachment's `original_filename` is whatever the uploader sent. It is now stripped of control and bidirectional characters (an RLO can make `invoice<RLO>gnp.exe` display as `invoice exe.png`), separators are flattened, length is capped, and non-ASCII names get RFC 5987 `filename*`.
- **Preferences.** `notifications: dict` accepted arbitrary nested JSON despite the docstring promising primitives, and `extra="ignore"` returned 200 for keys it silently dropped. Now bounded, and unknown keys are rejected.
- **Password-protected downloads were broken.** `shareSession` was threaded into the guest comment flow but not into the stream or download calls, so Download returned a silent 403 after a correct password. Same for the stream URL fallback.
- **`log_open` never fired for password-protected links.** The initial GET short-circuits at `requires_password` and drops the flag, and the follow-up verify always sent `log_open: false` because a password was present, so those links never recorded an open. The flag is now set only once a call actually returns the full response.
- **basePath.** The prefix was applied in `clearTokens()` only. We added it to the other raw navigations (middleware redirects, the share page sign-in anchor, the folder viewer's log out and back buttons) via a small `withBasePath()` helper. Deliberately *not* applied to `router.push`/`<Link>` targets, so this stays correct if `basePath` is ever set in `next.config.js`.
- **`useAuth().logout()`** fired `router.push('/login')` on top of the full-page navigation `clearTokens()` already performs. Removed the duplicate.
- `ChangePasswordRequest` and the superadmin setup schema now reuse the shared password field instead of restating it.

## Where we disagreed with our own review

- **The generic magic-code error is correct, so we kept it.** It looked like a lost "too many attempts" signal, but a code is only ever issued for a real account, so surfacing that case would confirm which addresses are registered. We only reworded the shared message so a rate-limited user knows to request a new code, and pointed the test at the constant so the two raise sites cannot drift.
- **Docs exposure.** We left `DISABLE_DOCS` warn-only. Flipping the default is a behaviour change for self hosters and that is your call, not something to slip into a review commit.

## Two open questions

1. `LoginRequest.password` had `max_length=72`. Anyone whose existing password is longer would get a 422 instead of authenticating, even though verification truncates to the same bytes the hash was made from. We relaxed it to a sanity bound so existing accounts keep working, and kept the hard ceiling on the paths that *set* a password. Shout if you meant the stricter behaviour.
2. `next.config.js` sets no `basePath`, so `NEXT_PUBLIC_BASE_PATH` is only a convention that each raw navigation has to honour, and static assets would not be served under the sub-path anyway. Wiring `basePath` properly is a bigger change and probably its own PR. Worth an issue?

## Merging with #203

The two branches touch the same share page, so we merged them locally and ran the result. Two conflicts, and one is worth flagging to whoever merges second: **this branch's H3 fix loses to #203 unless the conflict is resolved deliberately.** #203 still carries the older `fetchShareInfo` that puts the password in a query string, so taking its side of that hunk silently reintroduces the leak into nginx logs, history and Referer headers that H3 closed. Resolved toward this branch, and verified afterwards that no query-string password remains and `POST /verify` is still the path used.

The other conflict was just the CHANGELOG, both entries kept.

## One thing worth knowing about the CORS change

The tightened bucket CORS is a no-op on MinIO, which is the default backend. MinIO does not implement `PutBucketCors` at all: we tried both the old and the new rule sets against it and both fail identically with `NotImplemented`, so this predates your change and is handled fine by the existing warning that tells the operator to configure CORS themselves. Just worth knowing the hardening only actually lands on backends that implement the call. The origin strip still matters for those.

## Verification

Run the way CI does: `306` backend tests pass against a real Postgres, `tsc --noEmit` clean, `319` frontend tests pass, `next build` exits 0. Merged with #203: `333` backend tests, build clean.

Beyond the suites, on a live stack: the secure-link gate refuses anonymous callers on all six endpoints and admits authenticated ones, a password-protected link issues a session and then serves content with it, and the "opened" activity is recorded for a password link for the first time.

ravirajsinh45:
@rubenxyz both of these are merged, thank you for the persistence through all the rebases.

Before merging I pushed two commits to each branch rather than sending them back for another round trip. Everything is disclosed in the comment above on each PR, but the short version:

- On #202: the `secure` visibility gate was only enforced on two endpoints, so a secure link was still readable anonymously through the other six; and `visibility` was being dropped on creation, so choosing "Secure" in the dialog produced a public link. That second one predates your branch and is on `main` today, but it is the other half of the same problem, so it went in here.
- On #203: the S3 delete ran before the commit on both paths, a few malformed inputs returned 500 rather than 422, and the emails were only half white-labelled since six of the eight senders never carried an org name.

If any of that looks wrong to you, say so and I will change it. A contributor finding their diff quietly altered is worse than whatever it fixed, so please do push back.

Testing the two together on a live instance also turned up a set of things that were not review findings on either PR: the accent colour and the custom favicon both did nothing at all, and share links ignored their own theme and accent on the single-asset viewer. Those are in #264, which is mostly your branding feature, so a review there would be genuinely valuable.

Also filed #265 and #266 from the same session, both in the share-link area if you are interested.

rubenxyz:
Thanks for the thorough pass and for pushing the follow-ups — all of it reads right to me. The secure-link find is exactly why doing it as commits rather than another round trip was the right call, and the visibility-drop-on-creation one being on main today makes me glad it went in here.

On the two questions:

1. **LoginRequest max_length** — agreed on the relaxation. The limit was meant to match verification, but since bcrypt truncates to the same bytes on verify, rejecting longer passwords only breaks existing accounts. The hard ceiling stays where it matters — the paths that *set* a password.

2. **basePath** — yes, worth an issue. We run FreeFrame under a sub-path in production, so the convention-only approach keeps catching us in real deployments. I will file one for proper `basePath` wiring via `next.config.js` (static assets + raw navigations) as its own piece of work.

Nothing else to push back on from my side. Thanks again for the persistence through all the rebases.

PR #203, opened 2026-07-22
feat: Instance branding — white-label logos, org name, favicon (rebased)

rubenxyz:
## Summary

Server-backed instance branding with superadmin-gated uploads, custom logos,
org name, favicon, and a "Powered by FreeFrame" toggle.

**Rebased on upstream main (v1.7.0).** All conflicts resolved:
- Share page: `FolderAssetViewer` deletion accepted (#191), branding hooks grafted onto mobile-aware sidebar (#177)
- Sidebar: custom logo + PoweredByBadge merged with upstream storage indicator
- FolderShareViewer: branding store import added alongside upstream review imports

## What this does
- New `instance_branding` table (singleton, auto-seeded on first GET)
- `BrandingTab` inside Admin dashboard (`Users | Instance settings | Branding`)
- 6 logo slots: light, dark, favicon, apple-icon, login-logo + quick-upload "Replace All"
- `BrandingHead` component — dynamic `<title>`, `<link rel="icon">`, Apple touch icon
- Login, sidebar, share page, video player all respect instance branding
- Fallback chain: API → localStorage cache → hardcoded defaults
- Resend email sender/subject still uses "FreeFrame" (documented as follow-up)

## Review items addressed (from #90)
- ✅ **B1** — Migration re-parented to `54b1ad156f8f` (single head)
- ✅ **B2+B3** — Logo upload uses `generate_presigned_put_url` (public endpoint) + `content_type` query param
- ✅ **Rebase** — sidebar, share page, FolderShareViewer conflicts surgically resolved
- ✅ **Remove logo** — uses `TYPE_MAP` lookup, not broken `.replace()` chain
- ✅ **PUT nulls** — `exclude_unset=True` so `null` values can clear logos
- ✅ **Singleton dedup** — `_SINGLETON_ID` sentinel PK + `IntegrityError` retry
- ✅ **Brand flash** — logo URLs cached in `partialize` for localStorage hydrate
- ✅ **PoweredByBadge** — defaults to `text-text-tertiary` (theme-aware)
- ✅ **Login layout** — no duplicated `min-h-screen` wrappers
- ✅ **Route writes** — use shared `api` helper (token-refresh)
- ✅ **Login default** — kept `'email'` (magic-code first), not `'classic'`
- ✅ **Branding in Admin** — third sub-tab, not standalone nav item
- ✅ **`__import__('re')` → `import re`** at module top

## Not in this PR (product direction / follow-up)
- `primary_color` stored/validated but not wired to CSS — follow-up
- Email subjects still hardcoded "FreeFrame" — follow-up
- `branding-preview.tsx` added but not imported — kept for future live preview


rubenxyz:
Rebased on v1.7.0 (d5611f9) — all previous review blockers from #90 resolved:

- Upload flow: now uses generate_presigned_put_url (public endpoint), content_type forwarded from frontend, ContentType nullable when omitted (S3 accepts any MIME)
- Migration re-parented to current main head (single head, no branch)
- Remove logo: explicit TYPE_MAP replaces broken .replace() chain
- PUT null-clearing: exclude_unset (not exclude_none) so explicit nulls clear values
- Singleton dedup: _SINGLETON_ID sentinel PK + IntegrityError guard
- Brand flash: logo URLs in partialize persist to localStorage
- Routes use shared api helper (401 token-refresh + error handling)
- Login-default flip split out to separate PR (#146, merged)
- All other should-fixes from the original review addressed

Ready for re-review. Let me know if anything needs adjustment.

ravirajsinh45:
@rubenxyz there ci check failing can you please verify and resolve it ?

rubenxyz:
Just checking in — I pushed the CI fixes you flagged on the 23rd for both this and #202. Let me know if you need anything else from my side. No rush!

ravirajsinh45:
> Just checking in — I pushed the CI fixes you flagged on the 23rd for both this and #202. Let me know if you need anything else from my side. No rush!

ci still getting fail can you please check? let me know if you need any help on it

rubenxyz:
All checks are green now: https://github.com/Techiebutler/freeframe/actions/runs/31895620232

Root causes of the CI failures, all fixed:
- Frontend parse error: duplicate `import { cn }` in `folder-share-viewer.tsx` (not the sidebar `</div>` the earlier commit targeted)
- Broken JSX in `sidebar.tsx` from the earlier rebase fix (orphan `</nav>`, mismatched `</div>`) — rebuilt the block cleanly
- Backend 500s: singleton row was created with `org_name`/`powered_by_freeframe` = None pre-flush, so `model_validate` failed; now seeded with defaults, and the delete-logo test uses a real model row
- Alembic: branding migration was chained to v1.7.0's head (54b1ad156f8f) while upstream has since added migrations — re-pointed `down_revision` to the current head (a9ee0209151a), fixing 'Multiple head revisions'

Verified locally (next build, tsc, vitest, pytest) before pushing.

rubenxyz:
Rebased onto latest main (v1.9.0, 17db330) and CI is green: https://github.com/Techiebutler/freeframe/actions/runs/31934591727

Two things worth noting:
- The branch was still based on v1.7.0 — I've rebased it onto current main so the alembic migration chains onto the current head (b2c4d6e8f0a1). If main gains another migration before this is merged, alembic will need one more down_revision bump — happy to do it.
- Also restored a small regression: the sidebar Settings link now points non-admins to /settings/appearance again (upstream #168 conditional was lost in an earlier conflict resolution).

ravirajsinh45:
@rubenxyz this is in good shape, and thanks for the patience across the rebases. Folding branding into the Admin tab landed the way we hoped, and the presign rework from the #90 round is clearly better.

Same approach as on #202: rather than another round trip we pushed two follow-up commits to this branch (`a6ba4f71` and `06a499f4`), with your commits untouched. Full disclosure of every change below.

We also merged this branch together with #202 locally and exercised the result against a live stack, so the notes at the end are from real behaviour rather than reading alone.

## API

- **The S3 delete happened before the commit.** Both the update loop and the reset path deleted the replaced object first and committed after. If the commit then failed, the row rolled back to the old key whose object was already gone, leaving a permanently broken logo. Both paths now commit first and delete after; an orphaned object is much better than a dangling reference.
- **Bad input returned 500 instead of 422.** Four separate cases: an explicit `{"org_name": null}` reached the NOT NULL column (`exclude_unset` keeps explicit nulls); a non-string `org_name` hit `len()` and raised `TypeError`, which pydantic does not convert; `primary_color: 7` did the same through `re.match`; and `"#7c3aed\n"` passed the regex because `$` matches before a trailing newline, then overflowed the `String(7)` column. Also `"   "` passed the length check. All now raise clean validation errors, and an omitted field is still omitted.
- **`GET /instance/branding` was writing.** It is unauthenticated and hit on every login and share page load, but created and committed the singleton row, unlike the sibling instance-settings GET that documents itself as read-only. It now returns defaults without persisting; PUT still creates the row.
- **`LOGO_CONTENT_TYPES` was dead.** Only its key set was used, because the presign correctly signs with the client-supplied content type now. Removed, so it cannot mislead the next person; the whitelist check uses `LOGO_TYPES`.
- **Three inline copies of the superadmin check** replaced with the existing `require_admin` dependency, which instance settings already uses for the same kind of endpoint. We verified the predicate is identical before swapping, so this is not a permission change. The 403 detail string becomes "Admin access required".
- **Emails were still half-branded.** The subject line picked up `org_name`, but `base.html` hardcoded the product name in the header and twice in the footer, and `magic_code.html` said "sign in to FreeFrame" in the body while ignoring the `org_name` already being passed to it. See the section below, since this turned out to run deeper than the templates.

## Emails are now genuinely white-label

This was the largest remaining gap, so we finished it rather than leaving it as a follow-up.

Of the eight send tasks, only magic-code and invite ever carried an `org_name`. The other six (mention, comment, assignment, share, approval, project_added) never passed one, so `render_template` fell back to the hardcoded default and every one of them rendered the shared header and footer as FreeFrame no matter how the instance was branded. On top of that, `share.html` hardcoded the product name twice in its own body, and `MAIL_FROM_NAME` put it on the From line of *every* message, which is the part a recipient actually sees in their inbox before opening anything.

Threading a name through six routers would have been a lot of edits and one more thing to forget on the next email type, so we resolved it centrally instead:

- New `services/branding_service.resolve_org_name()` reads the instance row with a 60 second cache. Emails render in the Celery worker, which has no request-scoped session, so this is where the lookup belongs. It falls back to the default if the row or the database cannot be read, because a missing brand name must never be why an email fails to send. The branding PUT drops the cache, so a rename lands on the next email rather than up to a minute later.
- `render_template` now fills `org_name` for every template, exactly the way it already did for `year`. A name passed explicitly still wins, so the invite flow is unchanged.
- `MAIL_FROM_NAME` now defaults to blank, meaning "follow the instance name". Setting it still pins a fixed display name, so anyone relying on that keeps it. `.env.example` and `docs/deployment.md` no longer ship it pre-set to the product name, which would otherwise have quietly defeated the whole thing on a fresh install.
- We dropped the `Organization:` line from `project_added.html`. It was guarded on an `org_name` that no caller ever passed, so it has never rendered; now that the name is always present it would have started appearing and duplicating the header. Removing it keeps that email exactly as recipients see it today.

`apps/api/tests/test_email_branding.py` renders all eight templates and fails on any that reintroduces a hardcoded brand, and also covers the From line precedence and the database-unavailable fallback. We checked it actually bites by reverting one template and watching it fail. On the live stack with the org name set to "Acme Studio", all eight render the instance brand with no occurrence of the product name anywhere, and the From line resolves to "Acme Studio" from the database.

**Worth an upgrade note when this ships.** Anyone whose `.env` came from the old `.env.example` has `MAIL_FROM_NAME=FreeFrame` pinned, and a pinned value deliberately wins over branding, so their From line stays FreeFrame until they blank it. New installs get the right behaviour by default. We noticed because the dev `.env` we tested with had exactly that and the From line did not change until we cleared it.

## Web

- **`.ico` favicon uploads were failing with a 403.** The presign signs with `file.type || 'image/png'` but the PUT sent bare `file.type`, and browsers report `''` for `.ico`, which the favicon slot explicitly advertises. Since the content type is inside the SigV4 signature, the mismatch is a `SignatureDoesNotMatch` surfaced as a generic "Failed to upload file". One value now feeds both, in both uploaders.
- **Quick Upload reported the wrong URL for four of five slots.** `lastUrl` used a fixed-priority `||` chain, and the response carries every key currently set, so `logo_light_url` won every iteration after the first; that single URL was then fanned out to all five store setters. Each slot now reads back its own field. Related: the callback sat after the loop inside the `try`, so a failure on slot three discarded the two slots that had already committed server side. Slots now apply as they succeed.
- **Reset failures looked like success.** `handleResetAll` swallowed the error in an empty catch, so the confirm dialog closed and the admin reasonably assumed the reset worked when nothing had changed. It now surfaces the error and the dialog stays open. This needed one small prop on `ConfirmDialog`; the other three call sites are unchanged.
- **The Reset section was hidden** when only the colour or the attribution toggle had changed, leaving no way to revert them. Reset also now restores `powered_by_freeframe`. Since the toggle counts toward `hasCustomBranding`, leaving it out of the payload meant the Reset section stayed on screen after a reset that had already finished. The dialog copy now says the badge comes back rather than changing it silently, which was our reason for hesitating on it.
- **Persisted logo URLs could be expired.** They are presigned for an hour and were persisted with no timestamp, so reopening the app later rendered broken images before the refetch landed, with no `onError` fallback on the sidebar or favicon. We kept them persisted, because that was your brand-flash fix from #90, and added a timestamp so anything past 45 minutes is dropped on rehydration.
- **Logo variant ignored the `system` theme.** The comparison was against the raw stored preference, so a system-theme viewer on a dark-rendered page got the light-background logo. Now resolved through a shared hook.
- **`/setup` and `/invite/[token]` lost their logo entirely**, because the shared `(auth)` layout header was removed and the replacement was added only inside the login page. Restored in the shared layout, so the login page no longer renders it twice. Visible consequence: on `/login` the mark sits above the card again rather than inside it.
- **`BrandingPreview` was never mounted.** An 85-line component that nothing imported. Wired into the tab rather than deleted, since it reads exactly the fields that tab edits.
- Smaller ones: the per-slot Remove button ignored the admin `disabled` prop and fired a real DELETE for non-admins; the file input's `accept` was built from display labels like `"ICO,PNG"`, which browsers ignore entirely; the share page title read `orgName` imperatively with the wrong deps and stayed on the default for the whole visit; the share top bar's logo cascade did not match the password gate one screen earlier; `/settings/branding` still rendered as an orphan route and now redirects into the Admin tab; and `resetAll()` was exported as a client-only reset that would silently un-reset on the next fetch, so it is gone.

## Deliberately left alone

- **The sidebar and preview logo fallback** does not include `loginLogoUrl`, and we think that is correct, since that slot is specifically the login screen logo. Flagging it in case you meant otherwise.
- **The Branding nav item stays removed.** That consolidation was the agreed direction, not an oversight.

## Follow-ups we did not fold in

Reuse items only, none of which block merge: Quick Upload duplicates the single-slot upload flow and issues five sequential round trips where the schema already accepts all five keys at once; `fetchBranding` hand-rolls `fetch` instead of using the shared client, so it misses the 401 refresh path; the singleton get-or-create is duplicated from instance settings; and `_enrich_branding_response` re-signs up to five URLs on every unauthenticated page load with no caching. Happy to open an issue for these if you would rather not expand this PR.

## Merging with #202

The two branches touch the same share page. Two conflicts, and one needs a careful hand: **this branch still carries the older `fetchShareInfo` that sends the share password as a query string**, which #202 replaced with a POST body precisely because a query string lands in nginx logs, browser history and Referer headers. Whichever of the two merges second, that hunk has to resolve toward #202 or the leak comes back silently. Worth a rebase on this branch once #202 lands, rather than leaving it to the merge.

The other conflict was just the CHANGELOG, both entries kept.

## Verification

Run the way CI does: `326` backend tests pass against a real Postgres (up from 312, the difference being the new email branding suite), a single Alembic head, 108 API paths, `tsc --noEmit` clean, `319` frontend tests pass, `next build` exits 0. Lint warnings are byte identical to the pre-change baseline. Merged with #202: `333` backend tests, build clean.

On a live stack we also confirmed the parts that are hard to unit test: an anonymous `GET /instance/branding` returns defaults and leaves the table empty, all five malformed payloads return 422 rather than 500, a presigned favicon upload succeeds when the content types match and fails with exactly the `SignatureDoesNotMatch` this fixes when they do not, reset restores the attribution badge, and `/setup` and `/invite/[token]` render with the brand mark again.

One cosmetic thing we did not change: the server-rendered `<title>` is still the product name until hydration, since `app/layout.tsx` sets it statically and `BrandingHead` corrects it on the client. Pre-existing, and fixing it properly means making the layout metadata dynamic.

ravirajsinh45:
@rubenxyz both of these are merged, thank you for the persistence through all the rebases.

Before merging I pushed two commits to each branch rather than sending them back for another round trip. Everything is disclosed in the comment above on each PR, but the short version:

- On #202: the `secure` visibility gate was only enforced on two endpoints, so a secure link was still readable anonymously through the other six; and `visibility` was being dropped on creation, so choosing "Secure" in the dialog produced a public link. That second one predates your branch and is on `main` today, but it is the other half of the same problem, so it went in here.
- On #203: the S3 delete ran before the commit on both paths, a few malformed inputs returned 500 rather than 422, and the emails were only half white-labelled since six of the eight senders never carried an org name.

If any of that looks wrong to you, say so and I will change it. A contributor finding their diff quietly altered is worse than whatever it fixed, so please do push back.

Testing the two together on a live instance also turned up a set of things that were not review findings on either PR: the accent colour and the custom favicon both did nothing at all, and share links ignored their own theme and accent on the single-asset viewer. Those are in #264, which is mostly your branding feature, so a review there would be genuinely valuable.

Also filed #265 and #266 from the same session, both in the share-link area if you are interested.

rubenxyz:
Thanks for this — the review and follow-ups all look right to me. The commit-then-delete reorder, the read-only GET, and especially the centralized `resolve_org_name()` are all better than what I had, and finishing the email white-labelling properly rather than leaving it as a follow-up was the right call.

On the loginLogoUrl question: agreed, leaving it out of the sidebar/preview fallback is correct — that slot is specifically the login-screen logo, so cascading it elsewhere would blur the slots.

One heads-up: we run the instance under a sub-path, so the `MAIL_FROM_NAME` upgrade note applies to us directly — will blank it on the next deploy. No other objections.

PR #204, opened 2026-07-22
Fix/hls ladder upscale filtering

### LINKED ISSUE #201: transcode: HLS ladder upscales below-1080p sources, wasting storage and CPU
Found during the v1.7.0 prod-stack smoke test.

## Problem

The HLS ladder is applied unconditionally, so a source smaller than a rendition gets **upscaled** into it. A 640×360 source produced 1920×1080 and 1280×720 renditions.

`packages/transcoder/ffmpeg_transcoder.py:145` filters only on which qualities were requested, never on the source resolution:

```python
qualities = [q for q in job.qualities if q in QUALITY_MAP]
```

and `:156` scales into each unconditionally:

```python
f"[v{i}]scale={QUALITY_MAP[q][0]}:force_original_aspect_ratio=decrease,pad=…"
```

`force_original_aspect_ratio=decrease` preserves aspect ratio but does **not** prevent upscaling — it only shrinks to fit the box, and a smaller source is stretched up to it.

## Measured

A 6-second 640×360 test clip (95 KB source), on the prod stack with MinIO:

| rendition | segments | total |
|---|---|---|
| `0/` 1920×1080 | 100 + 114 + 100 KB | **314 KB** |
| `1/` 1280×720 | 64 + 69 + 65 KB | **198 KB** |
| `2/` 640×360 | 40 + 42 + 40 KB | 122 KB |

512 KB of upscaled output — **~5× the source** — for zero quality gain, plus the CPU to encode two renditions nobody should ever select.

The master playlist also advertises `RESOLUTION=1920x1080` for content that is genuinely 360p, so an ABR player will happily pick the largest, most expensive variant.

This scales badly for self-hosters: every 720p deliverable pays for a 1080p encode it cannot benefit from.

## Suggested fix

Filter the ladder by source height before building `filter_complex` — the `ffprobe` step just above already has the dimensions. Keep at least the smallest rendition so there is always an output.

The codebase already has the right idiom in `packages/transcoder/image_processor.py:34`:

```python
"scale='min(400,iw)':'min(400,ih)':force_original_aspect_ratio=decrease"
```

Clamping with `min(target,iw)` in the video path would fix the stretching too, but dropping the redundant renditions outright is better — it saves the encode, not just the bytes.

## Notes

Pre-existing behavior, not a v1.7.0 regression — everything else in the pipeline verified clean end-to-end during the smoke test (upload → transcode → HLS → playback, zero unexpected errors).

siddharthgoyal00:
## Summary

Fixes #201

Fixes the HLS ladder being applied unconditionally regardless of source
resolution — a 640×360 source was producing upscaled 1080p/720p renditions
with `force_original_aspect_ratio=decrease` (which prevents distortion but
not upscaling). Found during v1.7.0 prod-stack smoke test; pre-existing
behavior, not a v1.7.0 regression.

## Changes
- `packages/transcoder/ffmpeg_transcoder.py`: filter `qualities` against
  `source_height` (from the existing ffprobe step) before building
  `filter_complex`, so renditions larger than the source are dropped
- Falls back to the smallest requested rendition if every requested quality
  exceeds the source resolution, so the ladder is never empty
- Added tests covering: source smaller than full ladder, source smaller than
  all requested qualities (fallback), source larger than ladder (no-op /
  regression check)

## Testing

## Checklist

## Screenshots
all testes passes sucessfuly.

Screenshot 2026-07-22 210218

### DIFF
--- a/apps/api/tests/test_ffmpeg_transcoder.py
+++ b/apps/api/tests/test_ffmpeg_transcoder.py
@@ -25,7 +25,39 @@ def _make_job(qualities: list[str] | None = None) -> TranscodeJob:
         output_s3_prefix="hls/media-1/v1",
         qualities=qualities or ["1080p", "720p", "360p"],
     )
+def _mock_probe_side_effect(width: int, height: int):
+    """Build a subprocess.run side_effect returning a fixed video probe
+    (given width/height) and no audio stream, then a generic success
+    result for the main ffmpeg transcode call."""
+    def mock_run_side_effect(cmd, **_kwargs):
+        if "-select_streams" in cmd and cmd[cmd.index("-select_streams") + 1] == "v:0":
+            mock = MagicMock()
+            mock.returncode = 0
+            mock.stderr = ""
+            mock.stdout = json.dumps({
+                "streams": [{"r_frame_rate": "30/1", "duration": 6.0, "width": width, "height": height}],
+            })
+            return mock
+        if "-select_streams" in cmd and cmd[cmd.index("-select_streams") + 1] == "a":
+            mock = MagicMock()
+            mock.returncode = 0
+            mock.stderr = ""
+            mock.stdout = json.dumps({"streams": []})
+            return mock
+        mock = MagicMock()
+        mock.returncode = 0
+        mock.stderr = ""
+        return mock
+    return mock_run_side_effect
 
+
+def _get_ffmpeg_cmd(mock_run):
+    """Pull the main ffmpeg transcode command (the one with -filter_complex)
+    out of a mocked subprocess.run's call list."""
+    ffmpeg_calls = [c for c in mock_run.call_args_list
+                    if any("filter_complex" in str(a) for a in c[0][0])]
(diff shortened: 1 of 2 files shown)


ravirajsinh45 (approved):
Thanks for this, @siddharthgoyal00. Clean fix for a real bug: force_original_aspect_ratio=decrease preserves aspect ratio but doesn't stop upscaling, so a 640x360 source really was getting padded up to fake 1080p/720p renditions.

The fallback for "source smaller than every requested quality" (keep the smallest requested rendition rather than producing zero) is the right call, and the tests actually exercise it: mocking subprocess.run at the ffprobe/ffmpeg boundary and asserting on the real -filter_complex output, not just checking that transcode() returns success.

Verified independently: merges cleanly onto current main, full backend suite (225 tests, including your 3 new ones) passes. CI was already green on your branch.

One minor nit for a follow-up, not blocking this: the old unfiltered qualities line is left in as a comment rather than deleted. Worth cleaning up next time you're in this file.

PR #236, opened 2026-08-05
fix(web): pin image annotations to the picture, not the letterboxed box

### LINKED ISSUE #185: Image annotations positioned relative to the letterboxed container, not the rendered image — misalign across containers of different aspect ratio
## Summary
Image annotations (drawn with the Fabric.js canvas in image review) are authored and displayed relative to the full, letterboxed container box (`object-contain` centers/pads the `` inside its container), not the actual rendered image box. Because different containers letterbox the same image differently (sidebar collapsed vs. expanded, browser resize, compare/side-by-side panes, share view), the same annotation shows up in a different position relative to the image content depending on where it's viewed.

Video annotations do not have this problem: `VideoFrameConstraint` in `apps/web/components/review/video-player.tsx` fits the annotation overlay to the actual rendered video box (excluding letterbox bars), for both authoring and playback. There is no equivalent for images.

## Steps to reproduce
1. Open an image asset whose aspect ratio differs noticeably from its viewer container (e.g. a portrait image in a wide/landscape panel, so it's letterboxed left/right).
2. Enter drawing mode and draw a shape near one edge of the *visible image content* (not in the empty letterbox band).
3. Save the comment with the annotation.
4. View the same comment's annotation in a container with a different aspect ratio than the one you drew in — e.g. collapse/expand the comment sidebar (changes the image container's width), resize the browser window, or open the same asset/comment in the side-by-side compare view.
5. Observe: the annotation shifts relative to the image content — it no longer sits over the same feature of the image — because its coordinates were captured/rescaled against the full container's box rather than the image's actual rendered box.

## Root cause
- `apps/web/components/review/image-viewer.tsx` (`SingleImage`), lines 104-112 — the annotation overlay is mounted in a plain full-container div, not the image's `object-contain` box:
  ```tsx
  {annotationOverlay && (
    <div
      className="absolute inset-0"
      style={{ pointerEvents: isDrawingMode ? 'auto' : 'none' }}
    >
      {annotationOverlay}
    </div>
  )}
  ```

- `apps/web/hooks/use-drawing.ts`, lines 65-67 and 92-93 — the Fabric authoring canvas is sized from the parent container's box, not the image's rendered box, so users can draw across the whole letterboxed area:
  ```ts
  const parent = el.parentElement
  const parentW = parent?.clientWidth ?? 0
  const parentH = parent?.clientHeight ?? 0
  ...
  const w = sharedPendingSize?.w ?? parentW
  const h = sharedPendingSize?.h ?? parentH
  ```
  Coordinates are captured in that container space and persisted as `_canvasWidth`/`_canvasHeight` (lines 325-326).

- `apps/web/components/review/annotation-overlay.tsx` (read-only render), lines 38-39 and 63-66 — sizes its Fabric canvas from the container's `offsetWidth`/`offsetHeight`, then rescales stored objects by a simple linear ratio:
  ```ts
  let w = containerRef.current.offsetWidth
  let h = containerRef.current.offsetHeight
  ...
  const scaleX = w / origWidth
  const scaleY = h / origHeight
  ```
  This only corrects for uniform container resizing; it has no knowledge of the image's letterbox offset, so when the container's aspect ratio differs between the authoring view and the display view, the linear rescale does not compensate for the differing letterbox insets and the annotation lands in the wrong place relative to the image content.

- By contrast, `apps/web/components/review/video-player.tsx`, lines 48-124 (`VideoFrameConstraint`) computes the actual rendered video box (excluding letterbox bars) from `video.videoWidth`/`video.videoHeight` vs. the container's aspect ratio, and positions/sizes the annotation layer to exactly that box for both authoring and playback (used at video-player.tsx:355-357 and in the compare overlay). No equivalent `ImageFrameConstraint` exists for images anywhere in the codebase.

## Suggested fix
Add an `ImageFrameConstraint` component mirroring `VideoFrameConstraint`:
1. Read the ``'s `naturalWidth`/`naturalHeight` and the parent container's `clientWidth`/`clientHeight`.
2. Compute the actual rendered image box (accounting for the `object-contain` letterbox), the same way `VideoFrameConstraint` does for `videoWidth`/`videoHeight`.
3. Wrap both the authoring canvas (`use-drawing`'s `<canvas>`, mounted from `image-viewer.tsx`) and the read-only `AnnotationOverlay` in this constraint so annotation coordinates are always image-frame-relative instead of container-relative.
4. Recalculate on image load and on container resize (`ResizeObserver`), matching `VideoFrameConstraint`'s pattern.

This mirrors the existing video pattern and makes image annotation coordinates stable across viewers of different aspect ratio (single viewer, compare, share links, sidebar collapsed/expanded, window resize).

ravirajsinh45:
Closes #185.

Image annotations were authored and replayed against the full letterboxed **container** box rather than the picture inside it, so the same drawing pointed at a different part of the image depending on where it was viewed: sidebar collapsed vs expanded, a resized window, a compare pane, a share link.

### The mechanism

`use-drawing` sized the Fabric canvas from the parent container (`use-drawing.ts:65-67`), stored that size as `_canvasWidth`/`_canvasHeight` (`:325-326`), and `AnnotationOverlay` replayed it with a plain per-axis rescale (`annotation-overlay.tsx:65-79`). That rescale is only correct when the display container has the **same aspect ratio** as the authoring one. When it does not, the drawing is stretched non-uniformly *and* the picture underneath has moved, because the letterbox insets differ.

Measured in a browser, a 3:1 image annotated 25% down the picture in a 900x500 viewer and reopened at 400x700:

| | authored at | lands at | drift |
|---|---|---|---|
| before | 25% down the picture | **-29%** | **-54%** of image height, off the picture entirely |
| after | 25% down the picture | 25% | 0 |

### Why this is not a copy of `VideoFrameConstraint`

Video never had this bug, so the obvious fix is to mirror `VideoFrameConstraint`. That would have been subtly wrong, because the two elements use different CSS:

```
video-player.tsx  <video>  absolute inset-0 w-full h-full object-contain
image-viewer.tsx  <img>    max-w-full max-h-full object-contain
```

`<video>` fills its container and letterboxes internally, so the contain fit can be derived from the container. `max-*` only ever **shrinks** a replaced element, so an image smaller than its container renders at natural size and is not scaled up. Verified in Chrome against the exact shipped CSS:

| case | real element box | container-derived math |
|---|---|---|
| 100x200 img in 800x400, `max-*` | **100x200 @ (350,100)** | 200x400 @ (300,0) ✗ |
| 4000x1000 img in 800x400, `max-*` | 800x200 @ (0,100) | 800x200 @ (0,100) ✓ |
| 100x200 img, `w-full h-full` | 800x400 @ (0,0) | 200x400 @ (300,0) ✓ |

`renderedImageBox` therefore runs the contain fit **inside the element's own box** and offsets by where that element sits, which is correct for all three. It reads `offset*` rather than a client rect because this renders inside react-zoom-pan-pinch's `TransformComponent`, where a client rect would fold the zoom matrix in.

Aspect ratios are compared by cross-multiplying rather than dividing, so an exact fit (a 16:9 image in a 16:9 box) stays exact instead of drifting a fraction of a pixel through an intermediate ratio.

### Surfaces covered

- Single image viewer, which also covers the folder share viewer since that routes through `ImageViewer`
- Both side-by-side compare panes
- The wipe view

Wipe is **not** optional scope even though the issue does not mention it. Once authoring moves into image space, a viewer still rendering in container space goes from consistent-but-wrong to newly wrong. Leaving it would have been a regression introduced by this PR.

`AnnotationOverlay` and `AnnotationCanvas` are unchanged. Both size themselves from their parent, so mounting them inside the constraint is the whole mechanism.

### Tests

263 to 282 (+19, 3 new files), full suite green.

Red-green was verified per surface, and again at the end in one pass: reverting only the three wiring files fails exactly the four regression tests, one per surface, while the helper and component unit tests stay green.

The `renderedImageBox` unit tests pin the three browser-measured shapes above, so a future change back to container-derived math fails immediately rather than silently.

### One consequence to be aware of

**Image annotations saved before this change will shift.** They were only ever correct in the exact container they were drawn in, and are now interpreted in image space. There is no version marker in the stored Fabric JSON to tell old data from new, so no migration is attempted here. If preserving them matters, the cheap path is to start stamping a frame marker in `getJSON()` now so a migration becomes possible later. Happy to add that as a follow-up if you want it.


ravirajsinh45:
Ran a deep multi-agent review against the actual worktree (with tests/tsc runnable, not just the diff). Found four confirmed issues worth addressing before merge:

**Correctness gap**: `ImageFrameConstraint` (image-frame-constraint.tsx:75) observes the `` itself with ResizeObserver, not its container. For an image smaller than its viewer (very common with `max-w-full max-h-full`, which only shrinks, never upscales), a container resize just recenters the image without changing its own offsetWidth/offsetHeight, so ResizeObserver never fires and the annotation stays pinned at the stale offset. This is exactly the class of drift the PR fixes, just not fixed for the shrink-not-needed case. `VideoFrameConstraint`, which this mirrors, avoids it by observing the container instead of the video element itself. The new resize test only exercises the box-grows case, not recenter-only.

**Reuse**: `renderedImageBox` (media-frame.ts) and `VideoFrameConstraint`'s inline calc (video-player.tsx) independently reimplement the same contain-fit math, one by cross-multiplication (with a documented precision fix), one by division (still has the precision issue, no direct test coverage). Worth factoring into one shared, tested primitive.

**Reuse**: the jsdom geometry-stub helper is copy-pasted verbatim into three new test files (wipe-viewer.test.tsx, compare-overlay.test.tsx, image-viewer-annotation-frame.test.tsx). A shared test-utils helper would avoid three-way drift.

**Test coverage**: side-by-side compare pane A has no geometry-pinning regression test, only pane B does. Confirmed by stripping ImageFrameConstraint from pane A only and running the suite, the full 282-test suite still passes.

Will push fixes for these.

PR #238, opened 2026-08-11
fix: clear stale compare sync offsets on version switch

### LINKED ISSUE #182: Compare overlay: switching a version does not reset per-side sync offsets (offA/offB), silently mis-syncing the new pair
**Component:** `apps/web/components/review/compare/compare-overlay.tsx` (video compare / wipe-and-scrubber overlay, introduced in PR #169)

**Repro**
1. Open the compare overlay on a video asset with 3+ ready versions (e.g. v1, v2, v3).
2. Compare v1 (left/A) vs v2 (right/B).
3. Use the offset steppers in the scrubber to nudge A and/or B (`offA`/`offB`) to align the two clips (e.g. because v2 has a few extra frames of pre-roll).
4. Without clicking the "Sync" / reset-offsets button, switch either side's version via its `CompareVersionSelect` dropdown — e.g. change the left pane from v1 to v3.
5. Observe: the new pairing (v3 vs v2) plays back with the stale offset(s) from the v1/v2 comparison still applied, so the two clips are silently mis-synced. Nothing in the UI indicates the offsets are stale/carried over.

**Root cause**

`offA`/`offB` are read straight from URL search params and fed into `timingA`/`timingB`:

```tsx
// compare-overlay.tsx:87-88
const offA = parseOffsetParam(searchParams.get('offA'))
const offB = parseOffsetParam(searchParams.get('offB'))
```

The two version-select dropdowns' `onChange` handlers only update which version is shown — they never touch the offset params:

```tsx
// compare-overlay.tsx:313 (left/A version switch)
onChange={(v) => writeParams((p) => p.set('compare', v.id))}

// compare-overlay.tsx:341 (right/B version switch)
onChange={(v) => setCurrentVersion(v)}
```

`CompareVersionSelect` itself (`compare-version-select.tsx:72-76`) is a plain controlled dropdown that just invokes the `onChange` prop it's given — it has no concept of offsets, so there's no hidden reset happening there either.

The **only** code path that clears `offA`/`offB` is the explicit "Sync" button in the scrubber:

```tsx
// compare-overlay.tsx:499-501
onResetOffsets={() =>
  writeParams((p) => { p.delete('offA'); p.delete('offB') })
}
```

wired to the manual reset button in `compare-scrubber.tsx:229-236` (`title="Reset offsets — re-sync both sides"`).

So any offset dialed in for one version pairing silently carries over to whatever new pairing the user selects next, with no reset and no warning.

**Suggested fix**

Clear `offA`/`offB` whenever the *identity* of either compared version changes, not just when the user explicitly clicks Sync. E.g. in the two `onChange` handlers at compare-overlay.tsx:313 and :341, also delete `offA`/`offB` from the params before setting the new version:

```tsx
onChange={(v) => writeParams((p) => { p.set('compare', v.id); p.delete('offA'); p.delete('offB') })}
```

and for the right/B side, similarly clear the offset params (via `writeParams`, or by having `setCurrentVersion`'s caller also clear them) before/along with `setCurrentVersion(v)`. Alternatively, key `offA`/`offB` off the pair `(left.id, right.id)` so a URL replace / effect resets them automatically whenever either id changes.

vandit-bera:
## Summary
- `offA`/`offB` (per-side sync offsets) were calibrated for whichever version pair was being compared. Switching either pane's version via the compare dropdown left the old offsets in place, silently misapplying a stale sync correction to the new pair — and, per the issue, re-thrashing the decoder trying to hold a wrong alignment on mixed-fps comparisons.
- Both `onChange` handlers on `CompareVersionSelect` now clear `offA`/`offB` whenever either side's version changes.

## Test plan
- [x] Added two tests in `compare-overlay.test.tsx` asserting `offA`/`offB` are stripped from the URL after switching either pane's version.
- [x] `npx vitest run components/review/compare/__tests__/compare-overlay.test.tsx` — 26/26 pass
- [x] `npx tsc --noEmit` — clean

Fixes #182


ravirajsinh45 (changes requested):
Hey @vandit-bera, ran a deep review against the actual branch (with tests runnable) and found a couple of correctness gaps worth fixing before merge.

**Re-selecting the current version wipes calibrated offsets.** CompareVersionSelect has no guard against clicking the already-selected version, so reopening a dropdown and re-clicking the same item still fires onChange and deletes offA/offB even though the compared pair never changed.

**Right-side switch has a state-race window.** offA/offB clear via `router.replace` (deferred through React's startTransition), while `setCurrentVersion` is a synchronous Zustand write right after. That mismatch means a render can briefly show the new version with the stale offset still applied, before the URL update lands. The left-side handler doesn't have this because it bundles both changes into one writeParams call.

**Test coverage**: the two new tests only assert the URL string from the mocked router.replace, never the resulting state. Confirmed by removing `setCurrentVersion(v)` from handleSwitchRight entirely, both tests still passed.

Happy to pair on fixes if useful.

PR #239, opened 2026-08-11
fix: clear stale pane annotation on compare version switch

### LINKED ISSUE #181: Compare view: stale comment-drawing annotation persists on a pane after switching that pane's version
### Summary
In the version-compare overlay (`apps/web/components/review/compare/compare-overlay.tsx`, added in #169), clicking a timeline comment marker that has a drawing shows that drawing overlaid on the corresponding pane (`annotationA` / `annotationB`, set via `showAnnotationFor`). If the user then switches that same pane's version using the version dropdown (`CompareVersionSelect` for the left or right pane) while playback is paused, the old drawing keeps rendering — now overlaid on the *new* version's footage, which it was never authored against.

### Steps to reproduce
1. Open a video asset that has 2+ ready versions and open Compare.
2. On either pane (left `compare` param or right `rightVersion`), click a comment marker on the scrubber that has an attached drawing/annotation. The drawing renders over that pane's video frame (paused).
3. Without pressing play, use that same pane's version selector dropdown to switch to a different version.
4. Observe: the pane now shows the newly selected version's footage, but the old drawing from step 2 is still rendered on top of it.

### Root cause
`compare-overlay.tsx` keeps the shown drawing in component-local state:

```tsx
const [annotationA, setAnnotationA] = React.useState<Record<string, unknown> | null>(null)
const [annotationB, setAnnotationB] = React.useState<Record<string, unknown> | null>(null)
const [lastAnnotationSide, setLastAnnotationSide] = React.useState<'a' | 'b' | null>(null)
```

The only effect that clears this state is keyed on `transport.isPlaying`:

```tsx
// Normal-player parity: starting playback clears the shown drawing ...
React.useEffect(() => {
  if (!transport.isPlaying) return
  setAnnotationA(null)
  setAnnotationB(null)
  setLastAnnotationSide(null)
  setDrawingSide(null)
}, [transport.isPlaying])
```

There is no effect watching the selected version for either pane (`left?.id` from the `compare` query param, or `right.id` / `rightVersion.id`). Since the component doesn't remount when a version is switched via the dropdown (`onChange={(v) => writeParams((p) => p.set('compare', v.id))}` for the left pane, `onChange={(v) => setCurrentVersion(v)}` for the right pane), `annotationA`/`annotationB` simply carry over across the version switch, even though `urlA`/`urlB` (and thus the rendered footage) have changed.

### Suggested fix
Add a per-pane effect that clears that pane's shown/in-progress drawing when its version id changes, mirroring the existing isPlaying-based clear:

```tsx
React.useEffect(() => {
  setAnnotationA(null)
  setLastAnnotationSide((prev) => (prev === 'a' ? null : prev))
  setDrawingSide((prev) => (prev === 'a' ? null : prev))
}, [left?.id])

React.useEffect(() => {
  setAnnotationB(null)
  setLastAnnotationSide((prev) => (prev === 'b' ? null : prev))
  setDrawingSide((prev) => (prev === 'b' ? null : prev))
}, [right.id])
```

This clears only the affected pane's drawing state (and cancels any in-progress authoring on that pane via `drawingSide`), leaving the other pane's drawing untouched.

### Affected file
`apps/web/components/review/compare/compare-overlay.tsx` (introduced in #169)

vandit-bera:
## Summary
- A pane's displayed comment drawing/annotation is frame-anchored to a specific comment on the version being viewed. Switching that pane's version via the compare dropdown left the old drawing rendered over the newly-selected version's footage.
- Both `onChange` handlers on `CompareVersionSelect` now clear that pane's shown annotation and focused-comment state on switch.

**Stacked on #238** (fix/182) — this branch is built on top of that PR's commit (same two `onChange` handlers), so the diff here currently includes #238's changes too. Merge #238 first; once it lands in `main`, this diff will narrow to just the #181 fix.

## Test plan
- [x] Added two tests in `compare-overlay.test.tsx` asserting the shown annotation clears after switching either pane's version.
- [x] `npx vitest run components/review/compare/__tests__/compare-overlay.test.tsx` — 28/28 pass
- [x] `npx tsc --noEmit` — clean

Fixes #181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Optional hardware-accelerated transcoding (NVENC / VAAPI) with HDR and Dolby Vision support.** Auto-detects an available GPU backend and falls back to the software pipeline when none is present, or when a hardware attempt fails at runtime (for example, out of VRAM). CPU-only and arm64 builds are unaffected by default. New env vars: `TRANSCODER_PIPELINE`, `TRANSCODER_OUTPUT`, `TRANSCODER_HDR`. (#127)
 
 ### Fixed
+- **Switching a version in compare no longer leaves the previous version's drawing on screen** — a comment's annotation stayed painted over the pane after that pane moved to a different version, anchored to a frame that was no longer there. It is now cleared when the pane showing it switches, and left alone when the other pane switches. (#181)
 - **Switching a version in compare no longer applies the previous pair's sync offsets** — offsets calibrated for one pair of versions stayed in the URL after switching either pane, silently misaligning the new pair. They are now cleared on a real switch, and left alone when you re-pick the version a pane already shows. (#182)
 
 - **Branding UI remains readable and correct across themes** — custom muted accents are now opaque, selected chips and controls use readable theme text, default logos pair light/dark artwork in previews and password gates, and the Apple home-screen icon uses its dedicated asset. (#275)
--- a/apps/web/components/review/compare/__tests__/compare-overlay.test.tsx
+++ b/apps/web/components/review/compare/__tests__/compare-overlay.test.tsx
@@ -478,6 +478,32 @@ describe('CompareOverlay per-pane annotation display', () => {
     expect(screen.queryByTestId('annotation-overlay')).not.toBeInTheDocument()
   })
 
+  it('switching one pane leaves a comment focused on the OTHER pane alone', () => {
+    // focusedCommentId is global, not per-pane, and the comment list and
+    // progress bar highlight from it. Clearing it on any switch silently
+    // unfocused a comment belonging to the pane that did not change.
+    commentsByVersion['v-3'] = [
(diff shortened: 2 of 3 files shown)


ravirajsinh45 (changes requested):
Hey @vandit-bera, found one real correctness regression plus a merge-order heads-up on top of the overlap with #238 here.

**Switching one pane's version clears the OTHER pane's focused comment.** focusedCommentId is global store state, not per-pane, but both switch handlers null it unconditionally. Reproduced directly: focus a comment on pane B, switch pane A's version, focus silently drops to null. This is new in this PR, the left handler never touched that field before.

**Test coverage**: no test guards the cross-pane case, so the regression above shipped with a fully green suite. A test that focuses a comment on one pane, switches the other pane's version, and asserts the first pane's focus survives would have caught it.

**Merge-order heads-up**: the PR body says the diff will narrow to just the #181 fix once #238 merges first. That only holds for a fast-forward/rebase-merge that keeps #238's commit SHA. This repo's actual convention is squash-merge, and squash-merging #238 then merging #239 as-is produces a real conflict in compare-overlay.tsx on the shared handler code (confirmed by reproducing it in both merge orderings). Worth rebasing #239 onto main right after #238 lands rather than assuming a clean auto-narrow.

PR #245, opened 2026-08-14
fix(web): hold a wake lock so a sleeping machine can't kill an upload

### LINKED ISSUE #244: Hold a Screen Wake Lock while an upload is running
### Problem

An upload lives entirely in the browser tab. If the machine suspends
mid-transfer, the loop pushing chunks dies with it.

What makes this worse than a normal failure is that **nothing runs afterwards**.
On a failed part, the `catch` in `startUpload` calls `POST /upload/abort`, which
marks the version `failed` — that is what makes it show up correctly in the
upload history. A suspend skips all of it:

- the version stays at `processing_status = 'uploading'`, which in the UI is
  indistinguishable from a stalled transcode — the user sees a spinner and
  assumes the server is busy, when in fact nothing is happening at all;
- the S3 multipart upload stays open, its parts fully billed but invisible in
  any ordinary object listing, until `reap_stale_uploads` gets to it
  (`stale_upload_timeout_hours`, default 24).

This is not hypothetical. Large files take a long time — a 3.75 GB upload on a
7.6 Mbit/s uplink is ~65 minutes — and users close laptops. We lost a 36 %
upload this way before adding the workaround described below.

### Proposed solution

Ask the browser for a **Screen Wake Lock** while an upload is in progress:

```ts
wakeLock = await navigator.wakeLock.request('screen')
```

Three details matter:

1. **Reference-count it.** Several uploads can run at once; the lock should only
   be released when the last one finishes, in the `finally` that already cleans
   up the abort controller.
2. **Re-acquire on `visibilitychange`.** Browsers drop the lock whenever the tab
   is hidden, and the dropped sentinel cannot be reused — without this, switching
   tabs once silently ends the protection.
3. **Best-effort, never blocking.** Without a secure context, in low-power mode,
   or in a browser without the API there is simply no lock. Every failure path
   should be swallowed: not having a wake lock is not an upload error.

This does not replace resumable uploads (#241) — it reduces how often the tab
dies, it cannot recover once it has. The two are complementary.

I have this implemented with tests and will open a PR.

### Alternatives considered

**A `beforeunload` warning.** Catches the deliberate tab close, but not a
suspend, which is the more common case. Also browsers heavily restrict what such
a prompt may say. Worth adding separately, but it solves a different half.

**Detecting the wake-up and resuming.** That is #241, and it is strictly better —
but also a much larger change. The wake lock is ten lines and helps immediately.

**Doing nothing and relying on the reaper.** `reap_stale_uploads` does clean up
after 24 hours, so nothing leaks permanently. It does not help the user who lost
an hour of uploading, and it does not fix the misleading `uploading` state in the
meantime.

Lennart-Pingpong:
### Summary

An upload lives entirely in the browser tab, so a machine suspending mid-transfer
kills the loop pushing chunks. Unlike a failed part, nothing runs afterwards —
the `catch` calling `/upload/abort` never executes — so the version is left at
`processing_status = 'uploading'`, indistinguishable in the UI from a stalled
transcode, and the multipart upload stays open until `reap_stale_uploads` gets to
it.

The browser is now asked for a Screen Wake Lock while an upload runs.

Closes #244

### Changes

- `retainWakeLock()` / `releaseWakeLock()`, reference-counted so the lock is only
  dropped when the last concurrent upload finishes
- Released in the `finally` that already cleans up the abort controller, so it
  also covers the cancel and error paths
- Re-acquired on `visibilitychange`, because browsers drop the lock when the tab
  is hidden and the dropped sentinel cannot be reused
- Every failure path swallowed: no secure context, low-power mode or a browser
  without the API simply means no lock, never a failed upload
- 6 tests covering acquire, release, the concurrent-upload refcount, the
  request-only-once case, the missing-API case and a refused lock

### Testing

- [ ] Backend tests pass — **not run: no backend files touched by this PR**
- [x] Frontend tests + build pass (`pnpm test` → 269 tests passing,
      `pnpm exec tsc --noEmit` clean, `pnpm build` succeeds)
- [ ] Tested manually in browser — the automated tests drive the API through a
      stub; the real-world effect (a laptop lid closing) is by nature awkward to
      assert in CI. Happy to run a manual pass if you'd like one before merging.

### Checklist

### Screenshots

n/a — no UI changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed
-- **A machine going to sleep no longer silently kills an upload** — the browser is now asked for a Screen Wake Lock while an upload is in progress, and releases it when the last one finishes. An upload lives entirely in the tab, so a suspend mid-transfer kills the loop pushing chunks; worse, nothing runs afterwards, so the `/upload/abort` call never happens and the version is left at `processing_status = 'uploading'`, which in the UI is indistinguishable from a stalled transcode. Best-effort by design: without HTTPS, in low-power mode, or in a browser without the API there is simply no lock, which never blocks the upload. The lock is re-acquired on `visibilitychange`, since browsers drop it whenever the tab is hidden.
-
 ### Changed
 - **Pinned `starlette` and `botocore` so self-hosted Docker builds are reproducible** — both were floating transitives, so rebuilding the same commit on a different day could silently install different versions with no diff and no PR. FastAPI declares `starlette>=0.46.0` with no upper bound, meaning builds were free to cross a Starlette major (the ASGI layer under the SSE endpoint and the middleware stack); `botocore` is where S3 request signing lives, and unreviewed moves there have broken S3 compatibility before. Both are pinned to the versions already resolving, so no installed version changes.
 
 ### Fixed
 - **A single transient storage error no longer discards an entire multipart upload** — each part is now retried up to 8 times with exponential backoff and jitter (≈254s of tolerance per part) instead of failing the whole upload on the first non-OK response. A multi-gigabyte file is several hundred sequential requests, so one 500/503 from the object store or a brief network drop was enough to throw away everything already transferred; on a slow uplink that can be an hour of work. The presigned URL is re-fetched on every attempt, since `presign_upload_part` expires after an hour and a large upload can outlive that. User-initiated cancels are never retried, and a 4xx is thrown straight through instead of being repeated eight times over four minutes — it would fail identically every time, so retrying it only delays the error. The part loop, previously duplicated between new-asset and new-version uploads, is now shared.
+- **A machine going to sleep no longer silently kills an upload** — the browser is now asked for a Screen Wake Lock while an upload is in progress, and releases it when the last one finishes. An upload lives entirely in the tab, so a suspend mid-transfer kills the loop pushing chunks; worse, nothing runs afterwards, so the `/upload/abort` call never happens and the version is left at `processing_status = 'uploading'`, which in the UI is indistinguishable from a stalled transcode. Best-effort by design: without HTTPS, in low-power mode, or in a browser without the API there is simply no lock, which never blocks the upload. The lock is re-acquired on `visibilitychange`, since browsers drop it whenever the tab is hidden.
 
 ## [1.7.5] - 2026-07-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


ravirajsinh45 (approved):
Thanks @Lennart-Pingpong. The approach is right and the three details you called out in #244 are the correct three. We did find a race in the reference counting, so rather than send it back we pushed a fix to your branch.

**Concurrent uploads requested a sentinel each.** `acquireWakeLock` guarded on `if (wakeLock) return`, but that is only set after the `await` resolves. `handleStartUpload` runs `pendingFiles.forEach(f => startUpload(f))`, and each `startUpload` runs synchronously as far as `retainWakeLock()` before its first `await`. So dropping three files produced three concurrent `navigator.wakeLock.request('screen')` calls, only the last of which was stored. The other two were never released and held the screen awake for the life of the page, which is the opposite of what the change is for.

Worth flagging why the suite did not catch it: `requests the lock only once for concurrent uploads` awaits between the two `retainWakeLock()` calls, so by the second call the first has already resolved and the guard works. That await is exactly what hides the race.

**A lock arriving after the last upload finished was held forever.** If an upload completes while `request()` is still in flight, `releaseWakeLock()` runs while `wakeLock` is still `null`, so it releases nothing; the sentinel then arrives and is stored with the holder count already at zero. Short files on a fast connection hit this.

The fix guards on a pending promise so only one request is ever in flight, and re-checks the holder count after awaiting so a sentinel that arrives too late is released immediately. We also moved `retainWakeLock()` above the `try`, since it sat inside it while the release sits in the `finally`, meaning a throw before the retain would have released a concurrent upload's lock. And the same CHANGELOG ordering nit as #243.

Two regression tests came with it, both failing before the change and passing after. Full suite is 271 green.

Please give it a sanity read when you get a chance. Merging after that and once CI re-runs.

Lennart-Pingpong:
Hi there - Sanity read done — the fix is right, and I verified it rather than just reading
it. I checked out the branch at your commit's parent and ran your two new tests
against the original implementation: both fail. Then the full suite on your
version: 279 green, `tsc --noEmit` clean, build succeeds.

The diagnosis of my own test is the part I want to acknowledge. `requests the
lock only once for concurrent uploads` awaits between the two `retainWakeLock()`
calls, so the first request has already resolved by the time the second runs and
the `if (wakeLock) return` guard does its job. It asserts the property I meant to
protect while arranging the one situation in which the bug cannot appear — and
the real caller (`pendingFiles.forEach(f => startUpload(f))`) does the opposite,
retaining N times inside a single tick. That test was worse than no test, because
it made the guard look verified.

Three things in your version I'd have missed:

- `if (wakeLock === sentinel)` in the release listener. Mine nulled `wakeLock`
  unconditionally, so a stale sentinel firing late would clear a newer, live one.
- Reading `wakeLock` into `held` before nulling it in `releaseWakeLock`, rather
  than releasing through the field.
- Moving `retainWakeLock()` above the `try`. I had it inside while the release
  sits in `finally`, so a throw between the two would have decremented a counter
  it never incremented — releasing a concurrent upload's lock. Same class of bug
  as the one you found, and I walked straight past it.

Happy for this to merge as is.

On the docs caveat: agreed, and it is worth stating plainly, since Wake Lock
needs a secure context and the HTTP-only setups in `docs/deployment.md` simply
will not have it. Want me to send a short PR adding a line to the deployment
docs? Something like: *"Uploads request a Screen Wake Lock to keep the machine
from sleeping mid-transfer. This requires a secure context — on an HTTP-only
deployment the protection silently does not apply, and a suspending machine will
still lose an in-flight upload."* Say the word and I'll open it against
`docs/deployment.md`.

PR #250, opened 2026-08-15
feat(web): upload parts concurrently instead of one at a time

Lennart-Pingpong:
A single PUT does not saturate a connection - most of a sequential multipart upload is spent waiting on round-trips rather than moving bytes. Sending a handful of parts at once multiplies throughput on exactly the long-haul links where large uploads hurt most.

Part order is preserved without sorting: each result is written to its own index, so parts may finish out of order while the list stays ordered by part number. Progress counts completions rather than the highest part seen, which would otherwise jump around.

When a part finally fails, the other workers stop picking up new parts but are left to finish what they already started - aborting them mid-flight would trade one error for several, and the upload is over either way.

Concurrency is a parameter with a default, so tests can drive it directly.

## Changes

- 

## Testing

## Checklist


ravirajsinh45:
@Lennart-Pingpong Thanks for this. We reviewed it hard, because it is concurrency code on the path that assembles someone's only copy of a master. The mechanism is right and the shape is what we agreed on, so rather than send it back we pushed the follow-ups straight to your branch as `1b5520e9`, the way we did on #243 and #245. Your commit is untouched underneath. Everything below is what changed and why, so nothing lands without you seeing it.

Before that, the question you left on #242: no, we did not want you to hold this for benchmarks. Your 7.6 Mbit/s result cutting against the change is the expected result rather than a problem with it. At that bandwidth-delay product a single stream already fills the pipe, so there is nothing for concurrency to recover. Gating on a benchmark of a case we agree is not the target case would not have told us anything.

We also rebased it: it was conflicting on `CHANGELOG.md` against the v1.8.0 release that went out this morning, and the branch had picked up the `starlette`/`botocore` entry that already shipped in 1.8.0, which is a normal rebase artifact.

## What we changed

### 1. A doomed upload kept running for about four more minutes before it reported

`if (firstError) return` stops a worker claiming a *new* index, but it is never consulted inside `uploadPart`, so a worker already in the 8-attempt ladder runs the whole thing out, and `Promise.allSettled` waits for it.

We measured this on a 5-part file where part 1 gets a permanent 403 and parts 2 to 5 get transient 503s:

| | this branch, concurrency 5 | concurrency 1, which is main's shape |
|---|---|---|
| time to reject | 255,170 ms | 0 ms |
| PUTs issued | 33 | 1 |

Through the real store, `POST /upload/abort` fires at t=255s instead of t=0. That leaves the version at `processing_status = 'uploading'` for the whole window, which is exactly the limbo your own wake-lock PR existed to close. A milder and more common variant: part 1 fails permanently while the others are slow but succeeding, and the progress bar keeps climbing 19, 38, 57, 76 after the upload is already dead, then snaps to Failed.

We want to push back on the stated rationale, because we do not think it holds. The comment says aborting the others mid-flight "would trade one error for several." It would not: the latch already discards every subsequent error, so stopping them costs nothing and surfaces the same single error.

The fix gives the workers their own `AbortController`, so a part failing for good can stop its siblings without being mistaken for a user cancel. One wrinkle we hit: aborting alone is most of it but not all, because a worker asleep in the backoff is not watching the signal and will still sit out the rest of its current delay, up to 128 seconds on the last rung. The sleep now races the abort event.

A smaller thing fell out of the same area. Because `firstError` was thrown before the abort signal was consulted, cancelling during that window showed the row flipping from Cancelled back to Failed. A cancel now outranks the errors it caused, and we kept the AbortError a part actually raised rather than synthesizing one, so the reason survives and your existing cancel test still means what it says.

### 2. Nothing in the file pinned the property the change is named for

This is the one we would most like you to look at, because the implementation was correct and nothing would have caught it becoming incorrect.

We ran mutation testing. Five mutations survived with all 10 tests green, including replacing `parts[index] = ...` with `parts.push(...)`, which is the ordering guarantee itself, and reverting the progress formula to `partNumber`.

Two causes, and the second is the one worth your time:

- The new ordering test slept a uniform 10 ms for every part, so under fake timers the completion order was literally 1, 2, 3. The test titled "keeps several parts in flight and still returns them in order" never produced an out-of-order completion.
- The pre-existing two-part test mocked `fetch` with a positional `mockResolvedValueOnce` chain. That was a valid per-part mock only because the old loop was sequential, so call order and part order matched by construction. Under the pool it hands out ETags by scheduling order. Make part 1's presign resolve one macrotask later than part 2 and that existing assertion fails on this branch and passes on `main`.

Taken together, no test in the file bound an ETag to a part number under out-of-order completion, which is the headline claim of the commit message and the CHANGELOG, and whose failure mode is an `InvalidPart` from S3 at the very end of a multi-gigabyte upload.

Both mocks now key off `body.part_number`, the delays are deliberately reversed so later parts finish first, and completion order is asserted explicitly so the ordering assertion cannot pass vacuously. Two tests are added for the fast-fail behaviour, one of which measures timing, because reverting the interruptible sleep changes only *when* the promise settles and the suite could not see that. Four previously-surviving mutations now fail.

## Two smaller things we folded in

**The concurrency knob was not reachable outside tests.** Your issue asked for a parameter "so it is testable and can be tuned per deployment if a storage backend turns out to dislike it," and the first half shipped but not the second: `UPLOAD_CONCURRENCY` was module-private and neither call site passed it. It is now `NEXT_PUBLIC_UPLOAD_CONCURRENCY`, wired through `.env.example`, the deployment env table, the web build arg and the prod compose args, matching the existing `*_CONCURRENCY` knobs. This matters more than it looks: we spent this week mapping what the multipart primitives actually do across the backends people point FreeFrame at, and per-deployment concurrency is the knob most likely to be needed.

**The CHANGELOG entry overclaimed, and that one was ours** since we resolved the conflict. It said uploads are faster, unqualified. Your issue was careful to say the opposite for bandwidth-limited links, and the user in #241 is exactly that case. It is now scoped to links a single stream cannot fill, and says plainly that a bandwidth-limited uplink sees no change.

## What we checked and cleared

So you know where the bar was, rather than only what failed it. We traced `nextIndex++` for a claim race, and it is race-free with no `await` between read and write. We fuzzed 300 randomized runs across 1 to 12 parts and concurrency 1 to 6 with injected 4xx, 5xx and network faults, and got zero holes in the parts array, zero misorderings and zero non-settling promises. `parts.length` cannot come up short because of `new Array(totalChunks)`. Progress is monotonic and caps at 95. Extra parts uploaded after a failure are bounded at `workerCount - 1`, as your comment says. `/upload/abort` fires exactly once and never with a part still in flight. `concurrency` of 0 or a negative number clamps to 1, and a zero-chunk file behaves identically to `main`. Cancelling a healthy upload rejects with `AbortError` and lands as Cancelled.

We also chased and dismissed three theories we had going in: that `firstError ??= err` could latch a falsy value and return a sparse array (real mechanism, no reachable falsy rejection, and FastAPI would 422 it anyway), that five files in flight means 5N concurrent presigns saturating the pool (the arithmetic does not support it, presign is a fast in-process handler against a 15-connection pool), and that the retry jitter being 0.2% of the final rung causes a thundering herd (measurable, no demonstrated harm, and those are distinct parts that must be sent regardless).

One last thing at low confidence, which we could not demonstrate and are flagging rather than asserting, and which we deliberately did not change here. Presigned part URLs expire after an hour, and `isPermanentStatus` treats 403 as permanent so an expired one is never retried. The per-attempt re-presign covers a wait inside the backoff but not a wait inside the browser's own connection queue. Five in flight per file lowers the number of simultaneously dropped files at which that starts to bite by roughly 5x. It needs a real browser on a slow link to confirm, and the cheap hedge either way is a shorter expiry for part presigns, or not treating 403 as permanent once the signature's own expiry has passed. Worth its own issue rather than a rider here.

All four checks are green on `1b5520e9`, so this is going in. We also filed #251 to #254 for pre-existing things we hit while reading this path; none of them are yours. Thanks again, and please do push back on any of the above if you disagree; it is your change and we would rather argue it out than have you find these in the diff.

ravirajsinh45 (approved):
Approving. Mechanism and shape are the contributor's; the follow-ups in `1b5520e9` are ours and are described in the comment above. All four checks green, 319 web tests passing, four previously-surviving mutations now killed.

PR #255, opened 2026-08-15
fix(api): build the upload completion list from storage, not from the browser

### LINKED ISSUE #131: Docs/Bug: Multipart browser uploads fail with generic "Load failed"  if bucket CORS doesn't expose ETag header
### Describe the bug

**Describe the bug**
Large file uploads (multipart, client-side direct-to-S3 via presigned URLs)
fail partway through with a generic browser network error ("Load failed" in
Safari; similar generic errors in other browsers), even though the backend
API logs show no error at all - the request never even reaches the API for
the failed part.
 
**Root cause**
The client-side multipart upload flow (`apps/web/stores/upload-store.ts`)
reads the `ETag` response header from each `PUT` part upload to assemble the
`parts` array for `CompleteMultipartUpload`:
 
```js
parts.push({ PartNumber: partNumber, ETag: putResponse.headers.get('ETag') ?? '' })
```
 
Browsers only expose response headers to JS for cross-origin requests that
are explicitly listed in the bucket's CORS `ExposeHeaders`. If the bucket's
CORS configuration doesn't include `ETag` in `ExposeHeaders`, `ETag` reads
back as `null`/empty client-side even though the S3-compatible backend sent
it correctly - completion payloads end up with empty ETags and the upload
fails, with no useful diagnostic on either side.
 
This is especially relevant for Hetzner Object Storage users: CORS
configuration isn't exposed in the Hetzner Console UI at all (API/CLI only,
supported since Aug 2025), so it's an easy thing to miss entirely when
self-hosting against Hetzner - and unlike AWS S3, there's no docs pointer
from FreeFrame's own setup guide about this requirement.
 
**Suggested fix / documentation ask**
Add an explicit CORS requirement section to `docs/deployment.md` (or
wherever S3 configuration is documented) covering the minimum required
bucket CORS config for browser-based multipart uploads to work, e.g.:
 
```json
{
  "CORSRules": [
    {
      "AllowedOrigins": ["https://your-domain.example"],
      "AllowedMethods": ["GET", "PUT", "POST", "HEAD", "DELETE"],
      "AllowedHeaders": ["*"],
      "ExposeHeaders": ["ETag"],
      "MaxAgeSeconds": 3000
    }
  ]
}
```
 
Ideally also a startup/health-check warning if `ExposeHeaders` doesn't
include `ETag`, since this fails silently and is hard to diagnose without
digging into browser devtools network tab or client source.
 

### Steps to reproduce

upload a large video

### Expected behavior

upload successfull

### Screenshots

_No response_

### Area

Frontend (Next.js)

### Version / Commit

_No response_

ravirajsinh45:
## Summary

`/upload/complete` forwarded whatever parts list the browser sent straight to `CompleteMultipartUpload`. Completing with a part missing is **not an error** on most S3 implementations, so an under-reported list silently truncates the file. Verified against our own MinIO: upload parts 1 and 2, send only part 1, get a `200` and a half-length object. FFmpeg transcodes it, the version goes `ready`, and the raw master is the only copy the user has.

The server now calls `ListParts` and builds the list itself.

## Changes

- **`/upload/complete` derives and validates the parts list server-side.** Part numbers must run `1..N` with no holes, every part but the last must be the same size (also R2's hard requirement), and the sizes must total the size declared at initiate. A failure is a `409` naming the specific part, logged with the version, key and byte counts, since it can land after a very long transfer.
- **The client-reported list is now a fallback**, used only where `ListParts` is genuinely unavailable, and logged when it happens. `parts` becomes optional in the request schema.
- **Completing is idempotent.** Completing consumes the upload id, so a retry after a lost response used to fail, leave the version at `uploading`, and hand it to the reaper, which deletes `s3_key_raw` a day later. Retrying a *successful* upload destroyed it. A retry is recognised by comparing the stored object's size to the declared size.
- **Completing a version that is not `uploading` is refused.** Without this, replaying the request against a finished version rewound it to `processing` and queued a second transcode.
- **`/upload/abort` no longer leaves a version stuck, and no longer lies.** It called S3 before writing the DB, so an upload the reaper had already taken raised `NoSuchUpload`, 500'd, and pinned `processing_status = 'uploading'` permanently. Only "there was nothing to abort" is swallowed now; a permissions or throttling failure means parts are still in the bucket and is surfaced. A version that already finished is left alone, because the client fires abort from the catch of *every* completion failure.
- **The key is read from the version's `MediaFile`**, matched on both version and key, rather than taken from the request body.
- **Docs corrected.** `docs/deployment.md` and the startup CORS warning both told operators that `ExposeHeaders: ETag` was mandatory. It no longer is on any backend with `ListParts`.

## On issue #131

This does **not** close #131. That issue asked for a docs section and a startup warning, and both already shipped. What it documented was the requirement that the browser be able to read each part's `ETag`, and that requirement is what this removes: the server reads the parts from storage, so a bucket whose CORS omits `ExposeHeaders: ETag` no longer produces an upload that transfers every byte and then dies with an `InvalidPart` pointing nowhere near the cause. The docs #131 asked for are updated here rather than left describing the old behaviour.

## Portability

The default backend is MinIO, not AWS, and R2 / B2 / Spaces / Ceph are documented as supported. Three things this deliberately does:

- **Only ever echo back the backend's own `NextPartNumberMarker`, and require it to advance.** MinIO resolves a marker naming a part number it does not hold to an *empty page* rather than to the next part, and concurrent uploads (#242, just merged) create exactly those gaps. The advance check stops a backend that pages forever from spinning inside a request handler.
- **A part number listed twice is refused, not resolved.** Choosing between two writes of the same part needs the ETag to match bytes we cannot verify, and picking the stale one would assemble the wrong file silently.
- **`InvalidRequest` is not treated as "no `ListParts` here."** It is a generic 400; reading it as a capability statement would fail open into trusting the client. Probing MinIO with 15 malformed `ListParts` calls never produced it.

New deploy-time note: `ListParts` needs `s3:ListMultipartUploadParts` on AWS. No bucket policy is documented today, so nothing shipped is broken, but a least-privilege policy that omits it will now fail at completion.

## Testing

- `apps/api/tests/test_complete_upload_parts.py` — 25 tests on the validator and the S3 helpers
- `apps/api/tests/test_complete_upload_endpoint.py` — 15 tests driving the endpoint itself
- **267 passed, 2 skipped** for the full API suite
- Verified end-to-end against the real dev MinIO, not just mocks: a 3-part upload round-trips byte-exact; a short set and a hole are both rejected; a retry after success is correctly recognised as already-done

Reverting the whole server-side-derive block back to `parts = body.parts` now fails 8 tests. Before the endpoint tests were added it passed the entire suite, which is why they exist.

## Reviewer notes

Two things worth a second opinion:

1. **The total-bytes check compares against `file_size_bytes`, which the client declares at initiate and nothing verifies.** It catches every under-report we know how to produce, but it is not an independent integrity check. A real one needs per-part checksums, which are not portable across the backends we support.
2. **The equal-size rule is stricter than S3 requires** (S3 only mandates a 5 MB floor for non-final parts). It is what R2 needs and what catches a client whose chunk size changed mid-upload. It does mean that if `CHUNK_SIZE` in `upload-store.ts` is ever tuned, an upload in flight across the change cannot complete. That is an argument for pinning chunk size server-side at initiate, which #241 needs anyway.

## Relation to #241

`list_upload_parts` is step 1 of what #241 proposes, and the total-bytes check makes resume safer: re-selecting a different file fails closed instead of assembling a mixed master. Two follow-ups that belong with #241 rather than here: returning the missing part numbers as structured JSON instead of prose, and pinning chunk size per upload.


ravirajsinh45:
@Lennart-Pingpong Would you be willing to look over this one? No obligation, and it is not small, so please say no if it does not suit.

Asking you specifically for two reasons. You have spent more time in this upload path in the last two days than anyone, and this changes the contract #241 is being built against, so you would hit it either way. Better you hit it now, in review, than as a surprise when you rebase.

**What it means for #241.** After this, the client does not need to track ETags at all. The server reads the parts from storage when completing, so resume only has to answer "which part numbers still need sending", and the ETag CORS problem stops being something you have to reason about. `list_upload_parts` in `s3_service.py` is also, roughly, the `ListParts` wrapper your issue asked for, including the pagination loop you correctly flagged as necessary, so you may be able to build on it rather than write a second one.

**Two things in here that will constrain your design**, and where we would most value your opinion:

1. The validator requires every non-final part to be the same size, and derives that size from part 1. That is what R2 needs and it catches a client whose chunk size changed mid-upload. But it also means that if `CHUNK_SIZE` in `upload-store.ts` is ever tuned, an upload in flight across that change can never complete. For a resume spanning sessions and therefore releases, that is a real hazard, and it is the concrete argument for pinning chunk size server-side at initiate rather than reading a client constant. We think that pinning belongs in #241 rather than here, but if you think it has to land first, say so.

2. When the validator rejects, it computes the full list of missing part numbers and then throws all but the first away into an English sentence. A resume client wants that list. Returning it as structured JSON costs almost nothing and is arguably the API for the second half of #241. We left it out to keep this PR to one job, but if you would rather have it now, it is easy.

**A related thing you should know about**, since it works against you: the client fires `/upload/abort` from the `catch` of every completion failure, including the new 409s this adds. So the branch tells the user "please upload this file again" while the client has already aborted the multipart upload and freed the parts a resume would have reused. That is the abort gating you are already fixing, and this makes the path into it more routine.

**On the code itself**, the parts worth your scepticism:

- The `version.processing_status != uploading` guard in `complete_upload`. Our own review caught that without it, replaying this request against a finished version rewound it and queued a second transcode. Given `/upload/*` is exempt from the rate limiter, that was worse than what it replaces. We would like a second opinion that the guard is right and does not break a legitimate retry.
- The total-bytes check compares against `file_size_bytes`, which the client declares at initiate and nothing verifies. It catches every under-report we could produce, but it is not an independent integrity check, and we would rather someone said so out loud than have it read as stronger than it is.
- The `MultipartUploadGone` handling, which treats "the object is there at the right size" as proof a previous completion succeeded.

Everything is in the PR description with the reasoning. Tests are `test_complete_upload_parts.py` and `test_complete_upload_endpoint.py`; the second exists because our first pass only tested the helpers, and reverting the entire fix left the suite green.

Lennart-Pingpong:
sure i'd love to help. I'll look into it - if I'm not too late already. :) I'll check it out.

Lennart-Pingpong:
ok here we go — and thanks for pulling me in on it rather than letting me find it at
rebase time. Read the whole thing against `main`. Taking your three points in
order, then the two questions, then one thing I think is worth a follow-up.

## The `processing_status != uploading` guard

The guard is right, and I would keep it. Replaying a completion against a
finished version to queue transcodes is a real amplification primitive, and
`/upload/*` being exempt from the rate limiter makes it a cheap one.

But I think it currently **does break a legitimate retry**, and it does so by
standing in front of the idempotency you built for exactly that case.

The sequence:

1. Client sends `/upload/complete`. `_finish()` sets `processing` and commits.
2. The response is lost — flaky link, closed laptop, proxy timeout.
3. Client retries the same request.
4. The guard fires first and returns 409 `This upload is already processing.`

`_already_assembled()` never runs, because it sits behind `list_upload_parts()`
and the guard returns before we get there. So the one path that could say "this
already worked, you are fine" is unreachable in the very scenario it was written
for. It only triggers when the version is *still* `uploading` while the multipart
upload is gone — a narrower window (commit lost after a successful
CompleteMultipartUpload) than the one users actually hit.

What makes it more than cosmetic is what the client does with the 409. It fires
`/upload/abort` from the catch of every completion failure, and marks the entry
`failed`. Your abort handler is careful enough that nothing is destroyed — the
version is `processing`, so it is left alone, and the `_already_assembled` check
covers the `uploading` case. So the damage is not data loss. It is that **a user
whose upload succeeded is told it failed**, while the transcode of that very file
runs in the background. On a multi-gigabyte upload that is the difference between
"done" and "do it again", and they will do it again.

Suggested shape, reusing what is already there:

```python
if version.processing_status != ProcessingStatus.uploading:
    # A retry after a lost response lands here. If storage holds the finished
    # object at the declared size, this request already succeeded: report that
    # instead of a conflict -- but do not re-queue processing.
    if _already_assembled(s3_key, media_file.file_size_bytes, version.id):
        return CompleteUploadResponse(
            status=version.processing_status.value,
            asset_id=body.asset_id,
            version_id=body.version_id,
        )
    raise HTTPException(status_code=409, detail=...)
```

The replay protection survives intact: the expensive part — `_finish()`, the
status rewind and the `_trigger_processing` dispatch — still only ever runs from
`uploading`. A replayer gets a 200 that changes nothing rather than a 409 that
changes nothing, and a genuine retry stops being reported to the user as a
failure. It also costs one HeadObject on a path that is already the unhappy one.

Two details I would want in that version: return the *actual* current status
rather than hardcoding `processing`, so a version that has since gone `ready`
does not read as still transcoding; and keep the 409 for the case where nothing
is assembled, since that really is a conflict.

## The total-bytes check

Your instinct to say it out loud is right, and I would go further: it is not an
integrity check at all, it is a **consistency check of the client against
itself**. `file_size_bytes` arrives from the same browser that uploaded the
parts, and nothing corroborates it. A client that declares 2 GB and uploads a
consistent 2 GB of the wrong bytes passes cleanly.

That is fine, because it is not the threat. The bug you are fixing is a client
that *believes* it sent everything and did not — a dropped part, an aborted
worker, a browser that lost a chunk. In every one of those the declared total and
the stored total diverge, which is precisely what this catches. It converts a
silent truncation into a loud 409, and that is the whole job.

What it cannot do is detect corruption in transit, and I would make sure no
comment or doc line implies otherwise. Real integrity needs a content hash the
server can verify, and the only cheap one available is the per-part ETag —
which, as you note elsewhere, is MD5 on S3/MinIO/Garage but not guaranteed
everywhere. Worth a sentence in the docstring: *"consistency against the declared
size, not proof of content."*

One small thing: `file_size_bytes` is already load-bearing beyond this check —
`upload_guard_error` sizes the storage quota from it at initiate. So a client
lying about it is a pre-existing condition, not something this PR introduces.

## `MultipartUploadGone` → "the object is there at the right size"

I am comfortable with this one, and I think the safety comes from somewhere
slightly different than the size.

The key is `raw/{project_id}/{asset_id}/{version_id}/original.ext` — all
server-minted UUIDs. Nothing else can legitimately be sitting at that key, so
the question is never "is this someone else's object" but "is this *our* object,
finished or half-finished". For that, size is a sound discriminator: a multipart
upload that never completed leaves no object at the key at all, and a completed
one is exactly the declared length or the completion would have been rejected by
the checks above.

If you want to harden it for near-free, `head_object` also returns
`LastModified`, and requiring it to be at or after the version's `created_at`
rules out the one exotic case where a stale object at a recycled key could match
by size. I would not block on it — but the docstring currently reads as if size
is the proof, and the key layout is doing more of the work than it says.

## Question 1 — pinning chunk size server-side

I think it belongs in #241, not here, but it is load-bearing for #241 rather
than a nice-to-have, and there is a reason beyond the one you gave.

Your argument is the hazard: an upload spanning a release where `CHUNK_SIZE`
changed can never complete, because part 1 sets the expected size and later parts
disagree. Agreed, and for a resume that spans sessions that stops being
theoretical.

The stronger reason is that a resuming client **needs to know the chunk size to
slice correctly at all**. Resume answers "which part numbers still need sending",
but part 7 is only meaningful as a byte range, and the range depends on the chunk
size the upload was begun with. Today the client would infer it from its own
current constant — which is exactly the value that may have changed. It can be
recovered from `ListParts` (the size of part 1), but that is inference from data
that only exists while parts do, and it silently breaks for a resume of an upload
where part 1 is somehow absent.

So: pin it at initiate, return it from whatever endpoint resume asks, and have
the validator prefer the pinned value, falling back to deriving it from part 1
when the column is NULL. That fallback matters — #256 just added two nullable
columns for the same reason, and uploads begun before the migration will have
NULL here too.

If it is cheap to add the column alongside those two, doing it now would save a
second migration. But the *logic* can absolutely wait for #241.

## Question 2 — returning the missing part numbers as JSON

Yes, please — with one caveat about where.

A resuming client wants the missing list **before** it attempts a completion, not
as the body of a failed one. If the only way to learn what is missing is to
complete and be rejected, every resume starts with a deliberate failed request,
and the 409 path becomes load-bearing for the happy path. That is an awkward
contract, and it will read strangely in logs — "rejected completion of upload X"
firing routinely on successful resumes.

I would rather have a dedicated read: `GET /upload/parts?version_id=…` (or hung
off the version resource), returning what storage holds and what is missing.
`list_upload_parts` plus the diffing already in `_parts_from_listing` is nearly
all of it, and it is what my original issue asked for, so it is work #241 needs
regardless.

Adding the structured list to the 409 body as well costs nothing and is genuinely
useful for support — I would take both. I just would not want the error to be
the API.

## One thing I would flag for a follow-up

`_parts_from_listing` derives `chunk_size` from `parts[0]` and then requires
every non-final part to match. With the concurrency change in #250 now on `main`,
five parts are in flight at once, so a failure no longer leaves a contiguous
prefix — it can leave holes anywhere. That does not affect this validator, which
correctly rejects holes either way.

But it does affect the resume MVP as described in #241: "take the contiguous
prefix and restart at k+1" was reasoned about when the loop was strictly
sequential, and after #250 that assumption is silently pessimistic. It stays
*correct* — re-uploading parts that already exist wastes bandwidth but corrupts
nothing — so this is not urgent. It just means the MVP should read the part
numbers `ListParts` actually returns rather than assuming they form a prefix, and
the cost of doing it properly is about the same.

Happy to take any of this as PRs — particularly the guard change, which is small
and self-contained.

PR #263, opened 2026-08-16
fix(api): answer a retried completion with its status, not a conflict

Lennart-Pingpong:
The guard added in #255 refuses to complete a version that is no longer `uploading`, which is right: replaying a completion against a finished version rewinds it and queues a second transcode, and /upload/* is exempt from the rate limiter. But the same guard also catches the more common way to arrive - the first call committed `processing` and the response was lost on the way back.

_already_assembled() exists for exactly that case and was unreachable: it sits behind list_upload_parts(), and the guard returns before we get there. It only ran when the version was still `uploading` while the multipart upload was gone, a much narrower window than the one users hit.

The result was that a user whose upload succeeded was told it failed - and the client fires /upload/abort from the catch of every completion failure, so the report was not inert. On a multi-gigabyte upload that is the difference between "done" and "do it again".

Storage settles it. If the assembled object is at the key at the declared size, the request already succeeded, and saying so costs one HeadObject on a path that is by definition not the hot one. The version's own status is returned rather than a hardcoded "processing", so one that has since gone `ready` does not read as still transcoding.

Replay protection is unchanged: the status write and the transcode dispatch still only ever run from `uploading`. `failed` is deliberately excluded - that is a version somebody resolved as broken, and an object at the key must not turn it into a success.

Four tests: the retry succeeds for `processing` and `ready`, and is still refused when the version is `failed` and when the object at the key is short. The existing replay test now stubs head_object_size explicitly, since "nothing assembled" is what makes it a replay rather than a retry.

## Summary

Follow-up to the review of #255, and the one concrete change out of it.

The `processing_status != uploading` guard is right — replaying a completion
against a finished version rewinds it and queues a second transcode, and
`/upload/*` is exempt from the rate limiter. But the same guard also catches the
more common way to arrive at it: the first call committed `processing` and the
response was lost on the way back, so the client retries.

`_already_assembled()` exists for exactly that case and is currently
unreachable — it sits behind `list_upload_parts()`, and the guard returns before
we get there. It only runs when the version is still `uploading` while the
multipart upload is gone, which is a much narrower window than the one users
actually hit.

What makes it more than cosmetic: the client fires `/upload/abort` from the catch
of every completion failure and marks the entry `failed`. Nothing is destroyed —
the abort handler leaves a `processing` version alone — but **a user whose upload
succeeded is told it failed**, while the transcode of that same file runs in the
background. On a multi-gigabyte upload that is the difference between "done" and
"do it again", and they will do it again.

## Changes

- The guard asks storage before refusing: if the assembled object is at the key
  at the declared size, the request already succeeded and the endpoint says so
- Returns the version's **own** status rather than a hardcoded `processing`, so
  one that has since gone `ready` does not read as still transcoding
- `failed` is deliberately excluded — that is a version somebody resolved as
  broken, and an object at the key must not paper over the disagreement
- Replay protection is unchanged: the status write and the transcode dispatch
  still only ever run from `uploading`. A replayer now gets a 200 that changes
  nothing instead of a 409 that changes nothing
- Costs one `HeadObject`, on a path that is by definition not the hot one

## Testing

- [x] Backend tests pass — **303 passed**, full `apps/api/tests/` against a real
      Postgres with migrations applied (the `real_db` fixtures need it)
- [ ] Frontend tests + build — **not run: no frontend files touched**
- [x] Reproduced on a live v1.9.0 instance (see below)
- [ ] Tested manually in browser — staging a *lost* response honestly is awkward,
      so the replay below stands in for it; the four tests cover the state machine

**Reproduced against a real deployment.** Uploaded a file through the browser on
our v1.9.0 instance and let it finish (`ready`, transcode complete, no open
multipart upload). Then replayed the exact same `/upload/complete` request the
client had just sent:

```
HTTP 409 {"detail":"This upload is already ready."}
```

That is the response a client gets after a lost reply, and it turns into "upload
failed" in the UI for a file that is sitting there fully processed. The guard
itself behaved correctly — the version stayed `ready` and no second transcode was
queued — so this is purely about what the client is told.

Four tests, all failing before the change and passing after:

```
test_retry_against_a_finished_version_reports_success_not_conflict[processing]
test_retry_against_a_finished_version_reports_success_not_conflict[ready]
test_retry_against_a_failed_version_is_still_refused
test_retry_with_a_short_object_at_the_key_is_refused
```

Counter-checked by reverting `routers/upload.py` to `main` with the tests in
place: exactly the two retry cases fail, the other 17 in that file stay green —
including your `test_replaying_a_finished_upload_is_refused`.

That existing test now stubs `head_object_size` explicitly. It previously left it
unstubbed, which reached for a real S3 client; with this change the distinction
between a replay and a retry is precisely whether an object is assembled, so the
test states which case it is testing.

## Checklist

## Screenshots

n/a — no UI changes.


ravirajsinh45:
@Lennart-Pingpong Nice writeup, and the reproduction against a live v1.9.0 instance is convincing. One thing worth fixing before merge, plus two smaller items.

**The retry fast-path can turn a transient storage blip into a 500 instead of the intended 409.**

`_already_assembled()` now runs on the common retry-after-lost-response path (any version that's `processing`/`ready`), not just the narrow race window it used to guard. Both it and `head_object_size()` only catch `botocore.exceptions.ClientError`:

```python
def head_object_size(s3_key: str) -> int | None:
    ...
    try:
        return s3.head_object(Bucket=settings.s3_bucket, Key=s3_key)["ContentLength"]
    except ClientError as e:
        ...
```

Connection-level failures like `EndpointConnectionError`, `ConnectTimeoutError`, and `ReadTimeoutError` are `BotoCoreError` subclasses, not `ClientError` subclasses, so they aren't caught. If storage is briefly unreachable or slow during that HeadObject call, the exception now propagates out of `complete_upload` as an unhandled 500, on a branch this PR made much more frequently hit than before. That works against the goal here, since the client's catch still fires `/upload/abort` on a 500.

Worth widening the except in `_already_assembled` (and maybe `head_object_size`) to cover `botocore.exceptions.BotoCoreError`, and adding a test that stubs `head_object_size` to raise a connection error to confirm it degrades to 409 rather than 500.

Two smaller ones, neither blocking:

1. The retry path now costs a live `HeadObject` per replay against a finished version, and `/upload/*` is exempt from the rate limiter specifically because that loop was assumed to be a free DB check. You already call out the cost in the PR description, so this is more "confirming that's accepted" than a bug.
2. The `status` in the success response is read before the `_already_assembled` round-trip, so it can be a beat stale if a transcode flips `processing` to `ready` while the HeadObject call is in flight. Informational field only, no real side effect.

Lennart-Pingpong:
Good catch, and thank you for reading it that closely — the first one is a real
hole, and it is one this PR opened. Fixed in `c66ec1f`, pushed as a second commit
rather than an amend so your review still anchors to what you read.

## The 500 on a storage blip

You are right about the class hierarchy and about the consequence. A refusal
*from* the backend is a `ClientError`; never reaching it at all is a
`BotoCoreError`, and the two are siblings rather than parent and child, so the
old `except ClientError` let `EndpointConnectionError`, `ConnectTimeoutError` and
`ReadTimeoutError` straight through. That was survivable while
`_already_assembled()` only ran in the narrow "still `uploading`, multipart
upload gone" window. On the retry-after-a-lost-response path it is not, and your
point that a 500 is no softer than the 409 is the one that settles it — the
client fires `/upload/abort` from the catch of *any* completion failure.

`_already_assembled()` now catches both, and falls through to the same 409 it
would have produced before this PR:

```python
    try:
        return head_object_size(s3_key) == expected_bytes
    except (ClientError, BotoCoreError):
        # A HeadObject we cannot complete says nothing either way, and guessing wrong
        # here either loses a finished upload or reports a missing one as done.
        ...
```

Three test cases, one per connection error, asserting the 409 *and* that neither
the status write nor the transcode dispatch ran:

```
test_retry_survives_a_storage_blip_as_a_conflict_not_a_crash[blip0]  # EndpointConnectionError
test_retry_survives_a_storage_blip_as_a_conflict_not_a_crash[blip1]  # ConnectTimeoutError
test_retry_survives_a_storage_blip_as_a_conflict_not_a_crash[blip2]  # ReadTimeoutError
```

Counter-checked by reverting only the `except` line with the tests in place:
exactly those three fail, each with `assert 500 == 409`. Full suite green —
**306 passed**, `apps/api/tests/` against a real Postgres with migrations
applied. No frontend files touched.

**`head_object_size()` I deliberately left raising.** Its contract is "the size,
or `None` if the object is not there", and swallowing a connection failure into
`None` would be a lie every caller inherits — "storage was unreachable" would
arrive as "the object does not exist". `_already_assembled()` is the layer where
"we could not find out" has a defined meaning, and it is the only caller, so
widening it there covers every path that reaches storage this way. Happy to move
it if you would rather have the narrower blast radius at the boundary.

## 1. The HeadObject per replay

Confirming that is accepted, yes — but if the rate-limiter exemption is what
makes you uneasy, there is a cheap way to make a *replay loop* cost nothing while
a genuine retry still pays the one call. Since #256 the version carries its own
`upload_id`, and `/upload/presign-part` already gates on exactly that:

```python
if version.upload_id is not None and version.upload_id != body.upload_id:
    raise HTTPException(status_code=403, detail="Not authorized for this upload")
```

The same test in front of the storage call would do it. A retry replays the
request it already sent, so its `upload_id` matches by construction; a loop with
a made-up one is refused from the database, as it is today. Rows predating #256
have `upload_id` NULL and would fall through to the HeadObject, same as presign
accepts them.

I have not pushed that — it widens a PR you are already reviewing, and it is your
call whether it belongs here or in its own change. Say the word and it is a
three-line commit either way.

## 2. The stale `status`

Agreed, and I do not think it is worth closing. It is inherent rather than
incidental: a transcode can flip `processing` to `ready` after the field is read,
after a `db.refresh()` would have re-read it, and after the response is
serialised — a refresh would only narrow a window that cannot be shut, at the
cost of another query on the path we are already trying to keep cheap.

Worth adding as a data point: the web client discards the response body outright
at both call sites (`apps/web/stores/upload-store.ts` — `await api.post('/upload/complete', …)`,
result unused), and the real status arrives over SSE afterwards. So the field is
informational by construction today, not just by intent. If you would rather it
could not be stale at all, the honest fix is to drop it from the response instead
of refreshing it — but that is a breaking change for anyone driving the API
directly, so I would leave it.

ravirajsinh45 (dismissed):
Thanks @Lennart-Pingpong, this is careful work and the diagnosis is right: `_already_assembled` was written for exactly the lost-response case, and the #255 guard made it unreachable. The commit messages explain reasoning rather than restating the diff, which we appreciate.

We went through it in detail. There is one structural question about where the fix belongs, then a set of things we would want resolved before merging.

## The symptom this describes lives in the client

`apps/web/stores/upload-store.ts:498` awaits `/upload/complete` once. On any failure it goes straight to the catch, sets `status: 'failed'` and fires a fire-and-forget `/upload/abort`. The only automatic replay in `apps/web/lib/api.ts:48-53` is the 401 token refresh, which by definition never reached the handler body, and `uploads-panel.tsx:172` offers a failed row a Dismiss button and nothing else. #259 says this outright in its commit message: the shipped client never retries complete, it goes straight to abort.

So the sequence in the PR body (completion succeeds, response is lost, user is told it failed) plays out identically after this change, because nothing ever sends the second `/upload/complete` that the new branch answers. `refreshProcessingItems` only reconciles rows already in `processing`, and `partialize` persists `failed` rows, so the false Failed also survives a reload.

That does not make the server change worthless. It is the right answer for any non-web API client, and it has to land before a client-side retry would be safe to add. But the user-visible fix is the client not marking `failed` on a completion error without re-reading the version first, and that one also helps against servers that have not been upgraded yet. We are happy to take it as a follow-up if you would rather keep this PR focused, we would just want the PR body and the CHANGELOG entry to stop claiming the user-visible symptom is fixed here.

## Blocking

**1. `upload.py:329`, `body.asset_id` is never checked against `version.asset_id`.** The handler validates `body.version_id` (271) and `body.s3_key` (284) but not this one, and `_finish()` passes it straight to `process_asset`, which resolves the asset with no cross-check and builds `output_prefix = processed/{asset.project_id}/{asset_id}/{version_id}` plus an SSE publish to that project. This predates your PR, but the new response is a second consumer of the field, and `assets.py` already does it correctly at `upload.py:423` with `version.asset_id`. A `400` when `body.asset_id != version.asset_id` near the top of the handler closes both. Please also return `version.asset_id` in the new response.

**2. `upload.py:327`, the 200 asserts a transcode is under way without checking that one is.** `_dispatch_task` (`tasks/celery_app.py:100-111`) swallows every dispatch failure with a `logger.warning`, and `_reap_stale_uploads` only looks at `uploading` and `failed` (`tasks/cleanup_tasks.py:240`), so a version stranded at `processing` with no task has no recovery path and keeps counting against the storage cap. Before this PR the retry got a 409, the user saw Failed and re-uploaded. After it, the answer is `200 {"status":"processing"}` forever. The `MultipartUploadGone` path 25 lines down re-dispatches, and `abort_upload` at 462 does too, so the new branch is the odd one out. `assert dispatched == []` in the new test locks the gap in.

**3. `upload.py:320`, this puts an untimed storage call on the one endpoint family exempt from the rate limiter.** `get_s3_client()` is built from `_SIGV4_CONFIG` / `_NON_AWS_COMPAT_CONFIG` (`services/s3_service.py:39,45`), and neither sets `connect_timeout`, `read_timeout` or `retries`. Only `_STARTUP_S3_CONFIG` does, added precisely so a slow store cannot hang on boto3's ~60s default. So worst case is roughly 5 attempts at 60s. `complete_upload` is a sync `def`, so it holds an anyio threadpool slot, and `get_db` keeps the pooled connection open for the whole call, with a 5+10 pool per worker and no pool args in `database.py:11`. `middleware/global_rate_limit.py:40` short-circuits `/upload/`, which is the same exemption your comment cites, so an authenticated user can drive this unthrottled. A short per-call timeout (a couple of seconds, `retries` capped) would keep the check honest. The blip test stubs an instant raise, so it proves the 409 but never that the answer arrives in bounded time.

**4. `upload.py:185`, widening the catch changes `/upload/abort` as well, and neither the PR body nor the CHANGELOG mentions it.** The comment at 443-447 says the check exists because marking a version failed hands it to the reaper, which deletes the finished file a day later. After the widening, "we could not reach storage" returns the same `False` as "the object is definitely not there", so abort commits `failed` (464) and returns 204 with nothing operator-visible, and `_reap_stale_uploads` deletes `s3_key_raw` 24h later (`cleanup_tasks.py:263`). Before the change that `BotoCoreError` propagated as a 500 and the version stayed `uploading`, from which the `MultipartUploadGone` path could still rescue it. The reasoning for the widening is sound, but "unknown" needs to be its own answer: a tri-state out of `head_object_size` (present / absent / unknown) would let abort decline to decide instead of deciding wrong.

**5. `upload.py:346`, `list_upload_parts` has no storage-error handling at all, and it runs on every genuine completion.** The `try` catches only `MultipartUploadGone` and `MultipartListingUnsupported`; `list_upload_parts` re-raises everything outside `UPLOAD_GONE_CODES`, and `BotoCoreError` was never caught. A transient `SlowDown` or `RequestTimeout` at the end of a multi-hour upload escapes as a 500, the client fires `/upload/abort`, abort correctly reports not-assembled because `CompleteMultipartUpload` never ran, the version commits `failed`, and the reaper deletes a file that transferred intact. Same gap as item 6, on the hotter call.

**6. `upload.py:389` and `:430`, the sibling `except ClientError` blocks were not widened.** Your own justification names `/upload/abort` as the endpoint that must not 500, and 430 is that endpoint: `abort_multipart_upload` is a bare passthrough (`s3_service.py:266-273`), so an `EndpointConnectionError` escapes before the status write at 442-465 and leaves the version `uploading` for the reaper. Line 389 is the same shape on `complete_multipart_upload`, where a `ReadTimeoutError` on a completion that actually succeeded skips the `_already_assembled` recovery on 393.

## Worth resolving in this PR

**7. `upload.py:320`, the gate ignores `body.upload_id`.** `version.upload_id` is persisted at initiate (111, and `assets.py:386`), is never cleared anywhere, and is already the authorization check in `presign_part` (163). Comparing it here separates a genuine retry from a blind replay from DB state alone, with no network call, and would let the endpoint replay the response rather than synthesize one. As written the new test enshrines the weak form by posting `upload_id="junk"` and asserting 200, and the only other signal, `media_file.file_size_bytes`, is a number the client chose. Worth noting that `models/asset.py:61-64` claims `upload_id` is NULL once an upload completes or aborts; nothing implements that, which is why the discriminator is still available.

**8. `upload.py:311`, the status guard is an unsynchronized read.** There is no `with_for_update()` and no conditional update, though the codebase uses row locks elsewhere (`cleanup_tasks.py:164,355`). Two concurrent requests both read `uploading` and pass; A completes and calls `_finish()`; B's `complete_multipart_upload` then fails `NoSuchUpload`, lands in the `ClientError` branch at 389-394 where `_already_assembled` is now True, and calls `_finish()` too. Two `process_asset` runs write the same `processed/...` prefix and each publishes `transcode_complete`. An `UPDATE ... WHERE processing_status='uploading'` with a rowcount check makes the invariant your comment states actually hold.

**9. `upload.py:332`, 409 is the wrong code for the degraded path.** When `_already_assembled` returns False because storage was unreachable, the condition is explicitly unknown and transient, but 409 tells the client it will never succeed, and the body ("This upload is already processing.") contradicts the status line. A retryable 503 or 504 with `Retry-After`, kept distinct from the 409 that means the object is genuinely not there, is what "we could not find out" wants. As it stands, the exact harm the PR exists to remove survives on the path the PR introduces, and `test_retry_survives_a_storage_blip_as_a_conflict_not_a_crash` asserts it as intended behaviour.

**10. `upload.py:316`, `failed` is not a decision anyone made.** `process_asset` is `bind=True, max_retries=3, default_retry_delay=60`, and its except block commits `failed`, publishes `transcode_failed` and then retries, re-entering at 52 with `processing` (`tasks/transcode_tasks.py:75-82`). The row oscillates for about three minutes with the raw object intact, so byte-identical retries get 409 or 200 purely on timing. And for a version that exhausts its retries with the object sitting at the key at the right size, the exclusion refuses it forever and there is no reprocess endpoint, so the only recovery is a full re-upload. Either treat an assembled `failed` version as re-dispatchable, or say in the 409 detail that it needs a re-upload; the comment currently claims a human decision the state machine does not record.

## Smaller notes

**11. `services/s3_service.py:369`, `head_object(...)["ContentLength"]` is an unguarded subscript.** `KeyError` is neither a `ClientError` nor a `BotoCoreError`, so it goes through the widened catch as a 500, now on the common retry path. `HeadObjectOutput` has no required members, so botocore omits the key when the header is absent, which happens behind proxies that strip or chunk-encode HEAD responses. The adjacent code already knows this: `list_upload_parts` uses `p.get("Size")` and `_parts_from_listing` has an explicit branch for backends that do not report sizes. `resp.get("ContentLength")` is the fix.

**12. `upload.py:328`, the status is a snapshot taken before the HeadObject.** `version` loads at 271, the session is `autoflush=False` with no intervening refresh, and the transcode runs in a separate process, so 316, 323, 328 and 334 all read the pre-call value across a network round trip. A transcode that finishes during the call still reads `processing`, which is the case the comment says it fixes, and one that fails during the call still evaluates `completed_before` as True, returning 200 for a version that is now `failed`. A `db.refresh(version)` after the storage check settles both.

**13. `CHANGELOG.md:9`.** The entry adds a second `### Fixed` under `[Unreleased]`, which is what makes the PR conflicting; the existing one is three lines below. `docs/RELEASING.md:19-23` also wants a trailing `(#263 by @Lennart-Pingpong)` so the release cut can build the Contributors line. And since this is precisely and only an API contract change for non-web clients, it wants the same `### Upgrade notes` treatment v1.9.0 gave its analogous change, plus a correction to `CHANGELOG.md:44` and `docs/architecture.md:68`, which both still document the old behaviour.

**14. `tests/test_complete_upload_endpoint.py`.** Four things, all small:
- The new tests re-hardcode `mock_db.first.side_effect = [version, media_file]` instead of using the file's own `upload_rows` fixture (24) and `_stub()` (59), taking that positional script from two copies to six. A third query in the handler then breaks all six, and opaquely, since a short `side_effect` raises `StopIteration` and `raise_server_exceptions=False` turns it into a bare 500.
- `test_retry_against_a_failed_version_is_still_refused` and `test_retry_with_a_short_object_at_the_key_is_refused` assert only the status code and, unlike their two siblings, do not patch `_trigger_processing`. The two properties the guard actually protects, `version.processing_status` unchanged and `dispatched == []`, go unasserted, and a fall-through would run the real `send_task_safe` and spawn a broker-connecting thread instead of failing.
- In `test_replaying_a_finished_upload_is_refused`, the `list_upload_parts` sentinel now guards a call that path no longer makes, and its docstring still explains the rate-limiter exemption. After this change no test asserts the replay path is storage-free.
- The blip `parametrize` builds live exception instances at collection time and `raise blip` re-raises the same objects, so `__traceback__` grows on every raise while `exc_info=True` formats it. Parametrizing over the classes with `ids=` avoids that under `--lf` or repeat runs.

Happy to talk through any of these, particularly items 1 through 6 and the client question at the top. Thanks again for digging into this one, the underlying problem is real and worth fixing properly.

ravirajsinh45:
@Lennart-Pingpong first, an apology, and the reasoning behind it.

I pushed the review fixes directly onto your branch a few hours after leaving the review, before you'd had any real chance to respond. That was our call and it was a call about speed, not about your work or your response time — we wanted the upload path closed out in one pass and chose not to wait. You deserved the chance to do this yourself and I took it off you.

The fixes did turn out to interact, which is a good reason to land them as one commit but not a good reason to skip the author, so I don't want to hide behind it. Your commits are still here, the PR is still yours, and mine sits on top as a co-authored commit. If you'd rather own any part of this, I'll drop my commit and hand you the review notes to work from — that offer is real, not a formality.

## What changed

Your diagnosis was right and the shape of the fix is unchanged. What the review turned up is that storage answers three things, not two, and the third was being collapsed into "the object is not there":

- `_already_assembled()` returns `None` for "could not find out". Each caller now says what it means by that. A retried completion answers `503` with `Retry-After` instead of a `409`, because `409` tells the client it can never succeed and ours treats that as final. An abort leaves the version alone rather than recording a failure it cannot support, which matters because the reaper deletes a `failed` version's raw object a day later.
- Three storage calls had no handling for *never reaching* the backend. `list_upload_parts` was the worst: it runs on every genuine completion, not just a retry, so a transient `SlowDown` at the end of a long transfer became a 500, then an abort, then a deleted file.
- The HeadObject was unbounded, on the one endpoint family the rate limiter exempts. It runs on a separate client with short timeouts now, kept apart from the shared one because those timeouts would be wrong for `CompleteMultipartUpload`.
- A retry is told apart from a replay by `version.upload_id`, which `presign-part` already checks. That decides the replay case from stored state with no storage call, which restores the property your original replay test was asserting.
- The dispatch was fire-and-forget into a thread that swallowed failures, with the version already committed as `processing`. Nothing sweeps that state, so a broker outage stranded an upload permanently, and the only thing that ever surfaced it was a retry coming back `409` — which this PR deliberately stops doing. `send_task_safe` takes an `on_failure` now.
- `failed` was written before *each* Celery retry, so the row oscillated for the whole ladder while the file sat there intact. Your comment described `failed` as a decision somebody made; that is now true.
- The `uploading` check was an unsynchronised read that two concurrent completions could both pass. The version is claimed with a conditional update.

One of your tests is inverted rather than extended: `test_an_unreadable_head_object_does_not_claim_success` asserted that an unreadable HEAD marks the version `failed`. I checked the reaper before changing it — it deletes `s3_key_raw` for `uploading` and `failed` alike without re-checking, so leaving the version `uploading` has the same worst case but keeps a window in which a later call that does get an answer can rescue it. That reverses a decision you had documented, so push back if you see it differently.

## Worth knowing

This still is not reachable from the web client. It awaits `/upload/complete` once and goes straight to abort, so nothing sends the second request this branch answers. We fixed the client separately in #269 by re-reading the version's status before writing an upload off — a re-read rather than a retry, so it works against un-upgraded servers too. Adding a bounded client retry on top is the natural next step, and it is the thing that would finally make this branch's new path live. Happy to hand you that one if you want it.

13 tests fail against the previous code and pass against this. Thanks for the PR: the underlying bug was real, and neither the storage-blip class nor the stranded-dispatch problem would have surfaced without it.

ravirajsinh45:
Correction to my previous comment. I had an independent review run over what I pushed to your branch, and four of the changes I described to you as improvements were wrong. I have reverted them. I am sorry — I took the PR over on the grounds that I would finish it properly, and then described work to you as better than it was.

## The one that mattered

`_record_dispatch_failure` marked a version `failed` when the broker could not be reached. By that point `complete_multipart_upload` has already succeeded, so the raw object is fully assembled in storage — and `_reap_stale_uploads` sweeps `failed` and `uploading` on identical criteria and deletes `s3_key_raw`. A twenty-second Redis restart after a 40 GB upload would have scheduled that master for deletion 24 hours later.

That is the exact outcome the `_already_assembled` docstring calls unacceptable when a HeadObject causes it. I wrote that docstring and then implemented the same thing for the broker case, forty lines away.

## Also reverted

- **The `process_asset` retry change.** It removed one route into a permanent `processing` strand and added another: `self.retry()` raises `Reject` when the enqueue itself fails, and I wrote nothing on that path.
- **The conditional claim UPDATE.** The code is correct, but every test covering it used a `MagicMock` session, so deleting the `WHERE` clause entirely left the suite green. It needs the `real_db` fixture and two connections.
- **The abort tri-state.** This one was built on a false premise of mine. I argued that marking a version `failed` on an unreadable HeadObject was destructive compared to leaving it `uploading`. It is not — the reaper takes both on the same schedule, so the file dies either way. The only difference is Active versus Failed in the panel, and Failed is the better label. Your original test asserted the right thing and I inverted it for no gain; it is restored.

## What is on the branch now

Your two commits, plus one from me that keeps only the storage-lookup half:

- `ObjectSizeUnavailable`, so a HeadObject that returns no `ContentLength` stops raising an unhandled `KeyError`
- the `BotoCoreError` catch from your second commit, plus the missing handling on `list_upload_parts` and `complete_multipart_upload`
- a bounded probe client, so the lookup stops inheriting boto3's ~60s-per-attempt defaults on a rate-limiter-exempt endpoint
- your 200-instead-of-409 retry answer

Two corrections inside that half. The tri-state was applied at three of five call sites — the `MultipartUploadGone` branch, the path this PR exists for, still answered a failed lookup with a 409, which is what the guard forty lines above argues destroys versions. There is one `_storage_unreachable()` now and every path uses it. And the two `failed`-carve-out tests were vacuous: both sent a mismatched upload id, so they exited on that check and never reached the storage lookup they claimed to cover. Both passed unchanged against `main`. They reach it now.

I also dropped the claim that the upload-id check separates a retry from a replay. It does not — a replay of a captured request carries the correct id. It is a filter, and the comment says so.

Three mutations that previously left the suite green now fail: adding `failed` to `completed_before`, routing the gone-branch `None` back to 409, and letting abort treat `None` as assembled.

## Follow-ups

The four reverted items are real problems, they just need to land separately with tests that are not mocks: a `processing`-with-no-task sweeper that re-dispatches rather than one that marks failed, the retry-status change with the dropped-retry case covered, and the claim UPDATE proven against a real database. I will open those.

Your original diagnosis was right and it still is. Everything above is a critique of my follow-up commit, not of your PR.

PR #264, opened 2026-08-20
fix(web): make the branding controls actually take effect, and preview them properly

ravirajsinh45:
Follow-up to #202 and #203, both now merged. Those two landed the correctness fixes; this is the design and wiring work that came out of testing the merged result on a live instance, kept separate so neither contributor PR grew a second agenda.

@rubenxyz a review here would be very welcome, since most of this touches the branding feature you built. Two commits also landed directly on your PRs before merge, disclosed in comments on each; if anything there looks wrong, say so and I will change it.

## Why this exists

Running #202 and #203 together against a real stack turned up a set of problems that were not review findings on either PR, mostly things that only show once the feature is exercised end to end.

## Two controls that did nothing

**The accent colour** was written to a `--ff-primary` CSS variable. Nothing in the app read it; the only other mention of it in the entire web app was the settings copy describing it. So the picker validated a hex, persisted it, and changed nothing on screen.

It now sets `--accent` plus derived hover and muted shades, which is what every accent surface actually resolves through: `bg-accent` appears in 47 files, including the download button and the post-comment button a guest sees on a share link. The hover shade is derived from the colour's brightness rather than being a fixed tint, so a pale accent darkens on hover instead of producing a hover state nobody can see.

**A custom favicon** never replaced the default. Next emits its own `<link rel="icon">` for `app/icon.png` and serves `app/favicon.ico` at `/favicon.ico`, which browsers request by convention. Appending one more link left the custom icon as one candidate among several, and it generally lost. Any icon link the app does not own is now removed first.

## Share links ignored their own appearance

A link's theme and accent were only ever applied on the folder grid. The single-asset viewer was never passed `appearance` at all, so a link set to the light theme with its own accent opened dark in the instance's colours, with nothing to indicate the settings had been dropped.

The viewer was also authored dark-only, so simply setting `data-theme="light"` produced a half-flipped page: light chrome, white-on-white text, and white-alpha borders that vanished against a light background. 52 fixed colours are now theme tokens. The download button and the guest avatar chip were a fixed purple, so a link's accent repainted everything except the control people actually look at.

Both viewers now share one `useShareAppearance` hook. Having the logic inside the folder viewer is precisely why the other one never got it.

The instance accent is deliberately not marked `!important`, so a link's own accent still wins on its own page. There is a test asserting that, because making them important would silently break per-link accents.

## The brand mark was a placeholder

`logo-icon.svg`, `logo-icon-dark.svg` and `logo-full.svg` were hand-drawn "FF"-in-a-box placeholders, and every logo fallback pointed at them, even though the real artwork was already in the repo. `public/logo-icon.png` is byte-identical to `app/icon.png`, the favicon. So resetting branding fell back to a placeholder rather than the actual mark.

Every fallback now points at the real artwork, and the attribution badge uses the favicon shape with no wordmark, since the line already reads "Powered by FreeFrame" and the old glyph said the name twice. It links to this repository. The badge also moved out of the sidebar's 48px logo header, where it competed with the org name for the row that exists to show the brand, and now floats bottom-right over the content.

## The settings tab

The preview was a strip of the elements the controls below edit rather than the screen an admin is trying to picture. It had also never actually rendered: the component was written but never mounted, so wiring it in surfaced two latent bugs at once, the org name and badge colliding in one inline container, and the logo drawing twice.

It is now a scaled replica of the whole screen, authored at a literal desktop size so its proportions are the real app's by construction, with three screens: the app, the sign-in page, and an email. Sign-in resolves the logo differently, so the dedicated sign-in slot had nowhere to be seen before saving, and email is the one branding surface an admin cannot check without sending themselves one.

The tab had six near-identical heading-plus-card sections in a narrow column and five logo slots as identical full-width rows carrying four lines of text each. Now three groups, a grid for the slots, and three lines per slot. The name field was labelled "Greeting" and filed under Logos & icons; it is the workspace name and now says so.

## Verified

CI aside, this was exercised against a running stack: 328 frontend tests, 333 backend tests, `tsc` clean, lint unchanged from baseline, build clean.

## Known, deliberately not fixed here

Accent buttons render their label with `text-text-inverse`, which is hardcoded across those 47 files. A very pale accent will therefore have poor label contrast. Fixing it properly means deriving a readable foreground from luminance and threading it through every call site, which is a bigger change than this and wants its own decision: constrain the picker, warn in the UI, or do the full job.


ravirajsinh45:
@rubenxyz flagging this for you directly, since GitHub will not let me add you as a formal reviewer (review requests are limited to repo collaborators).

Most of this sits on top of the instance branding feature from #203, and a fair amount of it changes decisions you made, so your read would be the useful one. In particular:

- The attribution badge moved out of the sidebar logo header to a floating pill, and now links to the repo.
- The branding preview was rebuilt around a scaled replica of the real screens rather than the element strip.
- The settings tab was regrouped, and the `guidance` prop on the logo slots was removed entirely.

No obligation at all, and none of it is a criticism of the original: the accent colour and favicon problems in here were latent long before your PR, and the preview bug only appeared because your component finally got mounted.

rubenxyz:
Read through this after the merge — no objections from me on any of it. The floating attribution pill is a nicer pattern than the sidebar badge, the rebuilt preview is a real improvement, and wiring BrandingPreview in (rather than deleting it) was the right call. The accent-colour and favicon work closes the two gaps that mattered most. Thanks for flagging it.

PR #277, opened 2026-08-21
fix(web): polish branding fallback surfaces

### LINKED ISSUE #275: Branding polish: accent-muted alpha, contrast, logo pairing, touch icon, badge z-index
Follow-ups from the review of #264, none of which blocked the merge. Several are self-contained and good candidates for a first contribution.

- [ ] **`--accent-muted` went from opaque to 18% alpha** (`apps/web/lib/accent.ts`). `::selection` (`globals.css`) is now markedly fainter, and every `bg-accent-muted` chip picks up whatever sits behind it (`avatar.tsx`, `collection-card.tsx`, `guest-comment-input.tsx`). `upload-zone.tsx` compounds it with `bg-accent-muted/30`, an opacity modifier on a bare `var()` colour that Tailwind v3 cannot compose.
- [ ] **`bg-accent-muted text-accent` is an undisclosed contrast surface.** #264 documented contrast only for `text-text-inverse` on accent buttons. This pairing appears in `projects/page.tsx`, `notifications/page.tsx`, `header.tsx`, `asset-metadata.tsx`, `guest-comment-input.tsx` and `share/[token]/page.tsx`. A pale accent gives pale text on a pale 18%-alpha background.
- [ ] **White-on-transparent default marks render unpaired in three places.** `logo-icon.png` and `logo-full.svg` need the `.logo-dark`/`.logo-light` pair, which `sidebar.tsx` does correctly. Missing in `app-shell-mock.tsx` (twice — the preview shows a white-on-white mark in light theme), `branding-logo-upload.tsx` (its `defaultUrl` is never passed, so the "Logo (light background)" slot shows a blank swatch on `previewBg="bg-white"`), and the share password-gate card in `share/[token]/page.tsx`.
- [ ] **The default apple-touch-icon is displaced.** `branding-head.tsx` removes any `apple-touch-icon` it does not own and points it at `/logo-icon.png`, replacing the purpose-built `app/apple-icon.png`. iOS composites onto an opaque white tile, so the un-branded home-screen icon likely renders blank.
- [ ] **The attribution badge collides with the toast stack.** `app/(dashboard)/layout.tsx` anchors it `fixed bottom-4 right-4 z-20`; `components/shared/toast.tsx` uses the identical anchor at `z-[9999]`. Every toast renders on top of it, and the badge is an `<a>` with default pointer-events, so it intercepts clicks in roughly a 150×30px region at the bottom-right of every dashboard page.
- [ ] **`branding-head.test.tsx` covers a state production cannot reach.** It asserts that clearing the colour hands `--accent` back to the stylesheet, but `syncBranding` no longer yields that path in the way the test drives it. It gives false confidence about the exact behaviour that was broken before #264's fix.

luozejian:
## Summary
- make custom muted accents opaque and repair contrast on affected branding surfaces
- use paired light/dark default marks in branding previews and the share password gate
- restore the dedicated Apple touch icon and move toasts away from the attribution badge
- cover branding synchronization and the Apple icon fallback with regression tests

Fixes #275

## Validation
- `pnpm test` (44 files, 338 tests)
- `pnpm exec tsc --noEmit`
- `pnpm lint` (passes with existing warnings)
- `pnpm build` (exits 0; Next emitted a non-fatal standalone trace-copy warning)

## Screenshots
Not included: this environment did not run the full authenticated application stack. The visual fallbacks and tokens are covered by targeted regression tests.


apps/web/components/shared/toast.tsx
```diff
@@ -129,7 +129,7 @@ function ToastContainer({ toasts, onDismiss }: ToastContainerProps) {
     <div
       aria-live="polite"
       aria-label="Notifications"
-      className="pointer-events-none fixed bottom-4 right-4 z-[9999] flex flex-col-reverse gap-2"
+      className="pointer-events-none fixed right-4 top-4 z-[9999] flex flex-col-reverse gap-2"
```
ravirajsinh45: Moving the toast container to top-right (`right-4 top-4 z-[9999]`) puts it directly on top of the dashboard header and its Search/panel-toggle buttons (`z-20`). Every toast now blocks those controls for its duration. Move it back to bottom-right, or pick a slot that doesn't overlap the header.

apps/web/components/__tests__/branding-head.test.tsx
```diff
@@ -87,56 +94,37 @@ describe('BrandingHead accent wiring', () => {
...
-  it('falls back to the real mark when no favicon is set, rather than none', () => {
-    useBrandingStore.setState({
-      primaryColor: null,
-      faviconUrl: null,
-      appleIconUrl: null,
-      orgName: 'Acme',
-      loaded: true,
-      loading: false,
-    })
+  it('uses the dedicated Apple home-screen icon when no custom icon is configured', () => {
```
ravirajsinh45, line 106: The rewrite dropped the assertion that `link[rel="icon"]` falls back to `/logo-icon.png` (`DEFAULT_FAVICON`). Since this PR splits one shared default into `DEFAULT_FAVICON`/`DEFAULT_APPLE_ICON`, we need coverage on both halves, not just the new apple-touch-icon one. Please restore it.

apps/web/lib/accent.ts
```diff
@@ -38,6 +38,9 @@ export function accentVars(color: string | null | undefined): Record<string, str
   return {
     '--accent': color,
     '--accent-hover': shift(rgb, luminance(rgb) > 0.6 ? -0.14 : 0.16),
-    '--accent-muted': `rgb(${rgb[0]} ${rgb[1]} ${rgb[2]} / 0.18)`,
+    // Keep muted accents opaque. An alpha color lets the surface behind each chip
+    // change its appearance; blending with the theme's secondary surface gives
+    // every consumer one stable, readable color instead.
+    '--accent-muted': `color-mix(in srgb, ${color} 18%, var(--bg-secondary))`,
```
ravirajsinh45, line 44: `--accent-muted` went from translucent (`rgb(r g b / 0.18)`, blends with any backdrop) to opaque `color-mix(in srgb, {accent} 18%, var(--bg-secondary))`. That assumes every `bg-accent-muted` consumer sits on `--bg-secondary`. A few don't (notification rows only when unread, the header's `bg-bg-primary/90` toggle, the guest-comment-input badge on `bg-bg-primary`). Take a visual pass over those three before merge; if they look fine, ignore this.

apps/web/app/share/[token]/page.tsx
```diff
@@ -181,7 +181,12 @@ function PasswordGate({ onSubmit, error, loading }: PasswordGateProps) {
             <img src={displayLogo} alt={orgName} className="h-10 object-contain" />
           ) : (
             // eslint-disable-next-line @next/next/no-img-element
-            <img src={`/logo-icon.png`} alt="FreeFrame" className="h-10 w-10" />
+            <>
+              {/* eslint-disable-next-line @next/next/no-img-element */}
+              <img src="/logo-icon.png" alt="FreeFrame" className="logo-dark h-10 w-10" />
```
Comment (ravirajsinh45) [inline apps/web/app/share/[token]/page.tsx]: `PasswordGate` already resolves `theme` one line above and uses it for the branded case. This default-logo fallback ignores that and renders both `/logo-icon.png` and `/logo-icon-dark.png`, hiding one with CSS, so we fetch two PNGs where one would do. Use `theme` to pick the single src instead.

apps/web/components/settings/app-shell-mock.tsx
```diff
@@ -47,7 +47,12 @@ export function AppShellMock({ orgName, logoUrl, onLogoError }: MockProps) {
             />
           ) : (
             // eslint-disable-next-line @next/next/no-img-element
-            <img src="/logo-icon.png" alt="" className="h-7 w-7 shrink-0 object-contain" />
+            <>
```
ravirajsinh45, line 50: The logo-dark/logo-light JSX block is now duplicated across 6 files (3 existing, 3 new in this PR). File a follow-up to extract a shared component. This would've caught a couple of the issues flagged elsewhere in this review for free.

apps/web/components/__tests__/branding-head.test.tsx
```diff
@@ -1,25 +1,35 @@
...
-function setBranding(primaryColor: string | null) {
-  useBrandingStore.setState({
-    primaryColor,
-    faviconUrl: null,
-    appleIconUrl: null,
-    orgName: 'Acme',
-    loaded: true,
-    loading: false,
-  })
+function branding(primaryColor: string | null, overrides: Partial<InstanceBranding> = {}): InstanceBranding {
```
ravirajsinh45: New `branding()` fixture duplicates fixture logic already in `stores/__tests__/branding-unconfigured.test.ts` (which uses a less type-safe `as never` cast there). Not blocking, but consolidate when convenient so `InstanceBranding` field changes don't need updating twice.

ravirajsinh45 (changes requested):
Thanks for this, @luozejian. The accent-contrast fix and the paired light/dark default-logo pattern are the right approach, and the regression tests are solid. Left inline comments on the spots within this diff that need fixing before we merge.

A few more spots share the same patterns this PR is fixing, but fall outside the diff hunks so GitHub won't let me anchor a comment on them directly. Listing them here, all should be folded into this PR:

**Same low-contrast `text-accent`/`bg-accent-muted` pairing this PR fixes elsewhere:**
- `apps/web/app/share/[token]/page.tsx:1194` (password-gate lock icon, in this same file where you already fixed the avatar circle at line 265)
- `apps/web/components/shared/avatar.tsx:46` (fallback initials, used in 8+ places app-wide)
- `apps/web/components/settings/branding-preview.tsx:71` (the live branding preview panel itself)
- `apps/web/app/(dashboard)/settings/admin/page.tsx:253`, `settings/notifications/page.tsx:102`, `settings/profile/page.tsx:100` (all three settings page header badges)

**Missing `defaultUrl` on `BrandingLogoUpload`, same as the `logo_light`/`logo_dark` slots you fixed:**
- `apps/web/components/settings/branding-tab.tsx:416` (`login_logo` / sign-in logo slot, our highest-traffic unauthenticated screen)
- `apps/web/components/settings/branding-tab.tsx:404` (`apple_icon` slot, doesn't reflect the new dedicated `/apple-icon.png` fallback)

**Unpaired default-logo fallback, same fix as the password gate you already paired:**
- `apps/web/app/share/[token]/page.tsx:522` (share-page breadcrumb badge)

Please grep the repo for `text-accent` next to `bg-accent-muted` and sweep all remaining hits in one pass, so we don't keep finding these file-by-file.

Once the inline items and the list above are addressed, this is good to merge.

luozejian:
Addressed all six inline comments and the review-summary sweep in `49f3c09`:

1. Restored the toast container to bottom-right so it no longer covers dashboard header controls.
2. Restored explicit `DEFAULT_FAVICON` coverage for `/logo-icon.png`; the Apple touch icon fallback remains covered separately.
3. Checked the muted-accent consumers on non-secondary surfaces and switched the affected content to `text-text-primary`. The requested repo sweep now has no `text-accent` + `bg-accent-muted` combinations.
4. `PasswordGate` now uses the resolved theme to render a single default logo source instead of fetching both variants.
5. Filed follow-up #288 to extract the duplicated themed default-logo markup into a shared component.
6. Consolidated the branding response fixtures into typed `test/branding-fixtures.ts`; both test suites now use `makeInstanceBranding`, and the `as never` casts were removed from `branding-unconfigured.test.ts`.

Also addressed every item from the review summary:
- fixed the remaining share lock, avatar, branding preview, and settings-header contrast cases;
- added `/apple-icon.png` and `/logo-full.svg` defaults to their upload previews;
- made the share breadcrumb fallback choose one logo from the resolved theme;
- merged the latest `main`.

Validation:
- Vitest: 44 files / 339 tests passed
- `tsc --noEmit`: passed
- `next lint`: passed (existing warnings only)
- production build: passed
- `git diff --check`: passed

Ready for re-review.

ravirajsinh45:
Thanks @luozejian, this is merged as of #292, and the credit for it is yours.

Everything on the last review list is addressed. The accent work in particular went further than what was asked: making `--accent-muted` opaque at the token rather than patching each chip is the better fix, and it lines up with the opaque defaults already in `globals.css`. The shared `branding-fixtures.ts` and the restored favicon assertion are good additions too. My note about `share/[token]/page.tsx:260` was a stale line number on my side, you had already fixed it.

I could not push to your fork from my end, and once #286 and #290 landed this branch went conflicting on `CHANGELOG.md`, which GitHub cannot resolve server-side. Rather than leave it waiting on another round trip, I replayed your commits onto `main` in #292 with a `Co-authored-by` trailer and folded in one review fix. Closing this one as superseded, not rejected.

The fix, for the record. The share top bar draws its fallback mark inside the circle painted with `primaryColor`, so the page theme says nothing about what the artwork has to contrast with. The ternary was reading `useResolvedTheme()`, which is the viewer's own preference, while the page's `data-theme` is forced to the share link's theme by `useShareAppearance`, and neither of those is the surface the mark sits on. On a dark accent like `#1a1a2e` the light branch drew the black mark at 1.23:1 and it vanished. There is now a `prefersLightMarkOn()` in `lib/accent.ts` that picks from the accent's own brightness, reusing the same measure and 0.6 threshold as the hover shade. It also fixes the mirror case `main` had backwards, where a light accent like `#f59e0b` got a white mark at 2.15:1.

The password gate stays keyed off the theme since it genuinely sits on `bg-bg-secondary`, but it moved to the `logo-dark`/`logo-light` pair to match the sidebar and the settings mocks.

Two smaller things I took while I was in there: the CHANGELOG line about notifications moving to the bottom-right, since `toast.tsx` ended up byte-identical to `main` after the move and the revert, and the eslint directive left above the fragment in `AppShellMock`.

Thanks again for sticking with this one through two rounds.

PR #279, opened 2026-08-23
[codex] Fix page-bound PDF annotations

nicholasjennis:
## Summary

- replace the browser PDF iframe with a page-at-a-time PDF.js canvas renderer
- scope drawing and annotation overlays to the rendered page surface
- navigate to a comment’s stored PDF page in authenticated and shared viewers
- discard stale unsent drawings when changing pages or versions
- preserve PDF page numbers through guest identity submission

## Root cause

The annotation canvases were positioned against a viewer-level surface while the browser-owned PDF iframe controlled its own multi-page layout and scrolling. As a result, annotation coordinates were static relative to the app and could overlap multiple PDF pages.

## Validation

- `python -m pytest apps/api/tests/ -v` — 346 passed, 2 skipped
- `pnpm --filter web test` — 346 passed
- `pnpm --filter web exec tsc --noEmit` — passed
- `pnpm --filter web lint` — passed (pre-existing warnings only)
- `pnpm --filter web build` — passed
- manually verified page count, navigation, drawing-layer nesting, and stale-layer cleanup on a two-page PDF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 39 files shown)


nicholasjennis:
woah sorry this should not have been submitted

nicholasjennis:
Wrong repo

PR #282, opened 2026-08-24
chore(web): remove dead tab components from review share dialog

### LINKED ISSUE #265: Share links created from the asset review view are always named after the file, with no way to name them
There are two share dialogs, and only one of them can name a link.

`components/projects/share-create-dialog.tsx` (the project grid) has a "Link name" field and posts the value:

```tsx
onCreate({ title: title.trim() || defaultTitle, ... })
```

`components/review/share-dialog.tsx` (the asset review view) has no name field, and its create body never includes `title`:

```tsx
const body: Record<string, unknown> = {
  permission,
  allow_download: allowDownload,
};
if (password.trim()) body.password = password.trim();
if (expiresAt) body.expires_at = new Date(expiresAt).toISOString();

const res = await api.post<ShareLinkResponse>(`/assets/${assetId}/share`, body);
```

With no title, the API falls back to the asset's filename (`apps/api/routers/share.py`):

```python
title=body.title if body.title else asset.name,
```

The same file then lists links with `{link.title || link.target_name}` (line 754), so the row shows the filename. It looks like a display bug, but the filename really is stored as the link's title.

### Reproduction

Confirmed against a running instance. Posting exactly what the review dialog sends, for an asset named `Gemini_Generated_Image_vtn404.png`:

```
POST /assets/{id}/share  {"permission":"comment","allow_download":true}
-> title: "Gemini_Generated_Image_vtn404.png"
```

The same endpoint keeps a title when one is sent, so the API is not at fault:

```
POST /assets/{id}/share  {"title":"Client review - round 3"}
-> title: "Client review - round 3"
```

Every creation path on the API behaves correctly here (asset, folder, project and multi-share all persist a supplied title). It is only this one dialog that never supplies one.

### Effect

Anyone who shares from the asset view gets links named `Gemini_Generated_Image_vtn404.png`, `render_v3_FINAL_2.mp4` and so on, with no way to rename them at creation. They can be renamed afterwards through the other dialog, which is not an obvious path. With several links on one asset the list is unreadable, since every row carries the same filename.

### Suggested fix

Add a "Link name" field to the review dialog and include `title` in the create body, matching the project dialog. Defaulting the field to the asset name keeps today's behaviour for anyone who does not type one, so nothing regresses for existing habits.

Worth considering separately: the two dialogs have drifted (one has naming, visibility and appearance controls, the other does not), so a link's options depend on where it was created. Sharing the configuration UI between them would stop this recurring.

rubenxyz:
Closes #265.

## What I found investigating #265

The review-view share dialog (`components/review/share-dialog.tsx`) already has a **Link name** field — it creates links through `ShareCreateDialog`, whose configure phase carries the name input (prefilled with the asset name) and posts `title` in the create body. That flow has been live since the April dropdown redesign (merged via PR #76).

The title-less create body quoted in the issue belongs to `LinkTab` — the pre-redesign tab dialog, which nothing has rendered since the redesign. The file just never shed the old components, so they read as the live dialog.

## What this PR does

Removes the dead code so the file can not mislead again:

- `LinkTab`, `DirectTab`
- their helpers `PermissionSelect` and `CopyButton`
- their response types (`ShareLinkResponse`, `ShareLinksListResponse`, `AssetSharesResponse`, `TeamsResponse`)
- unused imports (`radix-ui` dialog/tabs/select, several lucide icons, `Input`, `copyToClipboard`, now-unused types)

546 lines removed, zero behavioural change — the dropdown component is untouched apart from its imports.

## Verification

- `tsc --noEmit` clean in `apps/web`
- No references to the removed components anywhere in the repo
- The live create path is unchanged: dropdown → ShareCreateDialog → configure phase with Link name

The dialog-consolidation idea from the issue (the two dialogs have drifted on naming/visibility/appearance) is a separate piece of work — happy to pick that up if you would like it filed.


ravirajsinh45 (approved):
Good digging, @rubenxyz, confirming `LinkTab`/`DirectTab` have been dead since the April redesign and that the live path (dropdown → `ShareCreateDialog`) is untouched. Verified no remaining references. Approving.

Side note, not on you to fix here: this deletion also removes the last FE references to `GET /organizations/{id}/teams`, `GET /assets/{id}/shares`, `POST /assets/{id}/share/team`, and the `Team`/`AssetShare` types. Worth noting since team-based direct sharing has never actually had a working frontend (`DirectTab` was dead code even before the April redesign, not a regression), we'll file this as a "revive team share" feature issue rather than a restore-what-broke one, since the backend's already there if we want to build it out.

PR #283, opened 2026-08-24
feat: short share URLs — 4-char codes resolving at the origin root

rubenxyz:
Follow-up to discussion #178. This is the feature running on my instance since mid-July, now rebased onto current main and generalized (no instance-specific paths).

## Why

```
Before:  https://kamiko.xyz/freeframe/share/geQF05jn5qFR2yH--wAk0fhpVX6uuS4OYBzirOUAeuE
After:   https://kamiko.xyz/6RBr
```

91 chars with hyphens/underscores reads like a malware link in a text message; 25 chars is typeable from a phone and embeds cleanly in QR codes.

## What

- `short_code` column on `share_links` (String(10), unique, indexed) + alembic migration `7c3a9d4b1e2f` chained onto the current head
- 4-char base62 codes generated with `secrets` and checked for uniqueness at creation, on all four creation paths (asset, folder, project, multi)
- `GET /resolve/{short_code}` → 302 to the share page; unknown codes land on the app root. Sub-path deployments redirect inside their `FRONTEND_URL` path, so both root and `/freeframe`-style deployments work
- Codes surface in `ShareLinkResponse` and `ShareLinkListItem`; the dashboard (links table, link detail, link-created screen) and share emails prefer the short URL, falling back to the full token URL when no code exists
- `scripts/backfill_short_codes.py` — one-shot assignment of codes to pre-existing links
- `docs/deployment.md` — proxy configuration for root-level codes (nginx catch-all → `/resolve/`, Cloudflare rule)

## Deliberate choices

- **Codes are additive, nothing existing changes** — old token URLs keep working forever; short URLs are only generated for new links (plus backfill).
- **No user-facing rename** — the code is generated server-side and cannot be picked/collided; links keep their title for the human-facing name.
- **Emails and dashboard prefer the short form** only when a code exists, so deployments without the proxy rule configured still work via the token URL (the fallback is automatic, not a config toggle).

## Verification

- 11 new backend tests (generation, collision retries, URL preference for root + sub-path deployments, resolve endpoint including deleted-link guard); full suite: 320 passed, 40 DB-dependent errors identical to the upstream baseline
- `tsc --noEmit` clean
- Running live on my instance since 2026-07-16 (all 15 existing links backfilled, resolve proxied through nginx)

Happy to adjust anything — e.g. if you would rather the code length or the redirect target differ.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **Short share URLs** — every share link is now issued a 4-character base62 code alongside its token, so links look like `https://your-domain.example/6RBr` instead of `https://your-domain.example/share/<46-char-token>`. Codes resolve at the origin root through a new `GET /resolve/{short_code}` endpoint (302 to the share page; unknown codes land on the app root) and are shown in the dashboard, the link-created screen, and share emails whenever present, with the full token URL as the fallback. Deployments that want the short form must route root-level codes to the resolve endpoint — see `docs/deployment.md` → Short share URLs. Existing links have no code until `scripts/backfill_short_codes.py` is run once. (#178)
 - **Optional hardware-accelerated transcoding (NVENC / VAAPI) with HDR and Dolby Vision support.** Auto-detects an available GPU backend and falls back to the software pipeline when none is present, or when a hardware attempt fails at runtime (for example, out of VRAM). CPU-only and arm64 builds are unaffected by default. New env vars: `TRANSCODER_PIPELINE`, `TRANSCODER_OUTPUT`, `TRANSCODER_HDR`. (#127)
 
 ## [1.10.0] - 2026-08-21
--- a/apps/api/alembic/versions/7c3a9d4b1e2f_add_short_code_to_share_links.py
+++ b/apps/api/alembic/versions/7c3a9d4b1e2f_add_short_code_to_share_links.py
@@ -1,0 +1,35 @@
+"""add short_code column to share_links
+
+Revision ID: 7c3a9d4b1e2f
+Revises: cdcf8e5a6437
+Create Date: 2026-08-24 00:00:00.000000
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision: str = '7c3a9d4b1e2f'
+down_revision: Union[str, None] = 'cdcf8e5a6437'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        'share_links',
+        sa.Column('short_code', sa.String(10), nullable=True)
+    )
+    op.create_index(
+        op.f('ix_share_links_short_code'),
+        'share_links',
+        ['short_code'],
+        unique=True
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(op.f('ix_share_links_short_code'), table_name='share_links')
+    op.drop_column('share_links', 'short_code')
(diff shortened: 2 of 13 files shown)


ravirajsinh45 (changes requested):
@rubenxyz first, an apology: this sat for nineteen days without a review, which is not the turnaround we promise. That is on us, not on the PR.

The feature is worth having and the implementation is careful. Codes are additive, old token URLs keep working, generation is `secrets`-based with collision retries on all four creation paths, and there is a backfill script and eleven tests. The before-and-after in your description makes the case on its own: a 91-character link does read like something you would not click.

## Blocking, and it is about deployment rather than code

The short URL is preferred whenever `short_code` exists (`share.py:77`), and every new link gets a code at creation. But nothing in the app resolves a root-level code: the PR adds no web route, so resolution depends entirely on the operator adding a reverse-proxy rule by hand.

That means a self-hoster who merges this and changes nothing starts handing out dead links immediately, in the dashboard and in share emails, for every link they create. The fallback in `_short_share_url` keys on whether a code exists, not on whether the proxy is configured, and codes always exist.

The proxy rule is also not a small ask. `location / { proxy_pass .../resolve/; }` as a catch-all ordered after every app route is the kind of thing that silently shadows the frontend when someone gets the order wrong, and it is nginx-and-Cloudflare only, so Caddy, Traefik and Apache operators are on their own.

**The fix we would take: resolve it in the app.** A dynamic route at `apps/web/app/[code]/page.tsx` that looks the code up server-side and redirects. Next resolves static segments before dynamic ones, so `/share/...`, `/projects`, `/login` and the rest keep winning; only an unmatched root path reaches it. That gets you the same `https://host/6RBr` with no proxy configuration at all, on every deployment, including sub-path ones where your `FRONTEND_URL` handling already does the right thing.

Everything else in the PR stays: the column, the migration, the generator, the backfill, the dashboard and email preference. `/resolve/{short_code}` is still worth keeping as the API-side lookup the route calls.

## Two smaller things, neither blocking

**A 4-char code shares a namespace with future root routes.** Nothing today is four characters, but `/docs` or `/help` later would shadow a code silently. Worth either reserving a one-character prefix (`/s/6RBr`, still 24 characters) or keeping a reserved-word list next to the generator so a collision fails at creation rather than at click time.

**62^4 is 14.7 million.** Fine for one instance, and the retry loop handles collisions, but worth knowing the birthday bound means a 1% collision chance at roughly 17,000 links. `MAX_SHORT_CODE_RETRIES` covers it; a note in the generator saying where five characters becomes the answer would save someone the arithmetic later.

rubenxyz:
Thanks for the review — and no apology needed. You were right that the fallback keyed on the code existing rather than the proxy being configured, which made a merge-and-change-nothing deployment hand out dead links. Rebuilt it the way you suggested:

**Resolved in the app now**

- `apps/web/app/[code]/page.tsx` — root-level dynamic route. It validates the 4-char shape, calls `/resolve/{short_code}` server-side, and redirects; unknown codes land on the app root, same as the old catch-all. Static segments resolve first, so `/share/...`, `/projects`, `/login` and the rest are untouched. `/resolve/{short_code}` stays as the API-side lookup the route calls.
- `middleware.ts` treats 4-char root paths as public, so anonymous visitors are not bounced to `/login`.

**The two smaller things**

- **Namespace:** took the reserved-word option (keeps the 25-char link). The generator skips `RESERVED_CODES` — the names of the web app's root routes plus plausible future ones — and a backend test scans `apps/web/app/` and fails when a new 4-char root route is not reserved, so a future `/docs` fails at creation rather than at click time.
- **62^4:** the generator docstring now states the birthday bound (1% at ~17k links) and that 5 chars (916M) is the answer beyond it. The retry loop already covers collisions.

**Sub-path handling**

Since the code now resolves at the deployment root, a sub-path install gets `https://host/freeframe/AbC1` and needs no proxy rule either. `_short_share_url` and the dashboard/link-created URL builders use the basePath-aware deployment root rather than the origin.

**Server-side lookup**

`API_INTERNAL_URL` when set (Docker Compose sets `http://api:8000`), otherwise `NEXT_PUBLIC_API_URL` resolved against the request origin. `docs/deployment.md` describes the behavior instead of proxy recipes.

**Verification**

322 backend tests pass (40 DB-dependent errors, identical to the upstream baseline), 343 frontend tests pass, `tsc --noEmit` clean, `next build` emits `ƒ /[code]`. Smoke-tested the built standalone server against a live API: `/AyOd` → 302 to `/share/<token>`, unknown code → app root, `/login` 200. Pushed as `1339a98`.

PR #286, opened 2026-08-25
fix(web): stop removing React-managed icon links

Thedude7054:
## Summary

Take 2 with this PR- the previous fork had to be removed and recreated

Next.js auto-generates React-managed <link rel="icon"> and <link rel="apple-touch-icon"> tags from the app/icon.png, app/favicon.ico, and app/apple-icon.png files. BrandingHead removed those links with el.remove(), detaching nodes React still owned — on the next navigation React tried to unmount one and crashed with:

> TypeError: null is not an object (evaluating 'instance.parentNode.removeChild')

This removes those app/ metadata files (favicon.ico moves to public/) so Next stops emitting the React-owned links, and BrandingHead is the sole owner of the icon links again.

## Changes

- Move favicon.ico to public/ so /favicon.ico still resolves by convention, and delete icon.png (byte-identical to public/logo-icon.png) and apple-icon.png. BrandingHead now owns the icon links exclusively and only ever removes non-React links.

## Testing

## Checklist

## Screenshots

None


apps/web/components/shared/branding-head.tsx
```diff
@@ -14,11 +14,15 @@ const DEFAULT_ICON = '/logo-icon.png'
...
- * nothing. Any icon link this component doesn't own is removed first.
+ * Next emits its own `<link rel="icon">` for any icon file under `app/`
+ * (icon.png, favicon.ico, apple-icon.png). Those are React-managed hoistable
+ * resources, so removing them with `el.remove()` left React holding a reference
+ * to a detached node and it would throw `null is not an object (evaluating
+ * 'instance.parentNode.removeChild')` the next time it committed. Those files
+ * have therefore been moved out of `app/` (favicon.ico lives in `public/` so
+ * browsers can still fetch /favicon.ico by convention), and this component owns
+ * the icon links exclusively. The removal sweep below is defensive only — it
+ * clears any stale non-branding links that aren't React-managed.
```
ravirajsinh45: This comment says the sweep below is "defensive only" and only clears links "that aren't React-managed," but the code doesn't actually check for that, it only excludes `[data-ff-branding]`. Any other future icon link, React-managed or not, still gets `el.remove()`'d. Either tighten the comment to say what the code actually does, or guard `setIcon()` against React-managed nodes so this can't regress. Non-blocking for this PR since there's no file under `app/` left to trigger it right now, but let's not ship a comment that overstates the safety here.

CHANGELOG.md
```diff
@@ -7,9 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+
```
ravirajsinh45, line 12: Double blank line here, and again at 16-17 before `## [1.10.0]`. Single blank line like the rest of the file.

ravirajsinh45 (changes requested):
Thanks for this, @Thedude7054, and for tracking down the actual mechanism (React loses the node it owns when we `el.remove()` it out from under it). One blocking issue before we merge: `apple-icon.png` needs the same treatment `favicon.ico` got here, not a straight delete. A couple of non-blocking notes below too.

**Blocking:**

**`apps/web/app/apple-icon.png` needs to move, not delete.** `icon.png` is fine to delete outright, it's byte-identical to `public/logo-icon.png`, which `DEFAULT_FAVICON` already falls back to. But `apple-icon.png` isn't a duplicate of anything else in the repo, and `branding-head.tsx`'s `DEFAULT_APPLE_ICON` already points at `/apple-icon.png` (added in #277), expecting Next's `app/` file-convention route to serve it. Deleting the file outright breaks that path (404), and there's no `public/apple-icon.png` to fall back to instead. Move it to `public/apple-icon.png` the same way you moved `favicon.ico` here, that keeps the URL working and keeps the dedicated 180x180 asset instead of silently falling back to the 1311x1311 `logo-icon.png`.

**Non-blocking:**

`apps/web/components/__tests__/branding-head.test.tsx`: the existing "removes the framework's own icon link" test fabricates the competing link with `document.createElement('link')`, which isn't a React-managed/hoisted node the way the real crash-triggering one was. It won't catch a regression of this exact bug. Worth a follow-up test that goes through Next's actual metadata/icon mechanism if there's a reasonable way to simulate it.

Once `apple-icon.png` is handled, this is good to merge.

Thedude7054:
All notes taken into account, good shout on ```DEFAULT_APPLE_ICON```. Let me know if their are any other changes that need to be made @ravirajsinh45

PR #309, opened 2026-08-30
refactor(web): extract themed default logo

### LINKED ISSUE #288: refactor(web): extract shared themed default logo
Follow-up from the review of #277. The paired `logo-dark` / `logo-light` fallback markup is duplicated across branding and shell surfaces, which makes it easy for variants, alt text, sizing, or loading behavior to drift.

Known consumers to consolidate include:
- `components/layout/sidebar.tsx`
- `components/settings/app-shell-mock.tsx`
- `components/auth/auth-branding-header.tsx`
- `components/shared/powered-by-badge.tsx`
- other fallback-logo consumers found during implementation

Acceptance criteria:
- Add a shared component for themed default icon/full-logo fallbacks.
- Support meaningful and decorative alt text plus caller sizing classes.
- Replace duplicated paired JSX blocks without changing light/dark appearance.
- Add focused tests for both themes and avoid loading an unnecessary hidden asset where the resolved theme is already available.

luozejian:
## Summary
- add a shared `ThemedDefaultLogo` with icon/full variants, meaningful or decorative accessibility modes, and caller sizing classes
- render one canonical asset per logo and theme it through `data-theme` CSS, preserving share-page overrides without loading a hidden second image
- migrate all six duplicated fallback blocks across sidebar, auth, share password gate, powered-by badge, and branding mocks
- use the themed fallback when a custom auth logo fails, while allowing a new logo URL to retry

## Testing
- `pnpm test` — 47 files, 358 tests passed
- `pnpm exec tsc --noEmit --pretty false`
- `pnpm build`

Closes #288

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Two consequences worth knowing. Guests who had identified themselves on a single-asset link will be asked once more, because the two paths stored that identity under different keys and the shared one wins. And the single-asset page no longer shows the instance's logo and share name above the player, since the folder path's asset view never did; making both show branding is tracked separately.
 
+### Fixed
+- **Default logos no longer load an invisible second theme asset** — sidebar, authentication, share-password, attribution and branding-preview fallbacks now share one themed icon/full-logo component. A failed custom authentication logo also falls back to the correct built-in mark and retries when branding supplies a new URL. (#288 by @luozejian)
+
 ## [1.12.0] - 2026-08-29
 
 ### Upgrade notes
--- a/apps/web/app/globals.css
+++ b/apps/web/app/globals.css
@@ -113,10 +113,15 @@
   }
 }
 
+/*
+ * The logo owns its filter. Keeping these rules outside Tailwind's layers and
+ * scoping both themes through data-theme makes them outrank normal utility
+ * classes without resorting to !important.
+ */
+[data-theme="light"] .themed-default-logo { filter: none; }
+[data-theme="dark"] .themed-default-logo { filter: brightness(0) invert(1); }
+
 @layer components {
-  /* The shared logo loads one dark-ink asset; dark surfaces recolor it white. */
-  .themed-default-logo { filter: none; }
-  [data-theme="dark"] .themed-default-logo { filter: brightness(0) invert(1); }
 
   /* Smooth focus ring for interactive elements */
   .focus-ring {
--- a/apps/web/app/globals.css
+++ b/apps/web/app/globals.css
@@ -114,11 +114,9 @@
 }
 
 @layer components {
-  /* Theme-aware logo switching */
-  [data-theme="dark"] .logo-dark { display: block; }
-  [data-theme="dark"] .logo-light { display: none; }
-  [data-theme="light"] .logo-dark { display: none; }
-  [data-theme="light"] .logo-light { display: block; }
+  /* The shared logo loads one dark-ink asset; dark surfaces recolor it white. */
+  .themed-default-logo { filter: none; }
+  [data-theme="dark"] .themed-default-logo { filter: brightness(0) invert(1); }
 
   /* Smooth focus ring for interactive elements */
   .focus-ring {
--- a/apps/web/app/globals.css
+++ b/apps/web/app/globals.css
@@ -114,12 +114,20 @@
 }
 
 /*
- * The logo owns its filter. Keeping these rules outside Tailwind's layers and
- * scoping both themes through data-theme makes them outrank normal utility
- * classes without resorting to !important.
+ * The logo owns its filter. Tailwind v3 flattens the @tailwind directives at build
+ * time and emits no @layer at-rules, so nothing here gains layer priority. What keeps
+ * these ahead of a utility is specificity — two selectors against a utility's one —
+ * plus sitting after the utilities @tailwind emits. That covers bare utilities only:
+ * a variant-prefixed one (hover:, group-hover:, peer-*) is emitted later at equal or
+ * higher specificity and would still win, which is why callers pass sizing and layout
+ * classes only.
+ *
+ * Dark is the unscoped default so it matches :root at the top of this file: a document
+ * that never received a data-theme is painted with the dark palette, and the mark has
+ * to be legible against it.
  */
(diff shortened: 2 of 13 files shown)


ravirajsinh45 (changes requested):
@luozejian thanks for pulling these six duplicated logo blocks into `ThemedDefaultLogo`, good direction. A few things to fix before merge:

**Blocking**

1. **CHANGELOG** — no entry under `[Unreleased]`. This fixes a real bug (auth logo fallback never retrying) and changes rendered output in 6 places, so it should be documented per AGENTS.md.
2. **Screenshots** — the PR touches visible logo rendering in the sidebar, auth header, share password gate, powered-by badge, and settings previews, but there's no before/after screenshot. Please add them.
3. **Theme filter can be silently overridden** — `.themed-default-logo`'s `filter` lives in `@layer components` (`globals.css:118`), and `ThemedDefaultLogo` merges an open `className` and forwards `style` with no guard (`themed-default-logo.tsx:37`). Any future caller passing a filter-related class or inline style (e.g. `className="drop-shadow-sm"`, `style={{filter: 'none'}}`) will silently break the dark-theme recolor with no error. Worth excluding `style` from the forwarded props and either scoping the filter more defensively or documenting the constraint clearly on the component.

**Worth a look, not blocking**

- `auth-branding-header.tsx`'s new `failedLogo` state never retries a URL that recovers after a transient failure (`branding-preview.tsx` already solves this the same way with a `useEffect` reset on URL change — could reuse that pattern).
- The theme-based logo-URL fallback ternary (`loginLogoUrl || (theme === 'dark' ? ... : ...)`) is now duplicated across `auth-branding-header.tsx`, `share/[token]/page.tsx`, and `sidebar.tsx` — since this PR already touches all three, might be worth extracting alongside the component work.
- `themed-default-logo.test.tsx`: the injected `<style>` tag (line 47) only gets cleaned up if its assertions pass, so a failing test leaks it into later tests in the file.
- `FreeFrameMark` in `powered-by-badge.tsx` is now a dead one-line wrapper — could be inlined at its single call site.

luozejian:
@ravirajsinh45 
Addressed the review feedback in a follow-up change:

- added an `[Unreleased]` / `Fixed` CHANGELOG entry
- made the logo filter component-owned: `style` and `srcSet` are excluded, caller `className` is documented as sizing/layout only, and explicit `data-theme` rules now sit outside Tailwind layers
- aligned auth image-error reset with the branding preview and added an A → B → A retry regression test
- extracted and tested the shared branding-logo URL resolver across sidebar, auth, share password gate, and branding preview
- made the CSS test cleanup unconditional with `finally`
- inlined the one-use `FreeFrameMark` wrapper

Validation:
- focused tests: 3 files / 15 tests
- full Vitest: 48 files / 365 tests
- TypeScript: passed
- lint: passed (existing warnings only; none in changed files)
- production build: passed

Screenshots are attached below. The intended visual output is unchanged; the comparisons cover the login/auth header and attribution, the share password gate, and the settings branding/sidebar preview.

ravirajsinh45 (approved):
Verified the pixel claim independently and picked up the remaining small items in 98b6276d. CI green on all four jobs. Thanks @luozejian.

ravirajsinh45:
@luozejian thanks for the quick turnaround on round 1. Five of the seven items landed cleanly, and rather than send this back for another round we picked up the remaining small ones ourselves and pushed 98b6276d to your branch.

One thing worth flagging first: your comment says "Screenshots are attached below" but nothing came through, so the upload must have failed silently. With no visual evidence to check the change against, we ended up verifying the pixel claim the long way, and that closer look is where the rest of this came from. Worth attaching them next time a PR touches rendered output, since AGENTS.md and the PR template both ask for it.

The good news is that the swap really is pixel-exact, and we confirmed it rather than assuming it. `logo-icon.png` is RGB(255,255,255) at every pixel with alpha above zero, `logo-icon-dark.png` is RGB(0,0,0), and their alpha channels are byte-identical, so `brightness(0) invert(1)` reproduces the white asset exactly. For the lockup, `sed 's/#18181b/white/g' logo-full-dark.svg` diffs clean against `logo-full.svg`, and neither file has gradients, strokes or opacity. Every migrated call site keeps its exact class list, and the old `.logo-dark` = dark / `.logo-light` = light direction is reproduced correctly at each one. The share password gate also quietly gained an accessible name in light theme that it never had before, which the PR does not claim but is a real improvement.

What changed in 98b6276d:

**1. The globals.css comment named the wrong mechanism.** "Outside Tailwind's layers" does not do anything on Tailwind 3.4: the `@tailwind` directives are flattened at build time and the output carries no `@layer` at-rules. What beats a bare utility is specificity, two selectors against one, plus sitting after the emitted utilities. We compiled the real config to check, and that covers bare utilities only. `.hover\:drop-shadow-sm:hover` (equal specificity, emitted later) and `.group:hover .group-hover\:brightness-110` (higher) still win. The comment says that now, which makes the caller contract legible instead of implied.

**2. Dark is now the unscoped default.** Both rules were attribute-gated, but `:root` and `[data-theme="dark"]` share the dark palette at the top of the same file, so a document that never received a `data-theme` painted dark while the logo kept its black ink. Narrow, since the head script sets the attribute on both its try and its catch path, but one selector closes it and it restores the file's own convention.

**3. `style` and `srcSet` are now actually dropped.** They were omitted from the props type but still reached the `` at runtime, because the spread precedes only `src`, `alt`, `aria-hidden` and `className`. Pinning both to `undefined` makes the JSDoc's ownership claim true without resting on the caller being typechecked.

**4. Three test fixes.** The prop-restriction test asserted a literal `[true, true]` equals `[true, true]`, which cannot fail under vitest since types are stripped; it is now a render that forces both props past the types and checks the DOM. The `globals.css` read was anchored to `process.cwd()`, which is the launching shell's directory rather than vitest's root, so it threw ENOENT if vitest ran from the repo root. And `applies the production theme filter despite a later filter class` over-claimed: `.filter-none` is one selector and loses on specificity whatever its position, so the assertions came out identical with the class present, absent, or first in the sheet. It is renamed to what it checks, and the theme scoping is now asserted explicitly so an unscoped rule cannot slip back in.

Each of those is mutation-checked: revert the runtime strip, delete a CSS rule, or unscope one, and the suite fails.

`tsc --noEmit` clean, 48 files / 365 tests passing, no new lint warnings, and all four CI jobs green, so this is merged as 5f9b0714. Thanks for taking this one on.

PR #314, opened 2026-09-04
perf(transcoder): upload HLS output several files at a time

Lennart-Pingpong:
## Summary

The segment upload is the longest phase of a transcode that has nothing to do
with encoding, and it runs one file at a time.

Each PUT to a remote object store is dominated by the round-trip rather than by
the segment. At the default segment length a 50-minute encode produces on the
order of 850 files, so the worker spends that whole stretch idle, waiting on
latency in series.

Measured against a real master (~3.9 GB source, HLS output) over a remote store:
**197s serially, 65s with eight threads.** The saving grows with exactly the two
things that make the problem hurt — a longer encode makes more segments, and a
more distant store makes each one cost more.

This is the same shape of change as #250 on the browser side, one layer down.

## Changes

- The `for f in hls_dir.rglob("*")` loop moves into `_upload_directory()` and
  runs on a `ThreadPoolExecutor`.
- **Eight threads, deliberately not configurable.** Above eight the gain
  flattens, and botocore's default connection pool holds ten — so the workers
  never queue for a socket, and the two thumbnail PUTs that follow still have
  room. A knob that could be raised past the pool would trade waiting on the
  network for waiting on the pool, which looks identical from the outside and is
  worse. Raising both together is a change worth making deliberately rather than
  by env var; happy to add it that way if you would rather have it.
- The pool is never wider than the work, so a two-file output does not create
  eight threads to do nothing.
- **The client is shared.** boto3 clients are thread-safe, and HLS segments sit
  far below the multipart threshold, so each `upload_file` is a single PUT that
  brings no threads of its own.
- The file list is sorted, so the returned keys no longer depend on directory
  iteration order.

## One behaviour change beyond speed

`ThreadPoolExecutor.map` is lazy. The results are drained inside the pool, so a
failed upload raises where the serial version raised. Without that it would sit
unread in its future, the transcode would report success, and the version would
go `ready` with segments missing from the bucket — a silent truncation of the
same family as the client-supplied parts list you removed in #255. There is a
test for exactly that.

## Something noticed while doing it, not changed

`uploaded_keys` is built and never read — nothing in the file or elsewhere
consumes the return value. It is still built here, because `map` returns it
anyway and removing it is a separate question. Say the word and it can go.

## Running in production since 7 August

We have been running this as a local patch on our own FreeFrame instance for
four weeks, over v1.9.0. Eleven transcodes have gone through it in that period,
and the worker log shows no failure anywhere in the upload path.

Upload phases from that log, as they happened:

```
568 files in 148.9s
435 files in  99.0s
697 files in  38.8s
```

The 197s/65s figure above is a controlled before-and-after on one master rather
than an average of those.

Being precise about what has that mileage and what does not. The mechanism is
unchanged from what runs in production: eight threads, one shared client,
`list(pool.map(...))` to drain the results. What is new here is the packaging —
the loop extracted into `_upload_directory`, the file list sorted, the pool
capped at the number of files, the log line, and the whole thing rebased onto
current `main` rather than the v1.9.0 file it was written against. Those parts
are covered by the tests, not by the four weeks.

## Testing

- [x] Backend tests pass — 409 passed, including 6 new ones in
      `apps/api/tests/test_transcoder_parallel_upload.py`.
- [x] Frontend tests + build pass — untouched by this change.
- [x] Tested manually in browser — four weeks of real uploads through the
      normal UI on our instance, per the section above, rather than a single
      click-through. The effect is server-side, so what a browser confirms is
      that the resulting HLS plays; that it does is the eleven transcodes.

The concurrency test observes how many uploads are in flight at once rather than
inferring it from the result, since every file lands either way. Reducing the
pool to a single worker makes it report a peak of 1 and fail, which is the check
that it is testing the property and not the plumbing.

## Checklist


ravirajsinh45 (approved):
Thanks @Lennart-Pingpong. The mechanism is right and the reasoning about where the time goes matches what we see. We checked the thing we most wanted to check: `list()` really does drain inside the `with`, so a failed segment upload raises where the serial version raised rather than sitting unread in a future. The `is_file()` guard survived the extraction too, so directories are still not handed to `upload_file`.

Approving. Four things we would like tidied first, none structural.

**The file count is off by about 5x.** The CHANGELOG and the docstring both say a 50 minute encode produces "on the order of 850 files". With `-hls_time 2` across the three rung ladder it is closer to 4,500. The argument gets stronger rather than weaker, so it is worth correcting rather than dropping.

**The thread count justification does not quite hold.** Two parts. The docstring says "the two thumbnail PUTs that follow" and there is one, and it runs after the pool has closed rather than alongside it. More substantially, `upload_file` builds its own `TransferManager` per call, so "eight workers against a default pool of ten" is not the headroom guarantee it reads as. Eight is still a fine choice. We would rather the comment said it was measured than presented arithmetic that does not survive a look at `upload_file`.

**`test_the_pool_is_never_wider_than_the_work` passes with the cap removed.** Deleting `min(len(files), _UPLOAD_THREADS)` leaves it green, so it asserts the outcome rather than the property. The concurrency test next to it does this correctly and is a good model for it.

**`_UPLOAD_THREADS` as a hardcoded constant.** We take the argument in the description and we are not asking for the env var. Given every other knob in this module comes from the environment, one line saying it is deliberately fixed and why would save the next person re-litigating it.

On `uploaded_keys`: leave it. Removing it is a separate question and this PR is better for not answering it.

Issue #316, opened 2026-09-04
Comment export on the project page offers only CSV, and clicking it does nothing


Lennart-Pingpong:
### Describe the bug

The comment panel is rendered in two places, and its export control only works
in one of them.

On the project page, with an asset selected and its comments showing in the
right-hand panel, the export menu offers **only CSV** — the DaVinci Resolve,
Final Cut Pro and Premiere entries are missing — and clicking CSV does nothing
at all. No download, no error, nothing in the console.

Opening the same asset in the review screen and using the same control offers
all four formats, and the download works.

The cause is the same for both symptoms. `CommentPanel` reads `currentAsset` and
`currentVersion` from `useReviewStore`, and that store is only populated by
`review-provider.tsx`, which is mounted on the review screen. The project page
renders `CommentPanel` directly (`app/(dashboard)/projects/[id]/page.tsx:966`),
without that provider and without passing an `assetId`, so `currentAsset` is
`null` there. Two consequences follow:

- the three NLE entries are gated on `currentAsset?.asset_type === "video"`,
  which is false, so only CSV is rendered
- `handleExport` opens with `if (!currentAsset || !versionId) return;`, so every
  format returns silently — including the CSV that is still on screen

The silent return is the part worth fixing first, whatever is decided about the
rest: the panel offers an action it knows it cannot perform, and then declines
to say so.

### Steps to reproduce

1. Open a project that has a video asset with timecoded comments.
2. In the asset grid, click the asset once so it is selected — do not open it.
3. In the Comments panel on the right, click the download icon ("Export comments").
4. The menu lists only CSV.
5. Click CSV. Nothing happens.
6. Now open the same asset (click through to the review screen) and use the same control: all four formats are listed, and the download works.

### Expected behavior

Either of these would be consistent; the second is the smaller change:

- The export works on the project page too, using the asset that is already
  selected there. `selectedAsset` is in scope at the call site, so the panel
  could take the asset and version as props and fall back to the store.
- Or the export control is not rendered where it cannot work, so nothing offers
  an action that silently does nothing.

What should not happen is the current combination: a menu that quietly drops
three of its four entries because a store is empty, and a fourth entry that
returns without a word.

### Screenshots

<img width="534" height="482" alt="Image" src="https://github.com/user-attachments/assets/61f6d042-6ca0-43c3-a94c-b162c16d774b" />
side panel in project page

### Area

Frontend (Next.js)

### Version / Commit
[...]

PR #312, opened 2026-09-04
feat(upload): resume an interrupted upload instead of restarting it

Lennart-Pingpong:
## Summary

Closes #241, in the order and the shape we settled on in the issue thread.

An upload that stops part-way (network drop, closed tab, reload) now offers **Resume**, which sends only the parts the storage backend is not already holding, and **Discard**, which throws the upload away on purpose. Before this, the client aborted the multipart upload on *any* error, so there was nothing left to resume from.

The branch has grown through two review rounds and two rounds of testing on a dev instance. What changed since the first push is summarised first; the design reasoning from the first push follows and still holds.

## Since the first review

**Round 1 (08.09.)**, your five points plus four things found testing by hand:

- In-flight rows are persisted and rewritten when a tab loads them. Without that, reloading during the upload of a second version left nothing to resume.
- `POST /upload/abort` takes `discard`. Discard disposes of the upload the way the stale-upload reaper does instead of marking it failed, or publishing an assembled one.
- The version switcher refetches after a discard (`versionsRevision`).
- The panel row names the file being uploaded and its project.
- Version numbering past soft-deleted versions was found here and shipped separately as #334.

**Round 2 (15.09.)**, rebased onto `main` without the numbering commit, and:

- **A second tab or a second device** no longer offers Discard or Resume on a transfer that is still running. The sending tab stamps its rows, another tab shows them as `Elsewhere` with nothing to press, and `/parts` and asset responses report `last_activity_at`. A request that finds the upload recently active is not itself recorded as activity, so a refusal cannot renew itself.
- **Discard** confirms first, keeps its row with the reason when it fails, re-reads the version under a row lock before deciding, and removes the asset only for an editor on the project.
- **Cancel** is sent as a discard and refreshes the version switcher.
- **The trash** leaves out assets that never had a version reach processing, and restore refuses them.
- Discard tests run against Postgres and check what was committed. `pollIntervalFor` is a pure function with its own test.
- CHANGELOG attribution and upgrade notes.

For a reader skimming: **one migration** (`asset_versions.chunk_size_bytes`), **one new endpoint** (`GET /upload/{version_id}/parts`), **one new optional request field** (`discard` on `POST /upload/abort`), **one new response field** (`last_activity_at` on versions), and **the trash listing now filters**.

## Changes

[...]

PR #317, opened 2026-09-04
fix(comments): refuse an empty marker export instead of downloading nothing

Lennart-Pingpong:
## Summary

Exporting a version's comments to an NLE produced a valid file with no events
when none of the comments carried a timecode:

```
TITLE: 255 ClausHartmann 1.2
FCM: DROP FRAME
```

The download succeeds, DaVinci Resolve imports it and reports nothing, and the
cause surfaces three applications away from where it is — looking like a broken
export rather than an empty one. Finding it took opening the file in a text
editor.

How it happened here, on a real instance: a reviewer commented through a
single-asset share link, which on that release rendered a plain textarea with no
timecode control, so all seven comments arrived with `timecode_start` NULL and
the times typed into the body (`"30:20 audio cuts out"`). #307 has since fixed
that side of it. This is the other half — that the export said nothing about it.

`build_markers` is right to skip comments with no timecode. The mistake is
serialising the empty result and calling it a download.

## Changes

- An NLE export that would contain no markers answers **422** with a coded
  detail, matching the `fps_required` path already in this handler.

- The message distinguishes the three situations that produce the same empty
  file, because they need different answers:

  - the version has no comments at all
  - it has comments, but none are attached to a timecode
  - the timecoded ones are all resolved, and this export excluded them

  The middle one is the case above and the one worth wording carefully: the
  comments are visibly on screen, so "no comments to export" would read as a
  lie. It names what a timecode is — a time typed into the text is not one —
  and points at CSV.

- **CSV is deliberately not blocked.** It carries every comment, timecoded or
  not, and it is what the message recommends; refusing it there would send the
  user in a circle. There is a test for that.

- In the client, every export error other than the fps prompt went to
  `console.error`. An unsupported frame rate, a non-video asset and now this all
  produced a click that did nothing visible. The reason is shown in the comment
  panel, beside the control that failed and alongside the fps prompt, which is
  where this component already answers for that button.

- `exportComments` read only the string form of `detail`, so every structured
  error collapsed to `"Export failed"`. It now reads `message`.

## Why the panel and not a toast

There is a `ToastProvider` in the root layout, and a toast was the first thing I
wrote. It made every existing test that renders `CommentPanel` fail, because
`useToast` throws without a provider and those tests have no reason to have one
— 44 tests across three files, plus the same trap for anyone adding a test
later. The panel already renders its own answer for this button (the fps
prompt), so the message goes next to it. Happy to switch it if you would rather
the app spoke with one voice here.

## Testing

- [x] Backend tests pass — 407 passed, 5 new in
      `apps/api/tests/test_comment_export_empty.py`, including the one that
      keeps CSV working when nothing is timecoded.
- [x] Frontend tests + build pass — 372 passed, 7 new across
      `lib/__tests__/export-comments.test.ts` and
      `components/review/__tests__/comment-panel-export-error.test.tsx`,
      `tsc --noEmit` clean.
- [x] Tested manually in browser — this started as a real export on our own
      instance, which is where the empty file came from.

One of the new tests caught a bug while being written: an `Error` with an empty
message left the panel rendering nothing, because the line is shown on
truthiness.

## Checklist

- [x] `CHANGELOG.md` entry added under `## [Unreleased]` — two `Fixed` entries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Exporting comments to an NLE no longer produces a silently empty marker file** — a version whose comments carry no timecode exported to an EDL with a header and nothing after it. The download succeeded, the file opened, and DaVinci Resolve reported nothing to import, so the cause surfaced three applications away from where it was and looked like a broken export rather than an empty one. The export now refuses, and says which of the three reasons applies: the version has no comments, none of them are attached to a timecode, or the only timecoded ones were resolved and excluded. CSV is unaffected and still carries every comment, which is what the message points at. (#84)
+- **A failed comment export now says so** — every error except the frame-rate prompt went to `console.error`, so an unsupported frame rate, a non-video asset and a version with nothing timecoded all produced a click that did nothing visible. The reason is shown in the comment panel, next to the control that failed. Structured error details from the API are read for their message rather than collapsing to "Export failed".
+
 ### Changed
 - **A share link to a single asset now opens the same review screen as one inside a folder** — the two paths rendered entirely different component trees, so sharing an asset on its own gave a plain video element and a bare comment box, while sharing the same asset inside a folder gave the real review stack. The single-asset path now renders that stack too, which brings it timecode-attached comments, annotation and mention controls, and a version switcher with version-scoped streams and comments. (#117, #123)
 
--- a/apps/api/routers/comments.py
+++ b/apps/api/routers/comments.py
@@ -703,6 +703,31 @@ def comment_deep_link(
     return {"url": url}
 
 
+def _no_markers_reason(rows: list, include_resolved: bool) -> str:
+    """Why a marker export has nothing to place on a timeline.
+
+    Three different situations produce the same empty file and need different
+    answers from whoever is looking at it -- most of all the middle one, where
+    the comments are right there on screen and only the timecode is missing.
+    """
+    tops = [r for r in rows if r.parent_id is None]
+    if not tops:
+        return "This version has no comments to export."
+    if not any(r.timecode_start is not None for r in tops):
+        return (
+            "None of this version's comments are attached to a timecode, so there is "
+            "nothing to place on a timeline. A comment carries a timecode when it is "
+            "written with the clock attached in the review screen; a time typed into "
+            "the text is not one. Export as CSV to get the comment text instead."
+        )
+    if not include_resolved:
+        return (
+            "Every timecoded comment in this version is resolved, and this export "
+            "excludes resolved comments."
+        )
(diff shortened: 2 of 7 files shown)


ravirajsinh45 (changes requested):
Thanks @Lennart-Pingpong. Refusing the export rather than serving an empty file is right, and the three way message is what makes it useful rather than merely correct. Keeping CSV open is right too, and we are glad there is a test pinning it.

One thing to fix, then this is good to go.

**`exportError` is never cleared except by another export attempt.** `setExportError(null)` at the top of `handleExport` is the only reset. Export v2 as EDL, get the refusal, switch to v1 with the version switcher, and roughly six lines of red text about v2 are still sitting above v1's fully timecoded comments. `VersionSwitcher` and `CommentPanel` are siblings in one mounted tree with no `key`, so nothing remounts. Compare mode has the same shape on both panes. Clearing on `currentAsset?.id`, `currentVersion?.id` and `exportVersionId`, and on opening the menu, covers it.

**The CHANGELOG has two `### Fixed` sections** under `[Unreleased]` now. Worth folding into one before this lands, since #312 and #314 are queued behind it and will conflict there anyway.

Two notes, neither blocking.

**The CHANGELOG line reads wider than the change is.** "A failed comment export now says so" is true on the review screen. On the project page it is not: `handleExport`'s `if (!currentAsset || !versionId) return` sits before the try/catch that is the only writer of `exportError`, so that click is still silent. That is #316 rather than this PR and we are not asking you to fix it here, but the sentence currently promises it. The same line names "a non-video asset" as a fixed silent failure, and the menu does not render the NLE entries for a non-video asset, so that path is not reachable from the UI either.

**The export control is offered to share link guests.** `share-review-screen.tsx:227` renders `CommentPanel`, and `exportComments` sends only `Bearer ${getAccessToken()}`, which is `null` for a guest. Before this PR the 401 went to `console.error` and nobody saw it. After it, a guest gets a bare "Export failed (401)" in the panel. That is pre-existing and this PR only makes it visible, so it is not yours to fix here. We will file it. Worth mentioning because it is the same persona the description opens with.

ravirajsinh45:
Closing this in favour of #326, and not because anything here was wrong.

#326 is built on this branch and carries this commit (`2075884`), so when it merges the empty-marker refusal lands with it. Keeping both open would mean asking you to fix the same review comment twice on two branches, which is the only thing that would be lost by leaving this open.

The one blocking item we raised here is still outstanding, so we have restated it on #326 rather than leaving it stranded: `exportError` is only ever cleared by another export attempt, since every `setExportError` call sits inside `handleExport` (`comment-panel.tsx:819`, `:928`, `:933`, `:956`). Clearing it on `currentAsset?.id`, `currentVersion?.id` and `exportVersionId`, and on opening the menu, covers it. The duplicate `### Fixed` sections under `[Unreleased]` are on #326 too.

Everything to do is on #326 now. Thanks @Lennart-Pingpong.

PR #322, opened 2026-09-04
fix(api): store the name an invitee types on the accept screen

### LINKED ISSUE #320: Accept-invite asks for a full name and the API discards it
### Describe the bug

The accept invite screen makes "Full name" a required field, validates it, and sends it. The API discards it, so the name an invited person types is never stored and their display name stays whatever the inviter's dialog derived from their email address.

`apps/web/components/auth/invite-accept.tsx`:

- `:32` — `if (!name.trim()) errors.name = 'Name is required'`
- `:152` — renders the field, labelled "Full name"
- `:85` — includes `name` in the POST body

`apps/api/schemas/auth.py:97-99`:

```python
class AcceptInviteRequest(BaseModel):
    token: str
    password: NewPassword
```

There is no `name`, so Pydantic's default `extra="ignore"` drops it, and `accept_invite` (`apps/api/routers/auth.py:157-183`) sets `password_hash`, `email_verified`, `status`, `invite_token` and `invite_token_expires_at`, but never `user.name`.

The name that survives is the one `POST /users/invite` was given. The Bulk Invite dialog does not ask for one and fills it in as `email.split("@")[0]`, so someone invited as `patrick.rosen@example.com` is displayed as `patrick.rosen` even though they typed "Patrick Rosen" on the way in. They can correct it later in Settings, but they have already been asked once and told nothing was wrong.

### Steps to reproduce

1. As a superadmin, invite an address through Bulk Invite.
2. Open the invite link, fill in "Full name" with a real name, set a password, accept.
3. Sign in and look at the profile, the members list, or a comment on an asset.
4. The display name is the local part of the email address, not what was typed.

### Expected behavior

The name entered on the accept screen is stored on the user.

### Suggested fix

Add `name` to `AcceptInviteRequest` with `min_length=1` and `max_length=255`, matching `User.name` which is `String(255)` and `nullable=False`, and assign `user.name = body.name.strip()` in `accept_invite`. A test that accepts an invite with a name and asserts the stored value would cover it.

### Notes

Found while triaging #318, which is about the derived name reaching NLE exports. This is the root cause for every invited user and is independent of what the invite dialog does, so it seemed worth its own issue. Credit to @Lennart-Pingpong, whose #318 is what surfaced it.

### Area

Backend (FastAPI), Frontend (Next.js)

shaurya703:
Fixes #320

Diagnosis confirmed exactly as written — `AcceptInviteRequest` didn't declare `name`, pydantic's default `extra="ignore"` dropped it, and `accept_invite` never touched `user.name`.

**No frontend change.** `invite-accept.tsx` already requires the field, validates it with `!name.trim()` and includes it in the POST body. It was right all along; the API was throwing the value away. Dropping the `frontend` label would be fair.

## One place the suggested fix isn't quite enough

The issue proposes `min_length=1, max_length=255`. `min_length` counts **characters**, so `"   "` satisfies it and then strips to `""` — which `User.name` accepts, being `String(255), nullable=False`, and which renders as a blank author on every comment. That's a quieter version of the same bug: the invitee is asked, answers, and something other than their answer is stored.

So `name` uses a `DisplayName` alias that trims and rejects a blank:

```python
DisplayName = Annotated[str, Field(min_length=1, max_length=255), AfterValidator(_require_non_blank_name)]
```

This follows `NewPassword` immediately above it, which exists for the same shape of problem — a length constraint counting the wrong unit — so the file already had the pattern and the reasoning.

Trimming inside the type rather than at the call site means the router assigns the stored form directly and the rule has one home. The length bound applies **before** the trim, so the stored value is always within `String(255)`.

**Required, not optional.** The only caller has always sent it, and a silently ignored field is precisely the defect being fixed — an omission should say so rather than repeat it. It is a contract change on the endpoint, so I've pinned it with a test rather than leaving it implicit; say the word if you'd rather it were optional for compatibility and I'll switch it.

## Verification

Five tests, each part checked by reverting it rather than only run green:

| reverted | result |
|---|---|
| router no longer assigns `user.name` | **2 red** |
| `DisplayName` without the blank check | **2 red** |
| field not declared — the shipped behaviour | **5 red** |

That last row is the reproduction: undeclaring the field puts the suite back to exactly its previous 20 passing tests in `test_auth.py`, which is what the bug looked like from the outside.

There's also a control asserting the rest of `accept_invite` is undisturbed — password set, `email_verified`, `status`, and both invite-token fields cleared — since the risk in touching this handler is breaking what it already did.

**365 pass, against 360 before.**

## What I could not run

42 errors in the full API suite need a live Postgres (`role "user" does not exist`). They fail **identically on unmodified `main`** here — the dev stack wants Docker and this machine has none, so `docker compose ... pytest` from CONTRIBUTING wasn't available to me. `apps/api/tests/test_auth.py` itself is fully mock-backed and runs clean. I also could not run `pnpm --filter web test`, though this branch changes no web code.


ravirajsinh45 (changes requested):
Thanks @shaurya703. The diagnosis is right and the `DisplayName` alias is the better fix: `min_length` counting characters means `"   "` would have passed and stored a blank, and following the `NewPassword` pattern directly above keeps the rule in one place. Asserting the stored value rather than the response in the tests is the right choice too.

## Blocking

**There is no CHANGELOG entry.** This is user-visible and it also changes the endpoint contract, since `name` is now required and a body without it gets a 422. That belongs under `[Unreleased]` with the contract change called out explicitly, because a self-hoster with a provisioning script that posts token and password will start getting 422s.

Required rather than optional is the right call and we are happy with it.

## Non-blocking

The same blank-name hole is still open one request later at `users.py:94-95`, where `PATCH /users/{id}` accepts `{"name": "   "}` and stores an empty string. Not yours to fix here; we will file it so `DisplayName` gets reused there.

shaurya703:
Pushed as `e387328` (rebased onto `main` for #314 first).

**The CHANGELOG entry is in**, under `[Unreleased] → Fixed`, with the contract change as its own paragraph rather than a clause:

> **Contract change for anyone posting to this endpoint directly.** `name` is required, so a provisioning script that sends only `token` and `password` now gets a 422 rather than a user with a name derived from their email. The accept screen already sent it, so nothing changes for people going through the UI.

Its own paragraph because that reader is the one who cannot see the change coming: from the script's side the request simply starts failing, with nothing in the UI to hint why.

Credited as `(#320, #322 by @shaurya703)`, following `(#288 by @luozejian)` and the rule in `docs/RELEASING.md`.

One wording correction I made while writing it, since it is the kind of thing that ends up in release notes: the discarded name did not leave the user *nameless*, it left them with **whatever the inviter's dialog had derived from their email address**. That is `apps/web/app/(dashboard)/settings/admin/page.tsx:42`, `const name = email.split("@")[0]`, which is also how your issue described it. The entry says that rather than implying the email itself was the display name.

**On `PATCH /users/{id}`:** agreed, and thank you for filing it separately rather than growing this PR. For whoever picks it up, `DisplayName` drops in as-is there — the alias already trims and rejects blank, so it is a type change on the field and nothing else. Happy to take that one too if it is not already spoken for.

The change here is markdown only; `apps/api/routers/auth.py` and `apps/api/schemas/auth.py` are untouched since your review.

ravirajsinh45 (approved):
@shaurya703 Thanks for the quick turnaround. The one blocking item is done, and this is ready to merge.

## Blocking: all clear

**The CHANGELOG entry is exactly the shape we asked for.** `e3873287` puts it under `## [Unreleased]` in the `### Fixed` subsection (CHANGELOG.md:17), no invented version, and the contract change gets its own callout paragraph at CHANGELOG.md:19 instead of being buried inside the main sentence: it names the 422, names who it actually breaks (a provisioning script that sends only `token` and `password`), and says plainly that nothing changes for people going through the UI. It also matches the shipped code rather than describing an intention, `name: DisplayName` at apps/api/schemas/auth.py:122 has no default, so a request without it really is a 422.

**Nothing we already reviewed moved.** `git diff 4e0678ad..pr322n -- apps/api/schemas/auth.py apps/api/routers/auth.py apps/api/tests/test_auth.py` is empty, so `DisplayName`, the required field, `user.name = body.name` at apps/api/routers/auth.py:175 and the five tests are byte-identical to yesterday's read. The third commit in the range, `f299c28d`, is the upstream main merge that brought in #314, not yours.

**Worth saying: the tests are load-bearing, and the schema comment is accurate.** We re-ran the suite with the fix mutated out. Deleting `user.name = body.name` turns `test_accept_invite_stores_the_name_the_invitee_typed` and `test_accept_invite_trims_the_name` red; undeclaring `name: DisplayName` turns all five red. We also checked the ordering claim in the `DisplayName` comment (apps/api/schemas/auth.py:44-45) empirically rather than taking it on trust: on pydantic 2.9.2 `max_length=255` is applied to the raw string before the `AfterValidator` trims, so what reaches `User.name` (`String(255), nullable=False`) can never overflow the column. Putting the trim inside the type instead of at the call site is the right call, the rule has one home and every consumer gets the stored form.

## Non-blocking, nothing to change in this PR

The reusable type is the interesting part of this change, and three other writers of the same column still accept what it rejects. All three are pre-existing, none are introduced here, and we are not asking you to widen the diff:

- `UpdateProfileRequest.name: str | None` (apps/api/schemas/auth.py:137) plus `user.name = body.name.strip()` at apps/api/routers/users.py:95 trims but has no blank check, so `PATCH /users/{id}` with `{"name": "   "}` answers 200 and stores `""`. That is #329, which we filed ourselves; correct call leaving it alone.
- `InviteRequest.name: str` (apps/api/schemas/auth.py:99) is unbounded and untrimmed and goes straight into `User.name` at apps/api/routers/users.py:69, so a name over 255 characters is a Postgres truncation 500 rather than a 422.
- `guest_name` (apps/api/schemas/comment.py:30) is guarded only by `if not body.guest_name` at apps/api/routers/comments.py:941, which catches `""` and `None` but not `"   "`, so apps/api/routers/comments.py:946 can create a `GuestUser` whose name is whitespace.

Applying `DisplayName` to the first two is a one-word change each. A separate PR is welcome if you want them, and we will pick them up otherwise.

One cosmetic note with no action needed: a name over 255 characters typed on the accept screen surfaces pydantic's raw validation message in the form's general error banner rather than under the Name field, because apps/web/lib/api.ts:64-68 joins FastAPI's 422 detail array into a string. It renders as text and does not break the screen.

Approving. Heads up that #322 through #327 all touch CHANGELOG.md, so whichever lands second will conflict there in the usual way; we will handle the ordering on our side.

PR #323, opened 2026-09-05
fix(web): mobile video scrubber unusable — no drag support, tiny hit area

CamReport:
## Summary

The video progress bar's playhead could not be dragged on a touchscreen — only a single tap-to-seek worked, and even that used a target too small to reliably hit with a finger.

## Changes

- Every mouse-event handler (and the window-level `mousemove`/`mouseup` listeners previously used for drag-outside-track) is now its Pointer Events equivalent, unifying mouse, touch, and pen through one code path. `touch-action: none` (`touch-none`) on the track stops the browser's own touch-scroll gesture from competing with an in-progress drag.
- The dragging pointer is captured on `pointerdown` (`setPointerCapture`), so it keeps driving the seek even once it moves outside the track's own bounds — this is what lets the window-level listeners be dropped entirely. Every handler filters by the captured `pointerId`, so a second pointer (e.g. a two-handed grip) can't hijack an in-progress drag. `pointerdown` also filters by `button`/`isPrimary`, and the drag ends via `onLostPointerCapture` if capture is lost without a matching release ever landing on the track — without these, a right-click-and-drag-away could permanently latch the bar dead for the life of the mounted component.
- The drag ends cleanly on `pointercancel` too (iOS fires this instead of `pointerup` when the system takes the gesture mid-drag — a pinch starting, an edge swipe, a long-press callout) — without this, every later pointer move anywhere on the page kept seeking the video.
- The touch target is widened with an invisible overlay, asymmetric (`-top-1` / `-bottom-3.5`) rather than a symmetric inset: it stays inside this component's own padding above (there's no room to spare into the video area rendered directly above it — going further up would intercept clicks and, in drawing mode, annotation strokes meant for the video/canvas) while clearing into the transport bar's own button gutter below. The comment-markers row is now `pointer-events-none` (each individual marker `pointer-events-auto`), so it no longer swallows most of that downward extension on an asset with an unresolved comment.
- The floating frame-preview/timecode tooltip's position is now clamped to the track's own width — a captured drag can otherwise push it past the player's edge, e.g. over the comment panel, even though the seek itself was already clamped.
- The playhead thumb was `opacity-0` until `group-hover`, meaning it was never visible on a device that can't hover. Default flipped to always-visible, with the hide/reveal-on-hover behavior moved behind `[@media(hover:hover)]:`, so desktop keeps the exact original look.

No visual or behavioral change on desktop (mouse) — confirmed against real computed layout, not just by inspection: the track's className is byte-identical to `main`'s apart from `touch-none` and a test id.

## Testing

- [x] Frontend tests (`pnpm test`) — 7 tests in `progress-bar.test.tsx` covering the drag lifecycle, two-pointer rejection, `pointercancel`, non-primary pointer/button rejection, `onLostPointerCapture` recovery, and hover-tooltip clamping; no new failures elsewhere
- [x] Tested manually (drag tracks continuously, the target is easy to catch without reaching into the video area or being swallowed by comment markers, the thumb is visible without needing to touch it first) and on desktop (hover behavior — thumb reveal, track thickening — unchanged)

## Checklist

## Screenshots

No visual change on desktop. Happy to add a screen recording of the mobile drag behavior if useful.


ravirajsinh45 (changes requested):
Thanks @CamReport. Pointer events are the right unification, and putting the ref on the hit area rather than the visual track is correct: both are `w-full`, so `getBoundingClientRect` gives identical width and left and the seek math is untouched. Moving the thumb reveal behind `[@media(hover:hover)]` is the right way to keep the desktop look.

## Blocking

**No `pointercancel` handler.** `progress-bar.tsx:358-359` listens for `pointermove` and `pointerup` only, and `setIsDragging(false)` lives solely in the pointerup path. iOS fires `pointercancel` whenever the system takes the gesture: a second finger starting a pinch, the bottom edge swipe, a long-press callout. The drag then never ends, the window listeners stay attached, and every later pointer move anywhere on the page seeks the video. On the devices this PR is for, that is worse than the bug it fixes.

**No `pointerId` filter.** `handleGlobalPointerUp` accepts any pointer, so on a two-handed phone grip the left thumb lifting off ends the right thumb's drag and seeks to the left thumb's x position. Capturing `pointerId` on pointerdown and ignoring other ids fixes both this and the cancel case cleanly, and `setPointerCapture` would let you drop the window listeners entirely.

**"No visual change on desktop" is not accurate.** The hit area at `:383` is `relative w-full h-6 flex items-center`, in normal flow where the 4px track used to be, so the row grows 20px on every device. The video area is correspondingly shorter and the comment markers below shift down. Absolutely positioning the hit area over the track, with `-inset-y-2.5` or similar, gets the touch target without the reflow.

## Non-blocking

`preventDefault()` at `:336` suppresses the compatibility `mousedown`, and there are 11 `document.addEventListener("mousedown")` outside-click closers in the app, including `video-player.tsx:151` on this same screen. After this, clicking the scrubber no longer closes an open format picker.

A test would be cheap here: pointerdown then pointermove then pointerup on the track, asserting the `onSeek` calls, fails on `main` and passes on this branch.

CamReport:
Thanks for the thorough review — all three blocking points confirmed against the code and fixed in 119e519:

- **pointercancel**: added a handler, guarded by the same pointerId check as pointerup, so a cancelled gesture ends the drag cleanly instead of leaving it stuck.
- **pointerId filtering**: the dragging pointer's id is captured on pointerdown (`setPointerCapture`) and every subsequent handler ignores events from any other pointer. This also let me drop the window-level `pointermove`/`pointerup` listeners entirely, per your suggestion — the captured element now receives them directly even outside the track's bounds.
- **Reflow**: the hit area is no longer a taller wrapper in normal flow. It's back to the original structure (handlers on the visual track div, matching upstream), with an absolutely-positioned invisible child that overflows outside the track's own box (`-inset-y-2.5`) rather than growing it — so there's no layout change on any device.

Also dropped the leftover `preventDefault()` on pointerdown (your non-blocking note) — it doesn't just suppress touch defaults, it suppresses the synthesized compatibility `mousedown` for *any* pointer type, which is exactly what broke the outside-click closers you flagged. Nothing in this component actually needs it (no selectable text nearby, and `touch-action: none` already covers the touch-scroll case), so it's just gone.

Added a test covering the drag lifecycle, the two-pointer case, and pointercancel — the scenario you called out as untested. All 3 pass; full `components/review` suite (120 tests) still green.

ravirajsinh45 (changes requested):
@CamReport this is a big step up, and all three blocking items are genuinely fixed. Worth naming what you did, because on two of them you went further than we asked.

**1. `pointercancel`.** Fixed, and better than a bolt-on handler. The teardown is now a shared `endDrag` (progress-bar.tsx:311-316) that clears `activePointerIdRef`, `isDragging`, `hoverTime` and the preview together, and both `onPointerUp` (:361) and `onPointerCancel` (:370) route through it, so `setIsDragging(false)` no longer lives in a single path.

**2. `pointerId` filter.** Fixed on all four paths, not just pointerup: :347 refuses a second pointerdown while a drag is live, :322 filters moves, :363 and :375 filter up and cancel. We walked the second-finger case end to end: pointer 1 is explicitly captured so pointer 2 is never retargeted to the track, and if it does land there its pointerdown early-returns without taking the latch.

**3. Hit area.** Fixed the way we suggested. The track's class list is now byte-identical to main's apart from `touch-none` and the testid, the `h-6 flex items-center` wrapper is gone, and net layout height change is zero. Gating the thumb behind `[@media(hover:hover)]` at :446 was a catch we did not ask for; the PR body's desktop claim now holds for the thumb too.

Both non-blocking notes are handled as well. Dropping `e.preventDefault()` restores the compatibility mousedown that the 11 document-level outside-click closers depend on, including the time-format dropdown at video-player.tsx:145-153 on the same screen, and `touch-none` keeps scroll suppressed without it. The three tests are real, not vacuous: reverting the `pointerId` guard at :322 and the `onPointerCancel` registration at :405 one at a time each turns a different test red.

Deleting the two `window` listeners in favour of `setPointerCapture` was the right call, and it is the part we would have asked for next.

One thing is still blocking, and it is in the new geometry rather than in the pointer logic.

## Blocking

**The hit-area overlay overflows the ProgressBar's own box, so it takes input from the bottom of the video, and the markers row eats most of what is left.**

`-inset-y-2.5` at :410 is 10px in each direction on a 4px track (:400), and the root's only padding is `py-1`, 4px (:393). So the overlay's box runs from `root.top - 6` to `root.bottom + 6`. In video-player.tsx the ProgressBar's wrapper (:356) has no padding of its own and sits directly under the video area (:319), so those 6px land inside the video container. Nothing in that container carries a z-index: the container is `relative`, the `<video>` is `absolute inset-0`, and VideoFrameConstraint sets `position: absolute` with left/top/width/height and nothing else (video-player.tsx:86-92). The overlay is a positioned element later in tree order in the same stacking context, so it wins the hit test across the whole overlap.

Three consequences, all on the container-wider-than-aspect case where the picture fills the container height and its bottom edge is the container's bottom edge:

- A click on the bottom 6px of the picture seeks instead of reaching `handleContainerClick` (video-player.tsx:289).
- In drawing mode it is worse. A stroke started in that band never reaches the annotation canvas and seeks the video instead, moving the frame the annotation is being drawn against.
- On desktop, just moving the mouse along the bottom edge of the video bubbles `pointermove` to the track (:401), which pops the frame preview and grows the bar. That is a visual change on desktop, which is the thing the PR body says it avoids.

Downward, the overhang is mostly dead. The markers row at :453 is `relative w-full h-6 mt-0.5`, a positioned sibling rendered after the track with no `pointer-events-none`, so it covers 8 of the overlay's 10 downward pixels on any asset with an unresolved timecoded comment, which is the normal state for a review tool. The delivered target there is 10 above plus 4 track plus 2 below, so 16px rather than 24, biased upward, and it silently changes shape depending on whether the asset has comments.

These have to be fixed together: clamping the top alone would drop the target to 10px on a commented asset.

The requirement is just that the hit area does not reach above the ProgressBar's box. One shape that gets there, if it helps:

- `-top-1 -bottom-3.5` on the overlay. The 4px up stays inside the root's own `py-1`, and 14px down clears the root's bottom padding by 10px, which is exactly the transport bar's button gutter ((48 - 28) / 2) when there are no markers, so it overlaps nothing in either state. That is 22px.
- `pointer-events-none` on the markers wrapper at :453 with `pointer-events-auto` on the dots, so the downward extension is actually live when the asset has comments.

If you want the full 24px, grow the root's bottom padding rather than the overlay's overhang. That pushes the transport bar down a few pixels, which is a trade we would take over covering the picture, and it is nothing like the 20px in-flow growth we objected to. Your call on the shape.

## Non-blocking

**The latch has no recovery path.** `activePointerIdRef` is cleared only in `endDrag` (:311), reachable only from `onPointerUp` and `onPointerCancel` on the track. There is no `onLostPointerCapture`, and `handlePointerDown` (:345) arms the latch and captures for any button with no `e.button` or `e.isPrimary` check. If the matching release never lands on the track, every later pointerdown returns at :347 and the bar is dead for the life of the mounted component, while `isDragging` stays true. Right-clicking the bar is the easiest way in. The seek on right-click is not new, main's mouse path did not filter the button either, but the permanently dead bar is, because the old code self-healed on the next press. Two lines close it: `if (!e.isPrimary || e.button !== 0) return` at the top of `handlePointerDown`, and `onLostPointerCapture={endDrag}` on the track.

**The comment at :351-353 is not right.** It says skipping `setPointerCapture` is harmless because "the pointerId filtering above still holds either way". Without capture the pointerup lands on whatever element is under the pointer, so `endDrag` never runs and the stuck drag this PR set out to fix comes back; the pointerId filter cannot help when the event never reaches the track at all. Worth correcting, because capture is now the sole drag-outside-track mechanism after the window listeners were dropped, and jsdom does not implement it, so deleting the call leaves all three tests green. A line in the test file noting that capture is not covered here would help whoever touches this next.

**`hoverX` is unclamped during a captured drag.** :328 sets `setHoverX(e.clientX - rect.left)` with no clamp, and that feeds the tooltip's `left` at :486. The deleted window handler only called `onSeek`, so this is new: drag past the player's left edge and the 160px frame preview and timecode render outside the player, over the comment panel, while the label stays pinned at 00:00. The seek itself is still clamped by `getTimeFromEvent`, so it is only the floating UI.

**CHANGELOG.** Add attribution to match the entry above it, `(#323 by @CamReport)`. Also worth rebasing on main before we merge; the branch predates the parallel HLS upload entry, so the CHANGELOG will conflict as it stands.

**Not this PR.** compare-scrubber.tsx:198 is still `onClick` only, with no drag and no widened target, so the compare view has exactly the gap you are closing for the main player. Happy to file that as a follow-up issue rather than growing this one.

Everything else here is ready. Sort out the hit-area geometry and we will get this merged.

CamReport:
Thanks — the geometry finding in particular is a genuinely different bug than the one this PR started with, and I hadn't traced the layering into `video-player.tsx` at all. Fixed in e0794c7:

- **Hit-area geometry**: replaced the symmetric `-inset-y-2.5` with `-top-1 -bottom-3.5`, matching the shape you suggested — stays inside this component's own `py-1` above (so it can't reach into the video area), clears into the transport bar's button gutter below. Also added `pointer-events-none` on the markers row with `pointer-events-auto` on each marker, so the row's own background no longer swallows most of the downward extension when there's an unresolved comment.
- **Recovery path**: added `if (!e.isPrimary || e.button !== 0) return` to `handlePointerDown` and `onLostPointerCapture={endDrag}` on the track — a right-click-and-drag-away (or any release that doesn't land back on the track) no longer permanently latches the bar.
- **The misleading comment**: corrected — it now explains that without capture, `pointerup` can land off-track and `endDrag` may never run there, rather than claiming the pointerId filter "holds either way."
- **`hoverX` clamp**: clamped to `[0, track.width]`, same bound `getTimeFromEvent` already applies to the seek itself — a captured drag past the edge no longer pushes the frame-preview tooltip over the comment panel.
- **CHANGELOG**: rewrote the entry (the previous revision's commit never actually updated it — it was still describing the round-1 fix) and added the `(#323 by @CamReport)` attribution. Rebased onto current `main` to resolve the conflict with the HLS-upload entry.

Added 4 tests for the new behavior (non-primary pointer, non-primary button, `onLostPointerCapture` recovery, tooltip clamping) and fixed the 3 existing ones, which didn't set `isPrimary`/`button` on their synthetic `pointerdown` events — jsdom defaults those falsy, unlike a real press, so they were passing for the wrong reason once the new guard existed. All 7 pass; full suite shows the same pre-existing failures as clean `main`, no new ones.

Also updated the PR description to reflect where the fix actually stands now rather than the round-1 text.

Not filing the `compare-scrubber.tsx` follow-up myself since you'd already offered to — let me know if you'd rather I take it.

ravirajsinh45 (approved):
@CamReport both blocking items are fixed, and the geometry one is fixed exactly right. Approving.

**The hit area no longer reaches into the video.** `-top-1 -bottom-3.5` puts the overlay's top at the ProgressBar root's own top edge, so there is nothing left to bleed upward. We measured it in a browser rather than reading the classes: the overlay spans -4 to +18 relative to the track top and the root spans -4 to +8, so the top edges coincide exactly.

**And the target is a real 22px.** We probed `elementFromPoint` down the whole overlay at the track's midpoint: **22 live pixels**, -4 through +17. We also checked the case that was actually broken before, an asset with an unresolved timecoded comment, and it is also 22, with the marker itself still clickable. That was the part worth verifying, since the old behaviour silently changed shape depending on whether the asset had comments. `pointer-events-none` on the row with `pointer-events-auto` on each marker is the right way to get that.

One thing we got wrong and you did not: we assumed the downward overhang past the root would lose hit-testing to the transport bar, which is a later sibling with a background. It does not, because the overlay is positioned and the transport bar is not, so the overlay paints in the later phase. Your 10px into the button gutter is genuinely live. We only know that because we measured it.

**The recovery path and the rest.** `!e.isPrimary || e.button !== 0` and `onLostPointerCapture={endDrag}` both hold up under mutation: deleting the guard turns two tests red, removing the capture handler turns one red. The `hoverX` clamp matches the bound `getTimeFromEvent` already applies. The corrected comment about capture is accurate now, including the note that jsdom does not implement it.

Seven tests, all passing, and they fail for the right reasons when the code they cover is removed. Catching that the three original tests were passing without setting `isPrimary`/`button` was a good spot; jsdom defaulting those falsy would have made them fail the moment the guard landed.

**We pushed one commit to your branch**, resolving a `CHANGELOG` conflict we caused by cutting v1.13.0 underneath you. Your entry moved to the new `[Unreleased]`; the released section is untouched. Nothing else changed.

On `compare-scrubber.tsx`: we will file that one, as offered. Thanks for a genuinely careful round.

PR #324, opened 2026-09-05
fix(web): comment inputs trigger iOS Safari auto-zoom on focus

CamReport:
## Summary

iOS Safari auto-zooms the whole page whenever focus lands on a text input with a computed font-size under 16px (so the zoomed-in text stays legible). All four text inputs in the review UI were 13px, so tapping into any of them on an iPhone zoomed the page in.

## Changes

- `comment-input.tsx`: the main comment textarea bumped from `text-[13px]` to `text-[16px]` on mobile, with `md:text-[13px]` reverting to the original size on desktop, where this Safari behavior doesn't apply.
- `comment-panel.tsx`: same treatment for the reply input, the comment-edit textarea, and the comment search input.
- `text-[16px]` (an explicit pixel value, the literal iOS threshold) was used rather than the `text-base` utility, to avoid depending on it matching exactly if the theme's type scale ever changes.

## Testing

- [x] Frontend tests (`pnpm test`) — same 30 pre-existing failures present on unmodified `main`; no new failures
- [x] Tested manually on a real phone — tapping into the comment box, the reply box, an existing comment's edit box, and the comment search box no longer triggers Safari's zoom. Desktop font sizes are unchanged.

## Checklist

## Screenshots

No visual change beyond eliminating the zoom-on-focus behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - **Default logos no longer load an invisible second theme asset** — sidebar, authentication, share-password, attribution and branding-preview fallbacks now share one themed icon/full-logo component. A failed custom authentication logo also falls back to the correct built-in mark and retries when branding supplies a new URL. (#288 by @luozejian)
+- **Focusing a comment input zoomed the whole page on iOS Safari** — Safari auto-zooms the page whenever focus lands on a text input with a computed font-size under 16px, and the main comment textarea, the reply input, the comment-edit textarea, and the comment search input were all 13px. Each is now 16px on mobile (the literal iOS threshold), reverting to the original 13px at the `md` breakpoint, where this Safari behavior doesn't apply.
 
 ## [1.12.0] - 2026-08-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - **Default logos no longer load an invisible second theme asset** — sidebar, authentication, share-password, attribution and branding-preview fallbacks now share one themed icon/full-logo component. A failed custom authentication logo also falls back to the correct built-in mark and retries when branding supplies a new URL. (#288 by @luozejian)
-- **Focusing a comment input zoomed the whole page on iOS Safari** — Safari auto-zooms the page whenever focus lands on a text input with a computed font-size under 16px, and the main comment textarea, the reply input, the comment-edit textarea, and the comment search input were all 13px. Each is now 16px on mobile (the literal iOS threshold), reverting to the original 13px at the `md` breakpoint, where this Safari behavior doesn't apply.
+- **Focusing a comment input zoomed the whole page on iOS Safari** — Safari auto-zooms the page whenever focus lands on a text input with a computed font-size under 16px, and the main comment textarea, the reply input, the comment-edit textarea, the comment search input, and the guest name/email prompt on share links were all under that threshold. Each is now 16px on a touch device, reverting to its original compact size only under a `(hover: hover)` media query — not a `md` viewport-width breakpoint, which still zoomed on a landscape phone (an iPhone in landscape is wider than `md`'s 768px).
 
 ## [1.12.0] - 2026-08-29
 
--- a/apps/web/components/review/__tests__/comment-input.test.tsx
+++ b/apps/web/components/review/__tests__/comment-input.test.tsx
@@ -212,3 +212,15 @@ describe('CommentInput compare drawing props (annotationActive / onToggleAnnotat
     expect(screen.getByTitle('Draw annotation').className).not.toContain('text-accent')
   })
 })
+
+describe('CommentInput iOS auto-zoom guard', () => {
+  it('sizes the textarea at the 16px iOS zoom threshold, and reverts to compact only under a real hover/pointer query — never a viewport-width breakpoint', () => {
+    render(<CommentInput assetId="a1" projectId="p1" assetType="video" onSubmit={vi.fn()} />)
+    const textarea = screen.getByPlaceholderText('Leave your comment...')
+    expect(textarea.className).toContain('text-[16px]')
+    // A `md:` (or any other width) breakpoint here would still trigger Safari's
+    // auto-zoom on a landscape phone, which is wider than the breakpoint.
+    expect(textarea.className).not.toMatch(/\bmd:text-/)
+    expect(textarea.className).toContain('[@media(hover:hover)]:text-[13px]')
+  })
+})
--- a/apps/web/components/review/comment-input.tsx
+++ b/apps/web/components/review/comment-input.tsx
@@ -465,7 +465,7 @@ export function CommentInput({
             )}
             <textarea
               ref={textareaRef}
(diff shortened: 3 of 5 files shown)


ravirajsinh45 (changes requested):
Thanks @CamReport. The diagnosis is exactly right, and using the literal `text-[16px]` rather than `text-base` so a theme change cannot silently re-arm it is a good call.

## Blocking

**`md:` gives the zoom back in landscape.** `md` is 768px, and an iPhone 15 Pro is 852px wide in landscape, a Pro Max 932px. Rotating the phone to watch a 16:9 cut is the normal posture for video review, and in that orientation `md:text-[13px]` wins and tapping the composer zooms exactly as before. The CHANGELOG entry states the breakpoint is "where this Safari behavior doesn't apply", which is not true of any current iPhone. A pointer or hover media query separates phone from desktop the way this needs; `md:` is answering a layout question.

**The very next input in the flow still zooms.** `share-review-screen.tsx:284-297` renders the guest identity prompt with `text-sm` inputs and `autoFocus`. A guest types a comment in the now-16px box, taps Send, and the modal zooms the page without them touching anything.

## Non-blocking

Worth deciding now: `components/ui/input.tsx:40` is `text-sm`, and it backs sign-in, invite-accept and setup. Fixing it there covers roughly thirty call sites in one line and makes the four here unnecessary. Happy for that to be a follow-up if you would rather keep this PR narrow.

One correction on the PR body: CI runs `pnpm test` inside the Frontend Build job and it is green on this branch, so the 30 failures you saw look local rather than pre-existing on `main`.

CamReport:
Thanks for catching both — fixed in 29edfde:

- **Landscape breakpoint**: you're right that `md:` was answering a layout question, not an input-capability one. Replaced it with `[@media(hover:hover)]:` on all four inputs — reverts to the compact size only on a device that can actually hover (mouse/trackpad), regardless of viewport width, so a landscape iPhone at any width still gets the 16px guard.
- **Guest identity prompt**: confirmed — `share-review-screen.tsx`'s `GuestIdentityPrompt` (the name/email modal) was out of scope of my original diff entirely. Both inputs were `text-sm` (14px), and the name field has `autoFocus`, so it was actually worse than the reported bug — it zoomed on open, before a guest touched anything. Fixed with the same `text-[16px] [@media(hover:hover)]:text-sm` pattern, preserving the original `text-sm` size on desktop exactly.

On the `components/ui/input.tsx` follow-up — agreed that's the better long-term fix for the ~30 other call sites, but keeping this PR narrow to the reported comment-input bug as planned. Happy to open that as a separate PR/discussion if useful.

And you're right about CI — I mischaracterized the 30 local test failures as pre-existing on `main` without actually diffing them against a clean CI-equivalent run; sorry for the noise there. Added a regression test asserting the zoom-guard survives at the class level (16px unconditionally, gated back down only by the hover query, never a width breakpoint).

ravirajsinh45 (approved):
@CamReport This is exactly the fix we were hoping for, and you took it further than we asked.

**Both blockers are resolved.**

**1. The `md:` breakpoint is gone, replaced with a capability query.** All four comment inputs now gate on `[@media(hover:hover)]:` instead of viewport width: `comment-input.tsx:468`, and `comment-panel.tsx:289` (reply), `:567` (edit), `:1242` (search). `git grep 'md:text-'` over the touched files returns nothing. That answers the input-capability question instead of the layout question, which is the right frame: iOS Safari reports `hover: none` in both orientations, so the 16px guard now holds on a landscape iPhone at any width.

We checked the part of this that could have failed silently. A media query adds no specificity, so if Tailwind had emitted the `@media(hover:hover)` block *before* the bare `.text-[16px]` rule, every desktop input would have quietly jumped to 16px and a className-only test would still have passed. Building the real CSS with the repo's own `apps/web/tailwind.config.ts` (Tailwind 3.4.17): the bare `.text-\[16px\]` rule lands first and the `@media(hover:hover)` block after it, so at equal specificity the compact size wins on hover-capable devices only. The arbitrary variant also extracts correctly from the JSX string literals. It does what it says on both desktop and touch.

`CHANGELOG.md:17` is rewritten and now accurate, including the reason the old approach was wrong ("not a `md` viewport-width breakpoint, which still zoomed on a landscape phone").

**2. The guest identity prompt is covered.** `share-review-screen.tsx:289` (name, still `autoFocus` at `:290`) and `:297` (email) are now `text-[16px] [@media(hover:hover)]:text-sm`, so the modal no longer zooms on open or on Continue, and desktop keeps its original 14px. Good instinct preserving `text-sm` there rather than flattening it to the composer's 13px.

**The regression test is the right shape.** `comment-input.test.tsx:216-225` asserts the 16px base, asserts the hover variant, and explicitly asserts `not.toMatch(/\bmd:text-/)`. A test that forbids the specific wrong fix from creeping back is worth more than one that just pins the current string. 21 tests in that file pass at this head.

---

### Non-blocking

None of these hold up the merge.

**The share password gate is the one real gap left.** `app/share/[token]/page.tsx:174` is still `text-sm` and line `173` is `autoFocus`, and `PasswordGate` renders live from `:389` when `state.stage === 'password_required'`. It sits one screen *before* the name/email prompt you just fixed, on the same guest flow the CHANGELOG now claims to cover, and because it autofocuses it zooms on mount rather than on tap. The input is fixed-height `h-9`, so `text-[16px] [@media(hover:hover)]:text-sm` costs no layout change. We are scoping this to the password gate specifically rather than asking you to sweep all ~36 sub-16px inputs in `apps/web`; it is just the first thing a share-link guest sees. Fold it in here if you like, or leave it and we will pick it up.

**Folder-browse search, same flow.** `components/share/folder-share-viewer.tsx:1063` ("Search assets…") is `text-sm` and live, rendered via `page.tsx:431`, with no `hidden md:` guard on its wrapper. It is structurally identical to the `comment-panel.tsx:1242` search input you did fix, so leaving it reads as an oversight rather than a boundary. Worth one more line while you are in there. For the record, the `text-xs` guest name/email inputs at `:613`/`:620` in that same file belong to `ShareCommentInput`, which is defined and never rendered; you were right to leave those alone.

**The regression guard covers 1 of the 6 changed inputs.** Only the `CommentInput` textarea is asserted on. That matters slightly more than usual because #326 is concurrently rewriting `comment-panel.tsx`'s toolbar region, with its hunk about five lines above the search input you changed at `:1242`. The two branches merge clean today, so this is prevention, not a live break; one parameterised assertion over the three `comment-panel` inputs would close it.

**CHANGELOG credit.** The entry at `CHANGELOG.md:17` ends without a `(#324 by @CamReport)` reference. The entry directly above it under the same `### Fixed` heading carries `(#288 by @luozejian)`, and we credit every external contributor by handle. Please add it so the release ships with your name on it.

**Please update the PR description before merge.** The body is now stale in three ways: it says "All four text inputs" (six now), it still describes `md:text-[13px]` "reverting to the original size on desktop, where this Safari behavior doesn't apply" (the exact claim you correctly removed from the CHANGELOG), and it still asserts "same 30 pre-existing failures present on unmodified main", which you already accepted in comments as wrong. Since the body becomes the merge commit message, we would rather it carry the new rationale than the old one.

**On `components/ui/input.tsx:40`:** agreed it is the better long-term fix, and leaving it out of this PR is right. We will open a follow-up issue for it so the separate PR you offered has something to close, since nothing is tracking it yet.

---

Approving. Two things before we merge: the PR description update, and CI. `gh pr checks 324` reports no checks on `29edfdef`, so the fork-CI gate needs approving on our side to get Frontend Build running on this head; we will do that now. Thanks for turning this around quickly and for picking the fix that actually answers the question.

Issue #319, opened 2026-09-04
Make the quality ladder configurable, and skip re-encoding a source that is already browser-safe


Lennart-Pingpong:
### Problem

<head></head><p cid="n2" mdtype="paragraph" class="md-end-block md-p md-focus" style="font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; text-align: start; text-indent: 0px; text-transform: none; white-space: pre-wrap; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; box-sizing: border-box; line-height: inherit; orphans: 4; margin-top: 0px; margin-bottom: 1.5rem; overflow-wrap: break-word; caret-color: rgb(184, 191, 198); color: rgb(184, 191, 198); font-family: &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, &quot;SF Pro&quot;, sans-serif; background-color: rgb(54, 59, 64);"><span md-inline="plain" class="md-plain md-expand" style="box-sizing: border-box;">Every upload is re-encoded into three renditions, and neither half of that is</span><span md-inline="softbreak" class="md-softbreak" style="box-sizing: border-box;">
[...]

Issue #82, opened 2026-06-24
mpeg audio only: You did not specify framerate for Video description [1] and there is no input framerate to follow.


casinojeffrey:
### Describe the bug

I uploaded an mpeg audio only and after processing the sytem is telling me: "You did not specify framerate for Video description [1] and there is no input framerate to follow. To resolve: Specify framerate, and resubmit your job."
any ideas why this is happening?

### Steps to reproduce

login to freeframe.io
upload an mpeg audio only file
upload will work but once it processes, the error warning appears

### Expected behavior

upload should work

### Screenshots

<img width="259" height="45" alt="Image" src="https://github.com/user-attachments/assets/0a4fca7f-dbd4-4668-80da-551b8f0c9009" />

### Area

Media Processing

ravirajsinh45:
hello @casinojeffrey, 
thank you for raising the issue, can you send exact file you trying to upload, i tried to reproduce issue with several mpeg audio file but it working normally as expected. drive link will be perfect

casinojeffrey:
hi @ravirajsinh45 i can do that in private as this is a client-related file that i cannot upload publicly. can you suggest a way to deliver the file?

ravirajsinh45:
@casinojeffrey 

ravirajsinhdabhi86@gmail.com 

You can sent me on the email if it not problem for you

PR #325, opened 2026-09-05
feat(transcoder): let a deployment choose which quality rungs to build

Lennart-Pingpong:
## Summary

Part 1 of #319. The ladder was a literal at the call site, so every upload was
encoded into 1080p, 720p and 360p whatever the instance was for.

`TRANSCODER_QUALITIES` defaults to exactly that ladder, so nothing changes for
anyone who does not set it. An instance whose reviewers all watch at full size
can now ask for fewer rungs and stop paying for the others in encode time and in
storage.

It lives in `apps/api/config.py` rather than following the `os.environ` pattern
of the other `TRANSCODER_*` vars, as you asked — that pattern exists because
`packages/` cannot import `apps.api.config`, and `transcode_tasks.py` already
imports settings.

## The rung above the source, which is where your first correction landed

You were right that the never-empty fallback restored such a rung and then
upscaled it, since `force_original_aspect_ratio=decrease` prevents exceeding the
box but not scaling up to it. Following your second review, the kept rung is now
clamped to the probed source size instead of restored at its nominal one. So
`TRANSCODER_QUALITIES=1080p` on a 640x360 upload produces one 640x360 rendition
rather than one upscaled 1920x1080 one, and #201 can no longer be reached
through configuration.

That is not the same as the `source` rung from the third part of #319, which
would skip the scale filter altogether. This still scales — to the size it
measured. Where nothing was probed there is nothing to clamp against and the
rung stands as before.

## The validation, which is most of the diff

This is the half that makes it more than a two-line patch, and it is your point
from the issue:

- An unrecognised rung name was dropped with no word anywhere.
- A value where *every* name was unrecognised produced `split=0` and an empty
  `-var_stream_map`, which ffmpeg rejects — so one typo failed every upload,
  through three Celery retries, ending at `failed` with nothing pointing at the
  setting.

`parse_qualities` now reports unknown names and falls back to the full ladder
when nothing valid is left. Falling back rather than refusing to start matches
how an unrecognised `TRANSCODER_OUTPUT` already resolves to its default via
`_OUTPUT_MODES.get(..., "h264_8")`; the difference is that this one says so.

Guarded in two places on purpose. `parse_qualities` covers the configured path.
The filter inside `transcode()` also falls back, because a `TranscodeJob` can be
built by any caller — `qualities` has a plain list default — and that filter is
the line which turns junk into `split=0`.

`QUALITY_MAP` moves to module level, since the rung names are configuration now
and something outside the method has to tell a valid one from a typo.

## Ordering

The rungs come back in ladder order rather than in the order they were typed,
because that order decides the variant indices in the manifest and should not
depend on how an operator happened to write the list. Duplicates collapse.

## Measured, not asserted

You corrected three claims of mine across the other issues, all of them things I
derived from reading rather than ran. So this one was run: real `transcode()`
against a real 1280x720 H.264 file, S3 stubbed, counting the rendition
directories that come out and probing each one's actual resolution.

| `TRANSCODER_QUALITIES` | renditions | resolutions |
|---|---|---|
| `1080p,720p,360p` (default) | 2 | 1280x720, 640x360 |
| `720p` | 1 | 1280x720 |
| `1080p` | 1 | 1280x720 |
| `1080,720,360` (typo) | 2 | 1280x720, 640x360 |
| `720p,quatsch` | 1 | 1280x720 |

Row 1 is the filter dropping the rung above the source, unchanged.

Row 3 is the clamp, re-measured after your second review. The same run produced
**1920x1080** before it: a rung above the source came back from the never-empty
fallback and was scaled up to its nominal size. It is the source's own size now.

Row 4 is the failure this closes. Both warnings appear and the transcode
completes; before, that value produced `split=0` and ffmpeg refused, failing
every upload through the retry ladder.

## Testing

- [x] Backend tests pass — 421 passed, 19 new: 12 in
      `test_quality_ladder_config.py` for the parsing, and 7 in
      `test_quality_ladder_applied.py` connecting the setting to a job and the
      job to a `-filter_complex`.
- [x] Frontend tests + build pass — untouched by this change.
- [x] Tested manually in browser — the effect is server-side and a browser only
      shows that the result plays; the table above measures the thing itself.

One case worth naming: an empty or unset value returns the default **without** a
warning. That is not a typo, it is an unset value, and warning about it would
train people to ignore the warnings that matter.

## Checklist

- [x] `CHANGELOG.md` entry added under `## [Unreleased]`, and
      `TRANSCODER_QUALITIES` documented in `.env.example`.


ravirajsinh45 (changes requested):
Thanks @Lennart-Pingpong. Validating the rung names at the configuration boundary and again in the transcoder is the right instinct, since a job can reach that code from any caller.

## Blocking

**The shipped text and the shipped code disagree.** `.env.example:140` says "Rungs above the source resolution are dropped", and `config.py` and the CHANGELOG say the same. But the height filter at `ffmpeg_transcoder.py:593-596` empties the ladder, and then `:597-599` puts the smallest requested rung back. Set `TRANSCODER_QUALITIES=1080p`, upload a 640x360 file, and you get a single upscaled 1920x1080 rendition: more encode time and more storage than the source, which is the opposite of the motivation in your own CHANGELOG line.

Clamping the rung to the source height instead of restoring it gives a one-rung ladder that actually means "source size", and it also keeps #201 from re-entering through configuration.

**The tests do not connect the setting to a transcode.** Revert `transcode_tasks.py:147` to the old literal and all twelve tests in `test_quality_ladder_config.py` still pass, because they exercise `parse_qualities` only. One test asserting the `-filter_complex` a configured ladder produces would pin the feature.

## Non-blocking

This is the first half of #319. The passthrough half is the one with the large win on the reporter's 36-minute master, so please drop the closing keyword and we will keep #319 open for it.

Lennart-Pingpong:
Thanks — you are right on both, and the first is the more embarrassing of the two because the text I shipped described the behaviour I meant rather than the behaviour I wrote.

The kept rung is clamped to the source rather than restored at its nominal size. TRANSCODER_QUALITIES=1080p on a 640x360 upload produces one 640x360 rendition. That makes a one-rung ladder above the source mean "source size", which is the only reading of it that saves anything, and it closes the last route back into #201 — including through configuration, which is the route this PR opened.

.env.example, config.py and the CHANGELOG say that now. test_source_smaller_than_all_requested_keeps_smallest asserted the old upscale, so it asserts the clamp instead, and there is a CHANGELOG entry under Fixed since the fallback also fires with the default ladder on a source below 360p.

Where nothing was probed there is nothing to clamp against and the rung stands as before. That path is the audio-in-a-video-container case from #82, which never reaches ffmpeg anyway.

The tests connect to a transcode now. You were right that reverting the call site left all twelve green. test_quality_ladder_applied.py adds seven that do not: two drive _process_video and assert the job carries the configured rungs, so the old literal fails them; five drive the transcoder and assert the -filter_complex those rungs produce — one rung giving split=1 and scale=1280:720 with a single -map and a var_stream_map of v:0, the default ladder giving split=3, the clamp giving scale=640:360 where it used to give 1920:1080, and junk reaching the transcoder still giving a ladder rather than split=0.

Re-measured, twice. The table in the description is from a real transcode() against a real file; row 3 is now 1280x720 where the same run produced 1920x1080 before the clamp.

And on a self-hosted dev instance end to end, since the mocked graph is not the same claim as ffmpeg actually doing it. TRANSCODER_QUALITIES=1080p, a 640x360 upload, and the master manifest that came back out of the bucket:

#EXT-X-STREAM-INF:BANDWIDTH=176719,RESOLUTION=640x360,CODECS="avc1.64001e,mp4a.40.2"
0/playlist.m3u8
One rendition, at the source's size. The same instance with a 1920x1080 source produced one rendition directory rather than three, which is the setting doing its job.

The closing keyword is out of the description, so #319 stays open for the passthrough half.

On that half: your list is the useful part of that thread — H.264 in a supported profile and level, yuv420p, a reachable moov atom, an audio stream the segmenter can cut, and the keyframe cadence from your earlier comment, which is the one I had missed entirely. I will pick it up as a branch inside transcode() gated on a config value, sequenced after #314 so it can call _upload_directory, and with the ffmpeg install step in the workflow so the framemd5 test actually runs. Not before this one is settled, so the two do not tangle.

API suite: 421.

ravirajsinh45 (approved):
@Lennart-Pingpong Both blocking items are fixed, and the non-blocking ask with them. I checked this one by running the code on both branches rather than reading it, so the evidence below is measured. Approving.

## Blocking, both resolved

**1. The fallback no longer upscales.** You took the clamp route rather than restoring the rung. `packages/transcoder/ffmpeg_transcoder.py:601` builds a per-rung target map, `scale_targets = {q: QUALITY_MAP[q][0] for q in requested}`, the fallback overrides only the kept rung at `:616` with `scale_targets[smallest] = f"{source_width}:{source_height}"`, and `:696` reads that map instead of the nominal size. That is the right shape: the rung name stays the rung name, and only what it scales to changes. I ran the real path with ffprobe stubbed on this branch and on `main`:

| source | `TRANSCODER_QUALITIES` | main | this branch |
|---|---|---|---|
| 640x360 | `1080p` | `scale=1920:1080` | `scale=640:360` |
| 426x240 | default | `scale=640:360` | `scale=426:240` |
| 1920x1080 | default | three rungs | three rungs, unchanged |

Row 1 is the exact case we cited, and the 1920:1080 target is gone. Rungs at or below the source are untouched. The three places that documented the old behaviour now match the code: `.env.example:141-142`, `apps/api/config.py:65-66`, and `CHANGELOG.md:11`.

**2. The setting is now pinned to a real transcode.** `apps/api/tests/test_quality_ladder_applied.py` covers both halves, the call site and the filter graph, and it survives the two mutations we named. Restoring the hardcoded `["1080p", "720p", "360p"]` at `apps/api/tasks/transcode_tasks.py:147` fails `test_the_job_carries_the_configured_ladder_and_not_a_literal`. Reverting `:696` back to `{QUALITY_MAP[q][0]}` fails `test_a_rung_above_the_source_is_built_at_the_source_size` and `test_source_smaller_than_all_requested_keeps_smallest`. Nothing else in the suite notices either mutation, so these are the only guards and they hold. Two details worth calling out: `_ffmpeg_cmd_for` asserts `calls` is non-empty before indexing, so the helper cannot decay into a vacuous pass if the graph stops being built, and you flipped the old test at `test_ffmpeg_transcoder.py:244` rather than deleting it, so the case that used to assert the upscale now asserts the clamp.

**The non-blocking ask is done too.** The body opens "Part 1 of #319" with no closing keyword anywhere, and #319 is still open carrying the passthrough half. Your paragraph separating the clamp from #319's `source` rung is the correct reading: this still scales, to the size it measured. Nothing was expected of you on #328; that ask sits on the issue.

## Non-blocking

None of this holds the merge. Item 1 is the only one I would like an answer on, either as a follow-up commit or a "deliberate" in a reply.

**The clamp replaces the rung's box wholesale, which changes default-config behaviour for wide and short sources.** `:616` swaps in the source dimensions instead of capping each dimension against the rung. The fallback fires whenever every requested rung is taller than the source, and nothing there says the source is also *narrower*. Measured, default ladder, no configuration set: a 1920x240 ticker gets `scale=640:360` on `main`, which ffmpeg resolves to 640x80, and `scale=1920:240` on this branch, so nine times the pixels on an instance that changed nothing. `f"{min(_rung_width(q), source_width)}:{min(_rung_height(q), source_height)}"` fixes it and leaves both new tests green; the 426x240 case still yields 426:240 and the 640x360 case still yields 640:360. If you would rather keep the current reading, that is arguable since it preserves source detail, but then the `CHANGELOG` entry should say so, because it currently promises only that nothing upscales.

**The clamped rung keeps its nominal CRF.** `:710` still does `_, crf = QUALITY_MAP[quality]`, so size now comes from the source while rate control comes from a rung that no longer describes the output. On the motivating example, `TRANSCODER_QUALITIES=1080p` on a 640x360 upload, I confirmed the generated command carries `-crf 24` rather than the 360p rung's `-crf 30`. Six CRF points is roughly a doubling of bitrate, so an operator setting a one-rung ladder to save storage gets a file about twice the size of the 360p rendition the default ladder would have produced. Defensible, since it is the only rendition, but deriving the CRF from the clamped height would keep the setting's cost promise intact. Related and purely cosmetic: the variant directory is still named for the nominal rung, so that 640x360 rendition lands in `1080p/`. Nothing downstream parses those names and ffmpeg writes the real `RESOLUTION` into `master.m3u8`, so it only shows to someone reading the bucket.

**Two nits.** `CHANGELOG.md:19` says the old output was a "padded 1920×1080 rendition", but there is no `pad` filter anywhere in `ffmpeg_transcoder.py`; `force_original_aspect_ratio=decrease` rescales and never adds bars, and for the 640x360 example the aspect matches exactly, so it was a plain upscale. In fairness the word is inherited from our own earlier #204 entry at `CHANGELOG.md:168`, so if you drop it here we will fix that one too. And `TRANSCODER_QUALITIES` is missing from the env-var table at `docs/deployment.md:275`, which already carries `TRANSCODER_PIPELINE`, `TRANSCODER_OUTPUT` and `TRANSCODER_HDR`; that table is the only place in the repo where a self-hoster tuning transcode cost would look.

Happy to merge as is, or after the `min()` one-liner if you want it in the same PR. Thanks for the measured table in the description, re-running row 3 after the second review is exactly the thing that made this quick to verify.

Issue #321, opened 2026-09-04
Comment export is offered to share-link guests but the endpoint has no guest auth path


ravirajsinh45:
### Describe the bug

The comment export control is rendered on the public share review screen, but the endpoint behind it requires an authenticated user and there is no share-token path. For a share-link guest the button is fully live, offers every format, and always fails.

Unlike #316, this is not a case of a control that quietly does nothing because a store is empty. `share-review-screen.tsx:61` mounts `ReviewProvider` with the `assetId` and `shareToken`, so `currentAsset` and `currentVersion` really are populated. That means:

- the three NLE entries render for a video asset, because they are gated on `currentAsset?.asset_type === "video"`
- `handleExport` passes its `if (!currentAsset || !versionId) return` guard
- the request actually goes out

`apps/web/lib/export-comments.ts:32` sends only:

```ts
headers: { Authorization: `Bearer ${getAccessToken()}` }
```

and `getAccessToken()` (`apps/web/lib/auth.ts:8-11`) reads `localStorage` and returns `null` for a guest, so the header goes out as `Bearer null`.

`export_comments` (`apps/api/routers/comments.py:706-716`) takes `current_user: User = Depends(get_current_user)` and then calls `require_asset_access(db, asset, current_user)`. There is no share-token variant of this route, so the answer is always 401.

Today the 401 lands in `console.error` and the click does nothing visible. #317 changes that: it renders export failures in the panel, so once it merges a guest gets a bare `Export failed (401)` next to the button. That is an improvement everywhere else and a regression here, which is why this is worth fixing alongside it rather than after.

Worth noting this got more reachable with #307, which put share-link viewers on the same review stack as everyone else.

### Steps to reproduce

1. Create a share link for a video asset that has timecoded comments. Do not require a password.
2. Open it in a browser with no FreeFrame session (a private window).
3. Switch to the Comments tab and click the export icon.
4. All four formats are listed. Pick any of them.
5. Nothing happens. The network tab shows a 401 on `/assets/{id}/comments/export`.

### Expected behavior

Either of these is consistent. The first is much smaller.

- **Do not offer the control where it cannot work.** Pass an explicit `canExport` prop (or gate on the presence of an access token) so the export button does not render for a share-link guest.
[...]

PR #326, opened 2026-09-05
fix(comments): export the asset the panel is showing, not the one in the store

Lennart-Pingpong:
## Summary

I filed this as "the export offers only CSV and clicking it does nothing". That
is the cold-load case, and it is the harmless one. You are right that
`currentAsset` is only null until someone opens an asset: nothing clears the
review store, `review-provider.tsx` registers no unmount cleanup, and the back
arrow is a client-side navigation. So after opening any asset and returning to
the grid, the store still holds it, the guard passes, the NLE entries render
from the stale asset's type, and the export **succeeds** — a 200 download of the
previously opened asset's comments, named after it.

Doing nothing was the good case.

## Changes

Option 1, in the shape you described. `CommentPanel` takes the asset to export
as a prop, next to the `exportVersionId` that already exists for compare mode,
and the project page passes what it has selected.

The `undefined` / `null` distinction is what the fix rests on: `undefined` means
the caller has no opinion, so the store is used and the review screen is
untouched; an explicit `null` means the caller has one and the answer is
nothing. Collapsing those two puts the stale asset straight back.

The missing-asset case is folded into #317's `exportError` line rather than
returning ahead of it, as you asked — a guard that returns silently is the shape
of bug this panel just had.

On which version "this version" means: the project page exports
`latest_version`, which is what the grid shows and what the comments beside it
were loaded for; the review screen keeps exporting whatever its switcher has
selected. That is a comment at the call site rather than something to infer.

## Two things deliberately left out

Both real, both confirmed, neither in this PR:

- The project page passes `onReply` and `onSubmitReply` as stubs, so the inline
  reply box opens and Send resolves without posting. `useComments` already
  returns `createComment` there, so it is a small fix — but a reply that
  silently discards is its own bug with its own tests, and bundling it into an
  export change makes both harder to review.
- The timecode click and the annotation jump write the global store with no
  player mounted to consume them.

Happy to take either as a follow-up, or to fold them in here if you would rather
have one pass over the call site while it is open.

## Testing

- [x] Frontend tests + build pass — 377 passed, 5 new in
      `comment-panel-export-asset.test.tsx`, `tsc --noEmit` clean.

- [x] Backend tests pass — untouched by this change.

- [x] Tested manually in browser, on our own instance, against two seeded assets
      with one timecoded comment each:

  - **From the project page, with the review store provably empty** — done in
    its own tab, with no asset opened in it, so nothing could have written the
    store — selecting an asset and exporting produced that asset's EDL: right
    title, right filename, marker at the right timecode. This is the decisive
    one. With an empty store the old code renders no NLE entries at all and CSV
    returns silently, which is exactly how the bug was reported, so a correct
    EDL can only have come through the new prop.
  - **From the review screen, in a separate tab**, the store-driven path still
    exports correctly. No regression on the path that was never broken.
  - Unplanned and verified for free: a resolved comment came out as
    `ResolveColorGreen` and an open one as blue.

  Not clicked through: the stale-store case — open one asset, return to the
  grid, select another, export. That is the case your review described, and it
  is what the unit test seeds directly, but I would rather say it was covered by
  a test than imply someone watched it happen.

The test that matters is that stale-store one: it puts a different asset in the
store, renders with the prop, and asserts the export ran against the prop's
asset. Without the fix it exports the stale one and passes nothing.

## Checklist


ravirajsinh45 (changes requested):
Thanks @Lennart-Pingpong. Threading the panel's own asset through instead of reading the store is the right fix, and it makes compare mode correct for free.

Note this PR is stacked on #317 (it carries `2075884`), so merging it merges that work too, and the blocker we left there is still open.

## Blocking

**`exportError` is still only cleared by another export attempt.** Every write is inside `handleExport`: `comment-panel.tsx:928`, `:933`, `:956`, with the state declared at `:819`. Export v2 as EDL, get the refusal, switch to v1 with the version switcher, and six lines of red text about v2 sit above v1's fully timecoded comments. `VersionSwitcher` and `CommentPanel` are siblings with no `key`, so nothing remounts. Clearing on `currentAsset?.id`, `currentVersion?.id` and `exportVersionId`, and on opening the menu, covers it.

**`[Unreleased]` still has two `### Fixed` sections**, at `CHANGELOG.md:10` and `:20`. #312 is queued behind this and edits the same block.

## Non-blocking

This closes the export half of #316, but the reply control on the same project-page panel is still broken, so please do not let the merge auto-close the issue.

The guest path in #321 is untouched: a share-link guest now gets a bare `Export failed (401)` where they previously got silence. That one is ours to fix, not yours.

Lennart-Pingpong:
Thanks — both fixed.

**`exportError` is cleared on identity change and on opening the menu.** An effect keyed on `currentAsset?.id`, `currentVersion?.id` and `exportVersionId`, plus a clear when the export button opens the menu, since that is the start of a new attempt.

The deps are the ids and not the objects on purpose: the review store hands out a fresh object on unrelated updates, and clearing on those would take the message off screen before it has been read. There is a test asserting the message *survives* a store write that changes neither id, alongside the three that assert it goes.

`comment-panel-export-error-clearing.test.tsx`, four tests. The three positive ones fail without the fix — checked by removing it.

**The two `### Fixed` sections are one.** `[Unreleased]` is now Changed then Fixed, with the four entries in one block.

Noted on both non-blocking points: the reply control on the project-page panel is still broken, so #316 should stay open — the closing keyword is out of the description. And #321 is yours; happy to test whatever you land there, since a share-link guest hitting that 401 is our own daily path.

Verified by hand on a dev instance: export with nothing to export, get the refusal, switch versions with the switcher, message goes; reopen the menu, message goes.

Web suite: **381**.

ravirajsinh45 (approved):
@Lennart-Pingpong Both blocking items are fixed, and the stale-refusal one is fixed the way we asked rather than approximated. Approving.

## What you changed

**The refusal now clears on the export target, not just on the next attempt.** The effect at `apps/web/components/review/comment-panel.tsx:824-830` keys on `[currentAsset?.id, currentVersion?.id, exportVersionId]`, which covers all three routes into this panel: the review screen (version from the store), compare mode (per-pane `exportVersionId`), and the project page (`exportAsset` + `exportVersionId`). Keying on ids instead of the objects is the right call, and so is guarding the menu clear at `:1230` with `if (!exportOpen)` so closing the menu does not wipe a message someone is still reading.

The new `__tests__/comment-panel-export-error-clearing.test.tsx` pins it rather than describing it. We reverted both halves of the fix and re-ran the file: 3 of the 4 cases fail, one per cleared path. The fourth, the "survives a render that changes neither" case, is a deliberate guard on the id-vs-object dep choice, which is a good thing to have written down.

**The CHANGELOG duplication is gone.** One `### Changed` and one `### Fixed` under `[Unreleased]` (`CHANGELOG.md:10` and `:16`), all three of your bullets folded into the single Fixed list above main's #288 entry, none lost. Against main the delta is a clean three-line insertion, and the merge resolved the #314 collision in main's favour, so the transcoder entry is intact. `origin/main` is an ancestor of the branch and the PR's own diff is only the 10 files you own.

Since #317 was closed in favour of this PR, this branch carries that work too; commit 2075884 came through intact.

## Blocking

None. Both items from the previous review are closed.

## Non-blocking (follow-ups, or fold in here if you prefer)

**1. An in-flight refusal can still land on the target the user moved away from.** The new effect clears an error that is already on screen, but `handleExport` writes the error *after* the await: `await exportComments(...)` at `:946`, `setExportError(...)` at `:965`, with nothing checking that the target is still the same one. If the asset or version changes while the request is out, the effect fires first and clears nothing, then the late rejection paints the old target's refusal over the new one, which is the exact symptom the commit message sets out to kill. Repro on the review screen: start an export, switch version with the switcher while the request is in flight; the panel does not remount, so v1's "None of this version's comments are attached to a timecode" lands above v2's comments. ArrowLeft/ArrowRight asset navigation (`assets/[assetId]/page.tsx:184-199`, a `router.push` with no remount) does it too. The repo already has the idiom: `review-provider.tsx:81/209/232` captures `const reqId = ++commentsReqRef.current` and drops the response when `reqId !== commentsReqRef.current`. Capturing `currentAsset.id` and `versionId` at the top of `handleExport` and skipping `setExportError` when they no longer match would close it.

**2. In compare mode, a right-pane switch wipes the left pane's refusal.** The deps at `:830` include `currentVersion?.id`, but when `exportVersionId` is given it is the only thing that decides what gets exported (`const versionId = exportVersionId ?? currentVersion?.id`, `:938`). The left pane gets `exportVersionId={left.id}` (`compare/compare-overlay.tsx:422`) while the right pane's version *is* the store's: `right = rightVersion` (`:84`), fed from `currentVersion` at `assets/[assetId]/page.tsx:459`, and `handleSwitchRight` calls `setCurrentVersion` (`compare-overlay.tsx:308-326`). So switching the right pane to v3 to compare against fires the left panel's effect and drops a message about v1 that is still true and may not have been read, which is the failure mode the id-based deps were chosen to avoid. Mirroring the export target, `[currentAsset?.id, exportVersionId ?? currentVersion?.id]`, fixes it and keeps all four new tests passing.

**3. Nit: the `if (!exportOpen)` guard is inert for pointer input.** The trigger button sits outside the `Dropdown`'s ref and `useClickOutside` (`:126-139`) listens on `mousedown`, so a mouse press on the button while the menu is open has already run `onClose` by the time the `click` handler reads `exportOpen`. The handler therefore always sees `false` for mouse users and the guard only does anything for keyboard activation. No user-visible consequence, since clearing on either edge is fine; flagging it only because the comment describes a state the handler cannot observe with a pointer. The mousedown/click double-toggle itself is pre-existing and shared with the visibility, filter and sort dropdowns, so not yours.

## On merge

Your PR body carries no closing keyword for #316 and names the dead reply control explicitly, which is exactly right. We will not let #316 close on merge: `projects/[id]/page.tsx:984-985` still passes `onReply={() => {}}` and `onSubmitReply={async () => {}}`, so the reply half is outstanding. We will either take that as the follow-up you offered or keep #316 open with the export half credited; either way it is our call, not something to hold this PR for. The #321 guest 401 is ours and untouched here, as expected.

Nice, focused work. Thanks for the quick turnaround.

PR #327, opened 2026-09-05
feat(web): drop files onto the asset area to upload them

### LINKED ISSUE #315: Let files be dropped onto the asset area, not only inside the upload dialog
### Problem

Uploading a file means clicking Upload, waiting for the dialog, and only then being able to drop it. The drop target exists — UploadZone handles dragenter / dragover / drop and reads dataTransfer.files — but it is rendered inside the "Upload asset" dialog, so it cannot be reached until after the click that it was supposed to save.

Meanwhile the asset area itself, which is the largest thing on the page and the thing people actually drag onto, ignores files entirely. Dropping a video there does what dropping a file on any page that does not handle it does: the browser navigates away from the app and opens the video.

The empty state makes it plainest. It says "Upload your first asset to get started" across a large empty rectangle, and that rectangle is inert.

Image

### Proposed solution

Make the asset area a drop target, and give the files to the same handler the dialog already uses.

On a drag carrying files, show an overlay over the asset region — the same affordance UploadZone shows today, at the size of the area.
On drop, call the existing handleFilesSelected. Nothing about initiate, parts or completion changes; this is only about where the files come from.
Files land in the folder currently open, which startUpload already takes as currentFolderId.
Two details that decide whether this feels right or wrong:

Tell a file drop from an asset drag. The grid is already a drag surface: asset-grid.tsx sets application/json with effectAllowed = 'move' so assets can be moved into folders. A drop handler that does not distinguish them would start an upload when someone drags a card across the grid. dataTransfer.types separates them cleanly — an OS file drag carries Files, the internal one does not — so the check is one condition rather than a redesign.

dragleave fires when the pointer crosses a child. Moving over a card inside the region raises dragleave for the region, so an overlay driven directly by those two events flickers on every tile the pointer passes. The usual fix is a depth counter incremented on dragenter and decremented on dragleave, with the overlay shown while it is above zero.

A natural extension, but not required here: a file dropped on a folder card or a folder in the tree could upload into that folder, since those already accept drops for moving assets.

### Alternatives considered

Leave it in the dialog. It works, and it is one click. But the click is the entire friction being described, and an affordance nobody can see does not get used: there is nothing on the page suggesting that a drop would be accepted anywhere.

Make the whole window a drop target. Fewer misses, and no need to hit a particular region. It also means the review screen, the folder tree and every dialog inherit it, and a drop meant for one of them would be caught by the page underneath. Scoping it to the asset area keeps the target where the mental model already is — this is the place my files live — and leaves the other surfaces free to handle their own drops later.

Only make the empty state a drop target. Smaller change, and it fixes the screenshot above. It also fixes the case that happens once per project and leaves the one that happens every day.

Lennart-Pingpong:
## Summary

Closes #315. The drop target existed but only inside the "Upload asset" dialog,
so it could not be reached until after the click it was supposed to save — and
dropping a file on the page itself did what any page that ignores files does:
the browser navigated away to display the video.

The product question you asked is settled in the issue: **the drop starts the
upload, no dialog.** A drop has already made every decision the dialog collects
— which file, which folder, and the name comes from the file — so confirming it
again is the friction being removed. The single-file rename field is skipped;
renaming from the grid covers that. `handleFilesSelected` is not reused, since
on its own it only sets pending state and shows nothing.

**A drop on a folder uploads into that folder**, which is the obvious reading of
the gesture and the second decision you flagged.

## Your three conditions

- **The handler is on the content region, not on AssetGrid.** So the empty state
  accepts a drop — the case the issue opens with, "Upload your first asset to
  get started" written across an inert rectangle — and so does the blank space
  below a short row of cards.
- **Gated on `canUpload && !showTrash && !showShareLinks`.** A reviewer never
  gets an affordance that ends in a 403, and neither of the other two views can
  receive an upload.
- **Folder cards and tree rows are handled in this PR**, not as a follow-up.
  They no longer light up for a drag they cannot serve — instead they serve it.

## Three details that decide whether this feels right

**Telling a file drag from an asset being moved.** The grid is already a drag
surface: `asset-grid.tsx` sets `application/json` on dragstart. Without the
distinction, dragging a card across the grid starts an upload. `carriesFiles`
reads `dataTransfer.types`, the part available during `dragover` — the items
themselves are protected until drop — and an OS drag always lists `Files`. One
predicate, used by the region and by both folder targets.

**Depth counters, not booleans.** `dragenter` and `dragleave` fire for every
child the pointer crosses, so a boolean drops the marking on each card it
passes. The region counts, and so does each folder — a card's own thumbnail
raises the same pair.

**Claiming is immediate, releasing is deferred.** The few pixels of gap between
two folder cards belong to the region, so a pointer moving from one folder to
the next reports "no folder" in between; clearing at once made the big frame
flash on and off in that gap. A folder claiming the drag cancels a pending
release, so the handover reads as one marking moving. Leaving the region and
dropping still clear at once — those are not handovers.

Only one marking is ever lit. While a folder is the target the region gives up
its own frame, because two frames at once do not say where the file will land.

## Two things testing turned up

**The uploads panel made every drop target on the page dead.** Its backdrop is
`fixed inset-0 z-40` with a click handler, so while the panel is open it covers
the viewport and swallows the drag before anything underneath sees it. It now
steps aside on a file drag. That is a change to `uploads-panel.tsx` and strictly
outside this feature — happy to split it out if you would rather have it on its
own, though the two are hard to notice apart.

**A drop is refused unless the pointer is inside the region at release**,
checked against the region's rectangle rather than inferred from the event
arriving. `preventDefault` still runs on the refused path: without it the
browser handles the file itself and navigates away, which is worse than an
unwanted upload.

## Testing

- [x] Frontend tests + build pass — 370 passed, 5 new on the predicate,
      including the internal-move case it exists to reject. `tsc --noEmit`
      clean, `pnpm build` compiles.
- [x] Backend tests pass — untouched by this change.
- [x] Tested manually in browser on our own instance, which is where the last
      three commits come from: empty project, project with cards, blank space
      below a short row, dragging an asset across the grid (no upload, as
      intended), folder cards, folder tree rows, the gap between two folders,
      the uploads panel open, and releasing outside the region.

The unit tests cover the discriminator. The rest of this is drag-and-drop
behaviour across a real pointer, and I would rather say it was clicked through
than imply jsdom proved it.

## Checklist


ravirajsinh45 (changes requested):
Thanks @Lennart-Pingpong. The `carriesFiles` guard in `apps/web/lib/drag.ts` is the right primitive, and checking `dataTransfer.types` rather than `files` is correct for dragover, where the file list is not yet readable. The counter approach for the flicker between adjacent folders is also the right shape.

## Blocking

**The sidebar tree never receives the handlers.** `FolderTree` destructures `onDropFiles` and `onFileDragOverFolder` at `folder-tree.tsx:297-298`, but the top-level `<FolderNode>` map at `:333-344` passes only `onDropItems`. `FolderNode` forwards both to its own children at `:276-277`, so the recursion is right, but nothing seeds it. Every row in the sidebar gets `undefined`, and the feature the CHANGELOG advertises is dead there.

**It is worse than dead.** `handleDragOver` returns early at `:96` when `onDropFiles` is falsy, before `preventDefault()`. The browser's default file-drop handling then fires, so dropping a file on a sidebar folder navigates the tab to `file:///…` and the user loses whatever is unsaved. The same shape is at `folder-card.tsx:180`, which means a `reviewer` or `viewer`, correctly given no upload handler, now gets a navigate-away where the old code silently absorbed the drop. Call `preventDefault()` on any file dragover you intend to refuse, then return.

**The project-root row promises and then swallows.** `folder-tree.tsx:313-327` calls `preventDefault()` unconditionally and lights up `isDragOverRoot`, but `onDrop` only parses `application/json`, so a file drop highlights the row and does nothing at all.

## Non-blocking

With the Upload dialog open, a drop on its backdrop passes the rect test at `page.tsx:725` and starts an upload behind the dialog.

A file-typed drag whose `dataTransfer.files` is empty on drop, which is what several promised-file sources on macOS produce, leaves "Drop to upload" on screen permanently, since the clear only runs on the success path at `folder-card.tsx:193-201`.

`drag.test.ts` covers the predicate only. A test that renders `FolderTree` and asserts a file drop on a row calls `onDropFiles` would have caught the first item, and it is the one behavior worth pinning here.

Lennart-Pingpong:
Thanks — you found a genuinely bad one, and testing the fix on a real instance found the same hazard in a place neither of us had looked. All three blockers and both non-blocking notes are fixed.

Your three
The sidebar tree never received the handlers. The top-level map passes both now, so the recursion is seeded. There is a test that would have caught it: folder-tree-file-drop.test.tsx renders FolderTree and asserts a drop on a row, on a nested row and on the project row reaches onDropFiles. Three of its tests fail on the unfixed tree — checked by putting the old map back.

The navigate-away. You are right that refusing has to be explicit, and that it was worse than the old behaviour for exactly the people who should not be uploading. FolderCard and FolderNode now preventDefault() on a file dragover before deciding, set dropEffect = 'none' when they refuse, and prevent the drop as well — the dragover alone is not enough, since the drop's own default is what navigates.

The project-root row uploads into the project root now, which is the folder it stands for. onDropFiles takes string | null for that; FolderCard still passes its own id, and the row only lights up when there is somewhere for the drop to go.

Your two non-blocking ones
A drop is handed to onDropFiles even when dataTransfer.files is empty. The card stops the event, so the region behind it never sees the drop and its marking comes down through that call and nothing else — skipping it left "Drop to upload" on screen permanently, which is slightly worse than the stuck highlight you described.

The region no longer accepts drops while the upload dialog is open. Its backdrop is portalled from inside the region, so the drop bubbled to it and passed the release check.

Found while testing, and the second one is the same hazard you flagged
Two folders could be lit at once, with the whole-area frame back up beside them, while the pointer crossed between them. Crossing raises the new folder's dragenter before the old folder's dragleave, so the release arriving last belonged to the folder already left behind — and it took the marking off the folder under the pointer, after which the region put its own frame back. The 90 ms deferred release was built for the gap between two cards and does not help, because the claim comes first.

Three changes, each removing one way for the two to disagree: a folder names itself when it releases and the region ignores a release from a folder that no longer holds the marking; the claim is re-asserted on every dragover, not only on entry; and the folder's highlight for a file drag is now driven by the region's target instead of per-card state, so "exactly one frame is lit" is a property of the state rather than of the cards agreeing. Item drags keep their own state and are unaffected.

Dropping a file on the two pixels of gap between two sidebar rows navigated the tab to file:///… and took the session with it. Safari, by hand.

That is your second blocker in the place a per-element guard cannot reach: the gap belongs to no element that knows about files, and neither does the header, the sidebar background, or the space beside the comment panel. Every one of them hands the drop to the browser.

The page is a closed surface for file drags now. Everything that accepts a file stops the event — the asset region, the folder cards, the folder rows, the project row — and the page root refuses whatever still arrives, cancelling the browser's handling and reporting none so the cursor says so. The cursor still says copy over the places that do take a file. Two tests pin the layering: a drop a row has taken does not reach the page, and one nothing claimed does.

I would understand if you want that last one as its own PR, since it is a page-wide guard rather than part of the feature. It is one commit and lifts out cleanly.

Verification
Web suite 432, tsc clean. Every fix checked by removing it and watching the test go red.

Tested by hand in Safari on a dev instance: drops on folder cards, on sidebar rows, on nested rows, on the project row, on the empty state, on the dialog backdrop, releasing outside the window, and crossing between folders in both the grid and the sidebar.

ravirajsinh45 (changes requested):
@Lennart-Pingpong this round lands all three blockers, and the approach on the refusal one is better than what we asked for. Thanks for the thorough pass.

**What you fixed**

- **Sidebar threading.** The top-level `<FolderNode>` map now forwards the full file-drop set (`folder-tree.tsx:389-392`: `onDropFiles`, `onFileDragOverFolder`, `fileDragTarget`), matching the recursive child map at `:296-299`, with the page supplying them at `page.tsx:594-596`. Pinned by `folder-tree-file-drop.test.tsx`, and those tests are real: reverting `folder-tree.tsx` to `f508b466` turns 9 of the 11 red.
- **Refusal before `preventDefault`.** Both files now prevent first and refuse second (`folder-tree.tsx:114-131` and `:149-153`, `folder-card.tsx:182-191` and `:208-211`). We only asked for the order swap; you also added a page-wide net, `refuseFileDrag` at `page.tsx:438-442` wired to the outermost element, with every accepting target calling `stopPropagation` (`page.tsx:427`, `:475`; `folder-tree.tsx:126`, `:350`; `folder-card.tsx:194`, `:213`) so it only fires on a drag nobody claimed. That closes the case we never named, the couple of pixels of gap between two sidebar rows. We checked the obvious risk in that design, a page-level `dropEffect = 'none'` stealing a drop from an accepting child, and the dialog's own `UploadZone` stops propagation at `upload-zone.tsx:41,46`, so it survives.
- **Project-root row.** `folder-tree.tsx:337-357` now branches on the drag type and lights only when there is somewhere for the file to go, and `:359-372` routes files to `onDropFiles?.(null, …)` with the page widened to accept the root at `page.tsx:493-497`.
- **All three non-blocking notes**, including the backdrop one solved differently and better: `canDropFiles = canUpload && !showTrash && !showShareLinks && !uploadOpen` (`page.tsx:408-409`) switches the region off rather than tightening the rect test, so the backdrop drop falls through to the page-level refusal instead of uploading. The empty-`FileList` case now hands the list to the caller unconditionally (`folder-card.tsx:219`, `folder-tree.tsx:159`) and the caller clears the marking at `page.tsx:470-471` before the length bail.

The release-ownership guard is the part we looked hardest at and it holds. `setFileDragFolderId((cur) => (from && cur !== from ? cur : null))` (`page.tsx:382-391`) is correct for the dragenter-before-dragleave ordering the comment describes; we walked A to B, A to gap to B, A to B and back inside the window, drop on card, drop on row, drop on root, and drag leaves window, and every path ends with exactly one owner or a clean null. Re-asserting the claim from `handleDragOver` (`folder-card.tsx:195`, `folder-tree.tsx:127`) is genuine self-healing, not belt and braces. Suite is green at 50 files / 381 tests and `tsc --noEmit` is clean.

## Blocking

**1. `!uploadOpen` closes one modal; the folder rename and delete dialogs are portalled from inside the region too** (`apps/web/app/(dashboard)/projects/[id]/page.tsx:408-409`)

The comment above `canDropFiles` names the mechanism exactly: a Radix dialog rendered inside the region portals to `document.body` but still bubbles through the React tree into the region's handlers. That mechanism is not specific to the upload dialog. `FolderCard` renders its own `NameDialog` at `folder-card.tsx:312` and `ConfirmDialog` at `:324`, and `FolderCard` lives in `AssetGrid`, which is inside the `dropRegion` div at `page.tsx:764-771`. Both dialogs use `Dialog.Portal` with `Dialog.Overlay className="fixed inset-0 z-50 …"` and pointer events enabled, so they cover the whole region and pass every drag event back to it.

Concretely: as an owner or editor, open the ⋯ menu on a folder card, choose Delete, and drag a `.mov` over the "Delete "Cuts"?" confirmation. `uploadOpen` is false, so `handleFileDragOver` claims the event and sets `dropEffect = 'copy'`, and on release the `released` rect check passes because the dialog is centred inside the region, so `startUpload` runs into `currentFolderId` while the user is still being asked whether to delete a folder. Same with Rename.

The dialogs you did not have to guard (`ShareCreateDialog`, `ProjectMembersDialog`, the asset rename and delete dialogs) are safe only because they happen to be rendered at the page root, outside the region, which makes the guard positional rather than principled. Either track "any modal is open" in the condition, or hoist those two `FolderCard` dialogs to the page root. Both close the whole class.

**2. The closed surface stops at the page component: the header, the sidebar rail and the Powered-by badge still navigate the tab to `file:///`** (`apps/web/app/(dashboard)/projects/[id]/page.tsx:541-543`)

`refuseFileDrag` is on the page's outermost div, and the commit message says "Nothing on this page can reach the browser any more". The page's root div is not the page the user sees. In `apps/web/app/(dashboard)/layout.tsx` the project page is `{children}` inside `<div className="relative flex-1 overflow-y-auto">` at `:62`, which is a sibling of `<Header/>` at `:60` (the sticky `h-11` breadcrumb bar spanning the full content width) and a child of `<main>`, itself a sibling of the global `<Sidebar/>` at `:48` (52px collapsed, 220px expanded, full height) and of `<PoweredByBadge className="fixed bottom-4 right-4 z-20 …"/>` at `:69`. `git grep` over `apps/web` confirms only four files have any drag handler at all, and none of Sidebar, Header or PoweredByBadge is one of them, so `dragover` is never cancelled there and the browser does its default file handling.

The badge is the sharp case. It is an `<a href>` pill with default pointer events floating over the bottom-right of the exact asset area the CHANGELOG advertises ("the blank space below a short row of cards"), and it renders on any deployment that has not turned the credit off. Drag a `.mov` toward that empty space, release a few pixels off onto the pill, and the tab goes to `file:///Users/…/cut-v3.mov` with the session. Same on the breadcrumb bar and anywhere on the left rail.

This is not a regression you introduced, and we are not asking you to own the whole layout. But the CHANGELOG sentence "Anywhere on the page that does not take a file now refuses it explicitly" is not true as written, and it is the same failure at a much larger target than the row gap that motivated the fix. The honest home for `refuseFileDrag` is the dashboard layout's root div. If you would rather keep the change inside the project page, that is fine by us; narrow the CHANGELOG sentence to what it actually covers and we will take the layout in a follow-up.

## Non-blocking

**Sidebar rows are switched off by Trash and Share Links** (`page.tsx:594-595`). `onDropFiles={canDropFiles ? handleDropFilesToFolder : undefined}` reuses one flag for the tree and the region. The `!showTrash && !showShareLinks` reasoning is right for the region, whose contents genuinely cannot receive an upload, but a sidebar row is not the region: it always means "upload into this folder" whatever the main pane shows. Click "Recently Deleted", drag a file onto the "Episodes" row, and the row takes the refusal branch at `folder-tree.tsx:117-120` with a no-drop cursor and no explanation. `!uploadOpen` is harmless here since the sidebar is behind the overlay; the trash and share-link half is over-broad.

**The project-root row keeps a local highlight instead of joining the shared marking** (`folder-tree.tsx:334`). The point of lifting the marking to the page (`fileDragTarget`, and the note at `folder-card.tsx:97-100` that "exactly one frame is lit" should be a property of the state) is undone on the one row this PR newly made droppable. The root row still lights from `isDragOverRoot` (`:352`, cleared at `:358`) and never calls `onFileDragOverFolder`, so it is invisible to the release logic: crossing up from a folder row sets it immediately while the folder row only schedules the 90 ms guarded release, and both carry `ring-accent/50` for that window, which is the condition the timer and the `from` guard exist to prevent. Its `onDragLeave` also has no depth counter and no `carriesFiles` check, unlike every other row, so it blinks as the pointer crosses its own icon and label. Reporting through `onFileDragOverFolder` with a root sentinel, or giving the tree a distinguishable root id, would make the invariant hold everywhere.

**Two cosmetic leftovers**, neither introduced here and neither worth a commit on its own: in share mode a file dropped on a folder card falls through to the region and uploads into the currently open folder rather than that card's folder (`asset-grid.tsx:313` passes `undefined` in share mode), and a folder present in both the sidebar and the grid lights in both places at once.

**Tests.** `folder-tree-file-drop.test.tsx` is a good addition and it covers the right things: drop routing, the root row, empty lists, refusal, item drags, which row is lit, and what bubbles to the page. `FolderCard` and the page-level region still have no direct test; the tree tests cover the analogous shapes, so this is a note rather than a request.

Nice catch on the uploads-panel backdrop stepping aside on a file dragenter (`uploads-panel.tsx:250-252`); we had not asked for that one. Fix the two blockers above and this is ready.

ravirajsinh45 (approved):
@Lennart-Pingpong both blockers are fixed, and the first one is fixed better than what we asked for. Approving.

**The portalled-dialog problem is solved generically.** We suggested either tracking "any modal is open" or hoisting the two `FolderCard` dialogs. You did neither and found the better answer: `pointerIsOver` is `e.currentTarget.contains(e.target)`, and a Radix dialog portals its DOM to `document.body` while keeping its place in the React tree, so the overlay is not a DOM descendant of the region even though the synthetic event still bubbles there. One line, and it closes the whole class rather than the two dialogs we happened to name. It also means the next dialog someone adds inside the region is covered without anyone remembering to add it to a list.

**The refusal moved to the shell.** `layout.tsx:57-58` covers the header, the sidebar rail and the attribution badge, which were the three surfaces we flagged as still navigating the tab away.

**`useFileDropRegion` is the right extraction.** Your reasoning in the docblock is the part that matters: a rule spread across four handlers and a timer, inline in an 1100-line page, could not be tested at all. The release-ownership guard and the 90ms handover are the subtle parts and they now have names and tests.

**The tests hold.** Both mutations turn them red: making `pointerIsOver` return `true` fails 3, removing the shell refusal fails 4. The suite is 459 passing on your branch.

**The CHANGELOG now says what actually shipped**, including that the refusal sits on the shell rather than the page, that the badge was one of the ways to lose the page, and that a share link is a separate surface and not covered. That last clause is the one we would have asked for.

One thing for later, not this PR. A share link a guest opens does still navigate the tab away on a stray file drop, as your own CHANGELOG line now says. That is a guest-facing surface, so it is worth its own issue; we are filing it rather than growing this.

PR #339, opened 2026-09-10
fix(web): opening an asset needs two taps on a touchscreen

CamReport:
## Summary

Opening a video from the grid, list view, or a share-link folder browser needed two taps on a touchscreen — a click selects an asset, a double-click opens it, and there's no discoverable double-tap gesture on touch. This wasn't a browser quirk; it's the intended desktop file-manager convention with no touch equivalent.

## Changes

Two independent, compounding fixes, applied identically across all three surfaces (`asset-grid.tsx`'s grid and list views, `folder-share-viewer.tsx`'s `AssetGridCard` and its own list row):

1. **Click routing.** A one-line `isCoarsePointer()` helper (`window.matchMedia('(pointer: coarse)')`) — a single tap on a coarse (touch) pointer now opens the asset directly instead of just selecting it; a mouse keeps the exact original select-then-double-click behavior. `pointer: coarse` over a viewport-width check, since this is about input capability, not screen size. The share-link list row's version also respects the existing `openInViewer` appearance flag, matching its own `onDoubleClick`'s existing guard.
2. **Hover-simulation gating.** Fixing (1) alone wasn't enough: each asset card carries `group-hover:`/`hover:` styling (thumbnail zoom, a selection checkbox, a "⋯" menu or download-button overlay, `SubfolderCard`'s own hover style), and iOS Safari treats a tap on anything with a `:hover`-driven ancestor style as "entering hover" — the click event itself doesn't fire until a second tap, regardless of what the handler would do. This is a genuinely separate mechanism from (1); both are needed together. Every such class is now gated behind `[@media(hover:hover)]:`, so none of these rules exist at all on a device that can't truly hover. Three elements were `opacity-0` until hover (the "⋯" menu, two download-button overlays) — gating alone would make them permanently invisible and unreachable on touch, so each defaults to visible instead, with the hide/reveal-on-hover behavior moved behind the same media gate. Desktop is pixel-identical either way.

Deliberately not touched: every other `hover:` usage elsewhere in `folder-share-viewer.tsx` (comment replies, share menu items, breadcrumbs, etc.) — a different interaction surface, out of scope here.

## Testing

- [x] Frontend tests (`pnpm test`) — 5 new tests covering the click-routing regression on `AssetGrid` (grid and list view: coarse pointer opens directly, fine pointer only selects, double-click still opens either way), all passing; no new failures elsewhere (same pre-existing failures as clean `main`)
- [x] Tested manually on a real phone (single tap opens immediately in all three surfaces) and on desktop (single-click still selects, double-click still opens, hover effects — thumbnail zoom, checkbox glyph, "⋯"/download visibility — unchanged)

**Known test-coverage gap, noted rather than hidden:** the identical logic in `folder-share-viewer.tsx` isn't covered by an automated test. `AssetGridCard` and the list row aren't exported independently of the full `FolderShareViewer` component, which fetches its own data on mount — covered by type-check, a clean production build, and manual verification instead.

## Checklist

## Screenshots

No visual change on desktop (hover behavior identical to before). Happy to add a screen recording of the touch behavior if useful.


ravirajsinh45 (changes requested):
@CamReport this is a good catch, and the second half is the part most people would have missed. Diagnosing that iOS swallows the first click on anything with a `:hover`-driven ancestor, separately from the click-routing question, is the difference between this working and half-working.

**What we checked.** `isCoarsePointer()` is evaluated at click time rather than at render, so an iPad that gains a trackpad mid-session behaves correctly without a re-render. `pointer: coarse` over `any-pointer: coarse` is also the right call: a touchscreen laptop should keep mouse semantics, and `any-pointer` would have changed them. The `opacity-100` base with `[@media(hover:hover)]:opacity-0` plus the gated reveal is exactly the shape we used for the scrubber thumb in #323, so the two are consistent. Your five tests are load-bearing: reverting the routing turns two of them red.

## Blocking, one thing

**The selection checkbox is now permanently invisible on touch.** `asset-card.tsx:135` keeps `text-transparent` as its base and gates only the reveal:

```
'bg-black/40 text-transparent [@media(hover:hover)]:group-hover:text-white/60 backdrop-blur-sm'
```

On a phone `[@media(hover:hover)]` never matches, so the tick never appears. Before this PR it did appear, via the same hover simulation the rest of the PR is removing. The button is still there as a dark square and still tappable, so it is not unreachable, but the affordance is now unlabelled.

It matters more here than it looks: once a single tap opens the asset, this checkbox is the only way to select on touch. It is also the one element where you used `text-transparent` rather than `opacity-0`, so the fix you already applied three times does not cover it. The same shape works: a visible base, with `[@media(hover:hover)]:text-transparent` and the gated reveal after it.

## Not yours, noted so nobody re-reports it

`folder-share-viewer.tsx:295`, `:422` and `:541` have the same invisible-on-touch shape and are untouched by this PR. We will pick those up.

## One correction, third time it has come up

The PR body says "same pre-existing failures as clean `main`". There are none: we ran the suite on your branch and it is 396 passed, 0 failed. CI runs `pnpm test` inside the Frontend Build job and it is green here too. Whatever you are seeing is local to your machine, and it is worth tracking down, because it means you cannot currently tell your own regressions from noise.

The honesty about the `folder-share-viewer` coverage gap is appreciated, and we agree with the reasoning: that component fetches on mount and its cards are not independently exported, so a test there would cost more than it is worth right now.

CamReport:
Thanks for the careful read.

**Checkbox fix pushed (65e11d4).** You're right — `text-transparent` was the unconditional base with only the reveal gated, the one control in this PR that didn't follow its own pattern. Now `text-white/60` is the base, hidden behind `[@media(hover:hover)]:text-transparent` and revealed the same way as everything else here. Confirmed visually that it's back to always-visible on a coarse pointer.

**The test claim — found it, and it's not upstream's problem.** This machine's default Homebrew `node` is v26.8.1; your CI (and presumably your own machine) runs the Node 20 this repo actually targets. Under 26.8.1, jsdom's `window.localStorage` comes back `undefined` in this test environment, which is exactly the `Cannot read properties of undefined (reading 'getItem')` I'd been attributing to a "pre-existing gap." Installed Node 20 alongside it and reran: 396 passed, 0 failed — matches your numbers exactly, same test file count. So there's no real regression I was masking, but there was a real gap in how I was verifying: I was trusting a stale, unverified assumption instead of checking why the failures were happening. Fixed on my end for future PRs, and sorry this took three rounds to run down.

Typecheck, full suite (on Node 20 now), and `next build` all clean on the pushed commit.

ravirajsinh45 (approved):
@CamReport fixed, and the Node diagnosis is the more valuable half of this. Approving.

**The checkbox.** `text-white/60` as the base with `[@media(hover:hover)]:text-transparent` before the gated reveal is exactly the shape the other three use, so all four controls in this PR now follow one rule. Desktop is unchanged: the base is overridden inside the hover media query, and the `group-hover` reveal wins on hover by source order.

Worth saying that no test covers it, and that is fine. jsdom evaluates no media queries and has no layout, so the only assertion available would be on a class string, which restates the implementation rather than testing it. We mutated the fix back to the broken form and the suite stayed green, which confirms that rather than being a complaint. The click-routing tests you did write are the ones that could be meaningful, and they are.

**On the Node version.** That is a genuinely good catch and it is worth more than this PR. Homebrew's default `node` being 26.8.1 while the repo targets 20, and jsdom's `window.localStorage` coming back `undefined` there, explains a "pre-existing failure" that was never real. It means you now have a suite that tells you about your own regressions instead of noise, which is the thing that actually compounds.

It also suggests we are not doing enough to prevent it: there is no `.nvmrc` or `engines` field pinning Node, so anyone arriving with a newer default hits the same wall and reasonably concludes the suite is flaky. We will file that.

Three surfaces, two mechanisms, and the second one was the part that made it work. Thanks for the report and the fix.

Issue #341, opened 2026-09-10
Mobile review screen: bottom comment sheet, compact transport, and safe-area handling


ravirajsinh45:
Umbrella for bringing the review screen's mobile behaviour up to standard. #338 fixed the sizing bugs @Paracid reported (stage sized from the media's own ratio in portrait, `dvh` roots, panel beside the video in landscape). This tracks the surfaces built on top of that.

Splitting it because the pieces have different risk and one of them cannot land on its own.

### A. Bottom comment sheet on mobile

A sheet pinned to the bottom with a peek band, so comments are reachable without losing the video from view. Replaces the current behaviour below `md`, where the panel is a full-width overlay that covers the player entirely.

New: a `useKeyboardInset` hook (the sheet has to move with the software keyboard) and the sheet component itself. Wired into the project asset page first.

### B. The same sheet on the share screen

`share-review-screen.tsx` is the surface a guest actually sees, and it is the one @Paracid reported against. It needs the same treatment as A, but written rather than adapted: the share path converged onto this component in #307 and its layout does not match the project page's.

### C. Compact transport row below `sm`

The transport row runs out of width on a phone and the controls crowd. Collapses the secondary controls behind an overflow menu below `sm`, leaving play, timecode and fullscreen inline. Touches `video-player.tsx` and `audio-player.tsx`.

### D. `viewport-fit=cover` and the safe-area audit

Tracked in detail in #340. **This one cannot land on its own.** `cover` extends the layout under the notch and home indicator, so it has to arrive together with `env(safe-area-inset-*)` padding on every edge-pinned surface, which is precisely what A, B and C add. Hence last.

It is also the fix for a report we have from a contributor testing on an iPhone 17 Pro, where a bottom-pinned surface sits behind Safari's own URL bar.

### Sequencing

Stacked, in order: A, then B on A, then C on B, then D on C. Each is independently reviewable; only D depends on the others being in place to be safe.

### Verify on a device

Not a simulator. Safari's bottom URL bar behaviour and the software-keyboard inset are both part of what this interacts with, and neither reproduces faithfully.

ravirajsinh45:
Found while testing A and B on an iPhone 16 Pro Max. Adding it to **C**'s scope, since C is already in `video-player.tsx`.

### Opening the sheet clips the video instead of resizing it

With the sheet past `peek`, the top of the picture goes off-screen while the sheet covers the bottom, so you get a squeezed band rather than the whole frame.

[...]

Issue #340, opened 2026-09-10
No viewport-fit or safe-area-inset handling, so edge-pinned surfaces cannot be inset correctly on iOS


ravirajsinh45:
### Problem

`apps/web` sets no `viewport-fit`, and uses no `env(safe-area-inset-*)` anywhere. Verified:

```
grep -rn "viewport-fit\|safe-area-inset\|env(safe" apps/web --include='*.tsx' --include='*.ts' --include='*.css'
# no matches
```

The viewport export (`apps/web/app/layout.tsx:21`) is:

```ts
export const viewport: Viewport = {
  width: "device-width",
  initialScale: 1,
  themeColor: "#0A0A0B",
};
```

Two things follow on a notched or Dynamic Island iPhone:

1. Any element pinned to an edge sits inside the safe area by default, which is correct but means the review screen can never use the full display; and
2. If we ever add `env(safe-area-inset-*)` padding, it will silently do nothing, because those values resolve to `0` unless the viewport is `viewport-fit=cover`. That is the trap: the padding looks present in the source and is inert at runtime.

This became visible while investigating a report against a build that *does* use `env(safe-area-inset-bottom)` on a bottom-pinned surface, where the padding had no effect for exactly this reason.

### Why it matters now

#338 made the review layout usable on a phone (`h-dvh`, media-ratio stage sizing, landscape side-by-side). Edge-pinned surfaces are the next thing that gets built on mobile, and this is the setting that decides whether their insets work.

### Proposed fix

- Add `viewportFit: "cover"` to the `Viewport` export.
- Audit every fixed or edge-pinned surface and add the matching `env(safe-area-inset-*)` padding: the video transport row, the comment composer, and anything `fixed bottom-0` / `fixed inset-x-0`. Without the padding, `cover` alone would push content *under* the home indicator, which is worse than today.

Those two steps go together; landing `cover` on its own is a regression.

### Notes

Should be verified on a real device rather than a simulator, since Safari's bottom URL bar behaviour is part of what this interacts with.

Issue #331, opened 2026-09-09
iPhone Safari: video review layout breaks in portrait and landscape


Paracid:
### Describe the bug

The video review layout is not usable on iPhone Safari.

In portrait mode, the video itself is displayed correctly, but the player/container becomes much too tall. This creates very large black areas above and below the actual 16:9 video, pushing the timeline and review controls far down the screen.

In landscape mode the problem is worse. The video takes up most of the available viewport, the lower review interface is not visible, and the page cannot be scrolled down to reach it.

### Steps to reproduce

1. Open a video review page in Safari on an iPhone.
2. View the page in portrait orientation.
3. Notice that the video player container is much taller than the actual 16:9 video, creating large black areas above and below the image.
4. Rotate the iPhone to landscape orientation.
5. Notice that the layout no longer fits the available viewport height.
6. Try to scroll down to reach the review controls below the video.
7. The page cannot be scrolled far enough, so part of the review interface becomes inaccessible.

### Expected behavior

The mobile layout should adapt to the available screen size.

In portrait mode, the video player should match the video aspect ratio without large unused black areas, and the timeline/review controls should appear directly below the video.

In landscape mode, the page should remain usable and allow scrolling to all review controls, or switch to a compact/fullscreen-friendly layout.

### Screenshots

<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/ebd71197-a6ae-45e3-aaa7-ad66fb3e94f1" />

<img width="1311" height="603" alt="Image" src="https://github.com/user-attachments/assets/ce3ce1f6-72f4-4571-b84a-366cd22741c7" />

### Area

Frontend (Next.js)

### Version / Commit

Self-hosted dev install from latest main, tested 2026-09-09

ravirajsinh45:
@Paracid thanks, this is an excellent first report. Describing the portrait and landscape symptoms separately is what made it quick to confirm, because they look like one bug and are actually two.

Both reproduce from the code, no device needed.

**Portrait bands.** The video stage is `flex-1` (`video-player.tsx:320`) with the video `object-contain` inside it, so the stage takes whatever height the column has left over, with no relationship to the video's aspect ratio. At 390px wide a 16:9 video is 219px tall inside a roughly 620px black box. Neither that file nor the review page has a single responsive breakpoint.

[...]

PR #342, opened 2026-09-10
feat(web): comments become a bottom sheet on a portrait phone

ravirajsinh45:
First of four in #341. Stacked: B, C and D follow on top of this.

## What and why

Below `md` the comment panel was a full-width overlay covering the player entirely, so reading or writing a comment meant losing the thing being reviewed. That is the friction behind #331, and the sizing fixes in #338 do not address it.

It is now a sheet pinned to the bottom with an always-visible 56px peek band, draggable between peek, half and full.

## Decisions worth reviewing

**The sheet overlays the media rather than resizing it.** Resizing would re-letterbox the picture on every frame of a drag. Instead the media column is translated by `--ff-media-shift`, so it stays centred in the space the sheet leaves. Audio is pinned to no shift: a waveform fills its box rather than being letterboxed inside it, so translating it would clip it.

**The mini transport appears only once the sheet covers the real one.** At peek the media row reserves the band with `pb-14`, so the player's own transport stays on screen and tappable. A second play button there would stack on it, each with its own independently updating clock. It keys off the live drag height rather than the committed state, so mid-drag the two never both appear.

**Landscape phones are unchanged.** They keep the side-by-side column from #338. Landscape is short on height, not width, and a sheet would eat what little there is.

**The panel's pieces are extracted** into `panelTabs`, `commentPanelNode`, `commentInputNode` and `fieldsBody`, so the desktop sidebar and the sheet render the same nodes and cannot drift apart. That is most of the diff on the page.

## New hooks

- `useKeyboardInset` — neither iOS Safari nor Android Chrome shrinks the *layout* viewport when the software keyboard opens; both resize only the visual viewport, so a bottom-anchored sheet sits behind the keyboard unless lifted by hand.
- `useMediaQuery`, `useMediaTransport`.

## A note on `env(safe-area-inset-bottom)`

The handle row asks for it, and it is currently **inert**: iOS resolves `env(safe-area-inset-*)` to 0 unless the app sets `viewport-fit=cover`, which it does not. That is #340, and it is deliberately last in the stack because `cover` without padding on every edge-pinned surface is a regression. The padding is written here so the row is correct the moment it lands.

## Verification

- `tsc --noEmit` clean, **410 web tests pass** (up from 391; the sheet ships 19 of its own), build passes, lint warning count unchanged from `main` (18 both sides).
- The sheet's own tests cover the snap states, the drag, tap-versus-drag slop, the flick-to-peek velocity bias, the drawing-mode lock, and the mini transport's appearance rule.

**Not verified in a browser.** The project asset page is behind auth and I did not have a session to drive it with. The sheet is exercised by unit tests only here; PR B puts the same component on the share screen, which is reachable as a guest and can be measured properly. Worth a look on a real device before this ships to anyone.


ravirajsinh45:
**CI gates run locally**, since `ci.yml` triggers on `pull_request: branches: [main]` and a stacked PR gets no checks at all.

Ran the same commands the workflow runs, on this branch:

| Gate | Result |
|---|---|
| Backend Tests (`pytest apps/api/tests/`) | 446 passed, 2 skipped |
| `python -m py_compile apps/api/main.py` | OK |
| `pnpm exec tsc --noEmit` | clean |
| `pnpm test` | 410 passed |
| `pnpm build` | OK |
| `pnpm lint` | 18 warnings, identical to `main` |

Not reproduced: the **Docker Build** job, which builds four images. It is unaffected by a frontend-only change, but that is a reason rather than a check, so flagging it rather than claiming full parity.

PR #343, opened 2026-09-10
feat(share): put the comment sheet on the share screen too

ravirajsinh45:
Second of four in #341. **Stacked on #342** — review that first; this PR's diff is against it, not `main`.

## Why this is separate from A

The share screen is what a guest actually opens, and it is the surface @Paracid reported #331 against. But it is not a port: the share path converged onto `share-review-screen.tsx` in #307, and there is no counterpart to copy the wiring from, so it is written to match A's shape rather than lifted.

## What it does

Same rule as the project review screen: bottom sheet on a portrait phone, side-by-side column on a landscape one. The panel's tabs, list and composer are extracted into `panelTabs`, `commentPanelNode` and `commentInputNode` so the sidebar and the sheet render the same nodes and cannot drift.

One thing that fell out nicely: the sheet's `z-40` was chosen in A to sit below a guest identity prompt. That holds here without changes, since `GuestIdentityPrompt` is `fixed inset-0 z-50`, so it still covers the sheet rather than appearing under it.

## Verification, and its limits

`tsc --noEmit` clean, **413 web tests pass**, `pnpm build` succeeds. All run locally.

**CI does not run on this PR.** `ci.yml` triggers on `pull_request: branches: [main]` only, so a PR based on another branch gets no checks at all. That is a property of the stack, not of this change: everything above #342 will show zero checks until it is retargeted at `main`. The four gates were run by hand instead, and the numbers above are from those runs.

Adds a test for `useMediaQuery`, which had none. It covers the *subscription*, not just the initial read: getting that wrong yields a layout that is correct on load and then ignores a rotation, which is exactly the bug that would survive a screenshot check.

**I could not verify this in a browser.** I tried: the extension's window resize reported success while `innerWidth` stayed at 1512 and `outerWidth` came back `[0, 0]`, so the viewport never dropped below `md` and the sheet never mounted. The one real signal I did get is a negative one, which is worth something: at desktop width the sheet is correctly absent and the sidebar renders.

So the sheet's own behaviour is covered by its unit tests from A, the media-query rule is covered here, and the composed result on a real phone is not verified by me. It should be looked at on a device before this ships.


ravirajsinh45:
**CI gates run locally**, since `ci.yml` triggers on `pull_request: branches: [main]` and a stacked PR gets no checks at all.

Ran the same commands the workflow runs, on this branch:

| Gate | Result |
|---|---|
| Backend Tests (`pytest apps/api/tests/`) | 446 passed, 2 skipped |
| `python -m py_compile apps/api/main.py` | OK |
| `pnpm exec tsc --noEmit` | clean |
| `pnpm test` | 413 passed |
| `pnpm build` | OK |
| `pnpm lint` | 18 warnings, identical to `main` |

Not reproduced: the **Docker Build** job, which builds four images. It is unaffected by a frontend-only change, but that is a reason rather than a check, so flagging it rather than claiming full parity.

PR #348, opened 2026-09-10
feat: optionally require a link in every project description

saskia-irinaa:
### What

Two new settings, both empty by default:

| setting | meaning |
|---|---|
| `REQUIRE_PROJECT_DESCRIPTION_PATTERN` | regex matched against a project's description |
| `REQUIRE_PROJECT_DESCRIPTION_HINT` | what the user is told when it does not match |

When the pattern is empty (the default) nothing changes at all - the check returns immediately, so existing instances see no difference.

### Why

Anything reading projects over the API - an automation, a reporting job, a review bot - has no way back to the brief a project was created from unless somebody wrote it down, and asking after the fact never works.

We run FreeFrame next to an automated video reviewer. It can score craft from the file alone, but to check a cut against the script it needs the brief, which lives in Trello. Today it has to guess from file names. With this setting the link is simply always there.

### Details

- enforced on **update** as well as create - a project that had to carry the link to exist should not be able to drop it on the next edit
- case-insensitive, and the link may sit anywhere in the description
- the error message is the operator's own hint. `"description is invalid"` sends people hunting for a formatting rule; `"Paste the Trello card link"` tells them what to do
- documented in `.env.example`

Example:

```
REQUIRE_PROJECT_DESCRIPTION_PATTERN=trello\.com/c/
REQUIRE_PROJECT_DESCRIPTION_HINT=Paste the Trello card link for this project.
```

### Tests

Added to `apps/api/tests/test_projects.py`: off by default, rejects a missing link, accepts one anywhere in the text, case-insensitive.

### Not included

No frontend change. The New Project dialog already renders the API error, so the hint reaches the user as-is. Happy to add an inline requirement note if you would rather surface it before submit - it would need the setting exposed on a public endpoint, which felt like a bigger change than this needs.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -148,3 +148,9 @@ STUCK_PROCESSING_TIMEOUT_HOURS=6 # Re-dispatch a transcode for versions stuck at
 SOFT_DELETE_RETENTION_DAYS=30  # Hard-delete rows soft-deleted longer than N days (cascade + S3 reclaim); 0 = disabled
 ORPHAN_SWEEP_GRACE_HOURS=0    # S3 orphan sweeper: reclaim raw/+processed/ keys with no DB row; 0 = disabled; >0 = only keys older than N hours
 ORPHAN_SWEEP_DELETE=false     # false = report-only (log only); true = actually delete orphans
+
+# Require a link in every project description (off by default).
+# A regex matched against the description on create and update; the hint is what the user is told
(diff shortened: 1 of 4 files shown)


saskia-irinaa:
Closing — this belongs in Aditor's own fork (videoaditor/freeframe), which is what our instance runs from, not upstream. Reopened there with a second change alongside it. Sorry for the noise.

PR #362, opened 2026-09-11
fix(web): viewer-identity menu unreachable on touch in the share folder browser

### LINKED ISSUE #349: Share folder browser: three controls are invisible on touch, and may swallow the first tap
Three controls in the share folder browser are invisible on a touch device, because their visible state depends on a hover that a phone never delivers.

`apps/web/components/share/folder-share-viewer.tsx`:

- `:295` — `bg-black/40 border-white/30 text-transparent opacity-0 group-hover/folder:opacity-100`
- `:422` — `bg-black/40 border-white/40 text-transparent`
- `:541` — the same shape as `:422`

On a device that can hover, these appear when the pointer is over the card. On a phone the reveal never fires, so the control is either fully transparent or at `opacity-0` for the whole session.

### Why now

#339 fixed this class of problem for the project asset grid: the affected controls take a visible base and move the hide-then-reveal behind `[@media(hover:hover)]:`, so a device that cannot hover simply gets the visible state. These three are in a different component and were out of that PR's scope, correctly.

They are worth doing because the share folder browser is the surface a **guest** sees, and a guest is the person most likely to be on a phone.

### Shape of the fix

The same one #339 used, which is already the house pattern after #323's scrubber thumb:

```
opacity-100 [@media(hover:hover)]:opacity-0 [@media(hover:hover)]:group-hover:opacity-100
```

and for the `text-transparent` ones, a visible colour as the base with `[@media(hover:hover)]:text-transparent` before the gated reveal.

Desktop should come out pixel-identical; the gate only adds behaviour on devices that currently have none.

### Worth checking while in there

Whether any of these three also swallow the first tap. #339 found that iOS treats a tap on an element with a `:hover`-driven ancestor style as "entering hover", so the click does not fire until a second tap. If these sit inside a hover-styled card, they may have the same two-tap problem independently of visibility.

CamReport:
Fixes #349, for the one control it named that's still actually broken.

## What was wrong

The dropdown that lets a signed-in viewer get back to the dashboard or log out (`FolderShareViewer`'s top bar, when `viewerName` is set) was `hidden group-hover:block` — revealed only by hovering the avatar button, with no click handler at all. A phone never delivers hover, so this menu was permanently unreachable on touch: not just invisible, unopenable.

## What's fixed

Replaced the hand-rolled hover reveal with the Radix `DropdownMenu` already used for the same shape elsewhere in this app (`asset-card.tsx`'s "···" menu) — click/tap opens it, outside-click and Escape close it, on every input type. Desktop is visually unchanged; only the trigger mechanism moved from CSS `:hover` to a real click handler with proper focus/dismiss behavior for free.

## On the other two controls #349 named

I couldn't reproduce the exact `text-transparent`/`border-white` snippets quoted in the issue against the current `main` — grepping the file for `transparent` and `border-white` turns up nothing. Checking blame: #339 (mine, merged the morning after #349 was filed) already gated the version badge, the grid card's download-button overlay, and the list row's own overlay behind `[@media(hover:hover)]:`, as part of its own click-routing fix to this same file. My read is #349 was filed against the pre-#339 state of the file and the three call-outs landed as a side effect of that PR, before it merged.

I did a full pass for any remaining un-gated `hover:`/`group-hover:` visibility rule in the file to make sure nothing else was missed — this viewer-identity menu was the only one left.

## Verification

Typecheck and the full suite are clean (`tsc --noEmit`, `vitest run`: 494 passed). No automated coverage added for this one specifically — `FolderShareViewer`'s top bar isn't exercised by any existing test, and mounting the full component means stubbing its folder/asset fetch, which felt like more scaffolding than this fix warrants on its own. Manually verified the click-to-open/outside-click-to-close/Escape-to-close behavior in a browser at both desktop and mobile viewport widths.

### DIFF
--- a/apps/web/components/share/folder-share-viewer.tsx
+++ b/apps/web/components/share/folder-share-viewer.tsx
@@ -1,6 +1,7 @@
 'use client'
 
 import * as React from 'react'
+import * as DropdownMenu from '@radix-ui/react-dropdown-menu'
 import {
   Folder,
   File,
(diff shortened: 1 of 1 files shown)


ravirajsinh45 (approved):
@CamReport Thanks, this is right, and your read of #349 is correct where mine was not.

I checked the claim rather than taking it on trust. `text-transparent` and `border-white` both return zero matches in that file on current main, and every hover-visibility rule left in it is gated behind `[@media(hover:hover)]:` except line 939, which is exactly the dropdown this PR replaces. The timeline backs it up too: #349 was filed 2026-09-10 12:15Z and #339 merged 2026-09-11 06:19Z. So the issue was written against the pre-#339 file, and three of its four call-outs had already landed as part of your own PR. That is my mistake for not re-reading the file before filing it.

The fix itself is the right mechanism. A `group-hover:block` reveal with no click handler is unopenable on touch rather than merely invisible, and moving to the Radix `DropdownMenu` already used by `asset-card.tsx` gets dismiss, focus and Escape behaviour for free. I also confirmed nothing in the share tree does a `.contains()` check that portalling the content would break.

Not blocking, and not your bug: the `data-[state=open]:animate-in`, `fade-in-0` and `zoom-out-95` classes on `DropdownMenu.Content` do nothing. `tailwindcss-animate` is not a dependency and `tailwind.config.ts` has `plugins: []`, so those class names are never emitted. Grepping the served stylesheet gives zero occurrences of `animate-in`, against two for `animate-fade-in`, which the config does define. Twenty files already carry them, including the command palette and the confirm dialog, so this is pre-existing and I have filed it separately as #365. Keeping them here is fine for consistency.

Agreed on skipping a test for this one. Mounting the whole viewer to reach its top bar is more scaffolding than the change earns.

PR #363, opened 2026-09-11
fix(web): a file dropped on a share link navigates the guest away

### LINKED ISSUE #351: A file dropped on a share link still navigates the guest away from the page
Dropping a file anywhere on a share link makes the browser navigate away from the app to display it, losing whatever the guest was doing.

### Why this is worth its own issue

#327 closed this for the dashboard. The refusal sits on the app shell (`apps/web/app/(dashboard)/layout.tsx`), so every surface inside it is covered: the header, the sidebar rail, the attribution badge, and the gaps between cards that were the original report.

A share link does not go through that layout. `apps/web/app/share/[token]/page.tsx` and `components/share/share-review-screen.tsx` are a separate tree with no equivalent, so a guest still loses the page on a stray drop. #327's CHANGELOG entry says so explicitly rather than implying the fix is universal.

It matters more here than it did on the dashboard. A guest has no session to come back to: they reopen the link, re-enter a name and email if the link asks for one, and lose any comment they had typed.

### Shape of the fix

`refuseFileDrag` already exists in `apps/web/lib/drag.ts` and is exactly one import and two handlers:

```tsx
onDragOver={refuseFileDrag}
onDrop={refuseFileDrag}
```

on the share tree's outermost element, the same way the dashboard layout does it.

### Decide separately: should a guest be able to drop a file to upload?

This issue is only about not losing the page. Whether a share link with an upload permission should accept a dropped file is a different question, and a bigger one, since it touches what a share token is allowed to do. Worth not bundling.

### Test

`apps/web/app/(dashboard)/__tests__/layout-file-drag.test.tsx` is the pattern: assert the drag is refused with an explicit `dropEffect` rather than merely swallowed, and that an asset being dragged between folders is still left alone.

CamReport:
Fixes #351.

## What was wrong

#327 refused an unclaimed file drag on the dashboard shell — the header, sidebar, and attribution badge are all siblings of the page rather than parts of it, and anything that reached the shell unclaimed was headed for the browser's own handling, which navigates the tab to `file:///...`. A share link renders through a completely separate tree (`app/share/[token]/page.tsx`) with no equivalent, so a guest dropping a file anywhere still loses the page. It costs more here than on the dashboard: a guest has no session to come back to, and any comment they had typed goes with the page.

## What's fixed

Same fix, same mechanism (`refuseFileDrag` from `lib/drag.ts` on `onDragOver`/`onDrop`), at the equivalent shell for this route: a new `app/share/layout.tsx`, playing the same role `(dashboard)/layout.tsx` plays for the dashboard tree. Folder-based upload is deliberately disabled in share mode (confirmed nothing under `/share` wires a legitimate drop target), so nothing here needs this to stop propagating — it only ever sees drags nothing wanted.

## Test

Added `app/share/__tests__/layout-file-drag.test.tsx`, following `(dashboard)/__tests__/layout-file-drag.test.tsx`'s own pattern: asserts the drag is refused with an explicit `dropEffect` rather than merely swallowed, and that a non-file drag is left alone. Confirmed the test fails against the old (fixed) code — reverted the layout to a plain pass-through and the "is refused" case failed as expected — then restored it.

Typecheck, full suite (`vitest run`: 494 passed), and `next build` all clean.

### DIFF
--- a/apps/web/app/share/__tests__/layout-file-drag.test.tsx
+++ b/apps/web/app/share/__tests__/layout-file-drag.test.tsx
@@ -1,0 +1,60 @@
+/**
+ * The share tree refuses a file drag the same way the dashboard shell does
+ * (see `(dashboard)/__tests__/layout-file-drag.test.tsx`).
+ *
+ * A share link has no drop target anywhere below this — folder-based upload
+ * is disabled in share mode — so a file dragged in from outside the browser
+ * always reaches here unclaimed. Left alone it navigates the tab to
+ * `file:///...`, which costs a guest more than it costs a signed-in user:
+ * there is no session to come back to, and any comment they had typed goes
+ * with the page.
+ */
+import { describe, it, expect } from 'vitest'
+import { render, screen, fireEvent } from '@testing-library/react'
+import ShareLayout from '../layout'
+
+const file = new File(['x'], 'notes.mov', { type: 'video/quicktime' })
+const fileDrag = () => ({ types: ['Files'], files: [file], dropEffect: 'none' })
+
+function renderShell() {
+  return render(
+    <ShareLayout>
+      <div data-testid="page">the share page</div>
+    </ShareLayout>,
+  )
+}
+
+describe('a file dropped on the share tree', () => {
+  it('is refused', () => {
+    renderShell()
+
+    expect(
+      fireEvent.dragOver(screen.getByTestId('page'), { dataTransfer: fileDrag() }),
+    ).toBe(false)
+    expect(
+      fireEvent.drop(screen.getByTestId('page'), { dataTransfer: fileDrag() }),
+    ).toBe(false)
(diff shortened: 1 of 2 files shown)


ravirajsinh45 (approved):
@CamReport Good, and the reasoning about a guest having no session to come back to is the right justification for doing this separately.

I checked the two things that could have gone wrong with wrapping the tree in a new element. The height chain is safe: the share surfaces use `min-h-screen` and `h-dvh`, both viewport-relative, so an intermediate div changes nothing. It would have broken had anything in there used `h-full`. And nothing in the share tree renders through a portal, so there is no sibling that escapes the wrapper the way the dashboard's badge and rail do, which is why the shell needed its handlers on the outermost element and this does not.

The tests mirror the dashboard's, including the non-file drag case.

PR #364, opened 2026-09-11
fix(web): compare view's scrubber has no touch support

### LINKED ISSUE #353: Compare view's scrubber has no touch support: click-only seek, 8px target, mouse-only markers
The compare view's scrubber never got the touch work the main player's did in #323, so on a phone it has the same problems the main scrubber had before that PR, plus one of its own.

Promised as a follow-up in the #323 review; filing it so it does not get lost.

### What is wrong

**The track only seeks on a click.** `compare-scrubber.tsx:198`:

```tsx
onClick={(e) => seekFromEvent(e.clientX)}
```

No drag, so a finger cannot scrub, only tap. And the track is `h-2` (8px) with no widened hit area, against 22px of live target on the main scrubber.

**The comment markers are mouse-only.** `ScrubberCommentMarker` (around `:88-92`) uses `onMouseEnter` / `onMouseLeave` / `onClick`, and its tooltip is driven entirely by that hover state. On a touch device the tooltip is unreachable, and per #339 a tap on an element with hover-driven styling can be swallowed entirely the first time.

### What the main player has, for comparison

`progress-bar.tsx` after #323: pointer events throughout with `setPointerCapture`, an `isPrimary`/`button` guard, `pointercancel` and `onLostPointerCapture` so a drag cannot latch permanently, `touch-none`, and a 22px hit area that stays 22px whether or not the asset has comment markers.

### Shape of the fix

Port that model rather than reinventing it. The pointer handling in `progress-bar.tsx` is the reference, and #323's own tests cover the cases that are easy to get wrong: a second finger, an interrupted gesture, a release that never lands on the track.

Two things specific to this component:

- It has **two** marker rows, above and below the track, so the hit-area geometry has to clear both. #323 solved the equivalent with `pointer-events-none` on the marker row and `pointer-events-auto` on each marker.
- Marker hover needs the `[@media(hover:hover)]:` treatment from #339, or a tap-to-open equivalent, so the tooltip is reachable without a pointer.

### Worth knowing

Compare mode is reachable on a phone today but was never designed for it, so this may be worth scoping alongside a decision about whether compare should be offered below `md` at all.

CamReport:
Fixes #353.

## What was wrong

The compare scrubber never got the touch work the main player's did in #323:

- The track only seeks `onClick`, with no drag — no touch equivalent of the window `mousemove` listener a drag needs, so a finger can tap-seek once but never scrub.
- The track is a plain 8px `h-2`, no widened hit area.
- The two comment-marker rows (above and below the track) use `onMouseEnter`/`onMouseLeave` for their hover preview, unreachable on touch, and the avatar dot's unguarded `hover:scale-110` carries the same risk #339 found on the main grid: a tap on anything with `:hover`-driven styling gets treated as entering hover on iOS, so the click doesn't fire until a second tap.

## What's fixed

Ported the model #323 already built for `progress-bar.tsx` rather than reinventing it: Pointer Events throughout, a captured drag with an `isPrimary`/`button` guard on pointerdown, `pointercancel` and `onLostPointerCapture` so a drag can't latch permanently, `touch-none` on the track, and an invisible absolutely-positioned hit-area extension that doesn't add to layout height. Also gated the marker avatar's `hover:scale-110` behind `[@media(hover:hover)]:`, matching #339's fix for the same class of bug.

One thing specific to this component: it has **two** marker rows (above and below the track) where #323's has one. Applied the same `pointer-events-none`-on-the-row / `pointer-events-auto`-on-the-marker treatment to both, and moved both rows to render after the track in the DOM — the track's hit-area extension reaches into both rows' gutters, and a marker has to be later in paint order than the extension to win hit-testing wherever the two overlap. #323's single row already rendered after its track for the same reason; this one needed it on both sides.

**Left as-is, deliberately:** the hover-driven tooltip preview stays mouse/trackpad-only. A tap already calls `onMarkerClick` (the same action a click does), so a touch user reaches the underlying comment in one tap regardless — there's no unique content in the transient preview to justify a separate tap-triggered version of it. Also didn't touch whether compare mode should be offered below `md` at all — #353 raises that as a separate scoping question and explicitly leaves it open.

## Verification

Typecheck, full suite (`vitest run`: 494 passed), and `next build` all clean. Extended the existing test file with #323's own coverage shape (drag continuation, two-finger-grip rejection, pointercancel, non-primary pointer/button rejection, `onLostPointerCapture` recovery) — 12 tests total, up from 7. Confirmed 4 of the 5 new tests fail against the pre-fix click-only code (stashed the fix, reran, restored) to make sure they're testing real behavior rather than passing vacuously.

### DIFF
--- a/apps/web/components/review/compare/__tests__/compare-scrubber.test.tsx
+++ b/apps/web/components/review/compare/__tests__/compare-scrubber.test.tsx
@@ -16,6 +16,10 @@ describe('CompareScrubber', () => {
     vi.spyOn(HTMLElement.prototype, 'getBoundingClientRect').mockReturnValue({
       left: 0, top: 0, right: 630, bottom: 8, width: 630, height: 8, x: 0, y: 0, toJSON() {},
     })
+    // jsdom doesn't implement this at all (it's undefined, which is why the
+    // component calls it with `?.`) — stub it so a dropped call is visible
+    // rather than silently passing.
+    Element.prototype.setPointerCapture = vi.fn()
   })
 
   afterEach(() => {
@@ -97,6 +101,22 @@ describe('CompareScrubber', () => {
 
     fireEvent.pointerDown(track, { pointerId: 2, clientX: 315, isPrimary: true, button: 0 })
     expect(onSeek).toHaveBeenLastCalledWith((315 / 630) * 63)
+  })
+
+  it('captures the pointer on pointerdown, so the drag survives the finger leaving the track', () => {
+    render(<CompareScrubber {...base} />)
+    const track = screen.getByTestId('compare-track')
+
+    fireEvent.pointerDown(track, { pointerId: 1, clientX: 63, isPrimary: true, button: 0 })
+
+    expect(track.setPointerCapture).toHaveBeenCalledWith(1)
+  })
+
+  it('marks the track touch-none, so a drag is not fought by the browser\'s own scroll gesture', () => {
+    render(<CompareScrubber {...base} />)
+    const track = screen.getByTestId('compare-track')
+
+    expect(track.className).toContain('touch-none')
   })
 
   it('positions markers at (tc + offset) / total and reports clicks per side', () => {
--- a/apps/web/components/review/compare/__tests__/compare-scrubber.test.tsx
+++ b/apps/web/components/review/compare/__tests__/compare-scrubber.test.tsx
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from 'vitest'
+import { describe, expect, it, vi, beforeEach, afterEach } from 'vitest'
 import { fireEvent, render, screen } from '@testing-library/react'
 import { CompareScrubber } from '../compare-scrubber'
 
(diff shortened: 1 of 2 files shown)


ravirajsinh45 (approved):
@CamReport This mirrors `progress-bar.tsx` closely and the details are right: pointer capture, the `isPrimary` and button guards, `pointercancel`, `onLostPointerCapture`, and the marker rows as `pointer-events-none` with `pointer-events-auto` restored per dot so a dot still wins over the hit-area extension.

Two gaps I found by mutating your fix rather than by reading the tests. Both leave all 12 passing:

1. Deleting `e.currentTarget.setPointerCapture?.(e.pointerId)` entirely. That is the line that keeps a drag alive once the finger leaves the track, which is most of what "drag support" means here. jsdom does not implement `setPointerCapture` (it is `undefined`, which is why your `?.` is needed), so asserting it needs a stub: assign `Element.prototype.setPointerCapture = vi.fn()` in setup and check it was called with the pointer id.
2. Removing `touch-none` from the track. That is what stops the browser scrolling the page instead of scrubbing. A class assertion is weak, but it is the difference between the feature working on a phone and not.

Neither is a defect in the code, and I am happy to merge as is if you would rather not add them. I am raising it because this is the second scrubber to get this treatment and the pattern will be copied a third time.

The rest of the coverage is strong: the second-pointer rejection, the non-primary cases and the lost-capture path are all pinned.

CamReport:
Good catch on both — added them rather than leaving it. Confirmed each fails against its own mutation before pushing (deleting the `setPointerCapture` call, and dropping `touch-none`), same way I'd checked the original 12 against the pre-fix click-only code. 14 passing now (6287c97).

PR #366, opened 2026-09-14
chore(deps): bump vite & vitest and patch vulnerable build-time transitives

Thedude7054:
## Summary

Brings the frontend's dev/build toolchain up to patched, non-vulnerable versions, as coordinated ahead of time. Every change here is a devDependency or a lockfile resolution — no runtime behaviour change, no app code touched.

## Changes

- **`vite` is now a direct devDependency, `^8.2.2`** (resolved 8.3.0). It previously only reached the tree transitively, which is how 8.0.1 sat in the lockfile despite the 2026 Vite CVE cluster (fixed in 8.0.16+). Declaring it pins the engine instead of letting it drift in unannounced.
- **`vitest` `^4.1.10` → `^4.1.11`.**
- **pnpm `overrides`** pull the other Dependabot-flagged build-time transitives up inside their existing `^`-ranges, so no parent had to change: `postcss` → 8.5.25, `nanoid@3` → 3.3.18, `glob@10` → 10.5.0, `js-yaml` → 4.3.1, `brace-expansion@1` → 1.1.18.
- CHANGELOG entry under `[Unreleased]`.

## Testing

Run from `apps/web`:

- `pnpm install --frozen-lockfile` — clean.
- `pnpm exec tsc --noEmit` — clean.
- `pnpm test` — 66 files, **487 tests pass**.
- `pnpm build` — succeeds.
- `pnpm lint` — exit 0.

Also confirmed the lockfile no longer resolves any of the previously-flagged versions (vite 8.0.1, vitest 4.1.10, @vitest/mocker 4.1.10, postcss 8.4.31, nanoid 3.3.17, glob 10.3.10, js-yaml 4.1.1, brace-expansion 1.1.12).

## Checklist

- [x] One focused change; devDependencies / build tooling only
- [x] CHANGELOG entry under `[Unreleased]`
- [x] Test / build / lint run and green
- [x] No documentation impact
- [ ] Screenshots — N/A (no UI change)


Thedude7054:
## Response to review

Thanks — went through each point:

**1. "Critical — frozen-lockfile validation fails (postcss importer mismatch)"** — I can't reproduce this. A clean `pnpm install --frozen-lockfile` (from a wiped `node_modules`) succeeds on this branch, and the CI Frontend Build job, which runs that exact command, is green. The lockfile's `postcss` importer entry (`specifier: 8.5.25 / version: 8.5.25`) is what pnpm wrote for the manifest's `^8.5.25` under the override, and it satisfies the range, so the frozen check passes. No change.

**2. "Vite 8.3.0 raises the toolchain's minimum Node to ^20.19 || ^22.12, docs still say Node 18+"** — Fair, and partly pre-existing: jsdom's `@asamuzakjp/*` deps already carried that `engines` constraint before this PR, but nothing admitted it. Fixed properly now:
- `apps/web/package.json` declares `"engines": { "node": "^20.19.0 || ^22.12.0 || >=24.0.0" }`;
- `docs/contributing.md` prerequisite updated from "Node.js 18+" to "Node.js 20.19+ or 22 LTS".

**3. "pnpm overrides don't apply to the Docker Compose dev path (npm install)"** — Real gap; fixed. The compose dev stack (`docker-compose.dev.yml`) installs with `npm`, which never reads `pnpm.overrides`. The same five pins are now also declared as npm-native top-level `overrides`, so both install paths enforce them. Verified empirically: pnpm only honors `pnpm.overrides` (with a top-level-only copy the lockfile reverts and `postcss@8.4.31` / `glob@10.3.10` come back), and npm only honors the top-level field (an isolated `npm install --package-lock-only` resolved exactly `postcss 8.5.25`, `nanoid 3.3.18`, `glob 10.5.0`, `js-yaml 4.3.1`, `brace-expansion 1.1.18`). Both copies are kept in sync.

Gate re-run after the changes: frozen install ok, `tsc` clean, 66 files / 487 tests pass, `pnpm build` and `pnpm lint` exit 0.

ravirajsinh45 (changes requested):
@Thedude7054 The security substance here is right, and I checked it rather than taking the body's word for it: vite 8.0.1 was genuinely affected, and nanoid 3.3.18, glob 10.5.0 and brace-expansion 1.1.18 all clear their advisories. Promoting `vite` from a transitive to a declared devDependency is the right call too, since it was governing the test toolchain while being pinned by nobody. Adding `engines` is housekeeping nobody asked for and the repo needed.

One blocker, and it is in the shape of the fix rather than the intent.

**`apps/web/package.json:75` breaks `npm install`.** The new top-level `overrides` block sets `"postcss": "8.5.25"` while devDependencies line 52 keeps `"postcss": "^8.5.25"`. npm refuses an override on a package the root depends on directly unless the two specs are byte-identical, so resolution aborts before any fetch:

```
npm error code EOVERRIDE
npm error Override for postcss@^8.5.25 conflicts with direct dependency
```

Reproduced in `node:20-alpine`, the dev image, with only the two package.json files differing: main exits 0, this branch exits 1.

That matters because `docker-compose.dev.yml:222` is `command: sh -c "npm install && npx next dev --hostname 0.0.0.0"`. It is the only npm invocation in the repo, and it is the path a contributor takes when they follow the quickstart. They get postgres, redis, minio and api healthy and `freeframe-web-1` exited, with nothing on :3000. There is no `restart:` key anywhere in that file, so it stays exited rather than looping. CI cannot see it: both install steps are `pnpm install --frozen-lockfile`, which is why all four checks are green on a manifest npm will not resolve.

The override itself is legitimate. next@14.2.35 pins postcss to 8.4.31 exactly, and that version carries four advisories. It is the unkeyed form that is the problem, so please key it the way its three neighbours already are:

```json
"postcss@8.4.31": "8.5.25"
```

That lifts Next's pinned copy and leaves the rest alone. I tried `"postcss": "$postcss"` as well and it fails with `Unable to resolve reference $postcss`, so that one is not the way out.

Worth adding alongside it: one CI step running `npm install --package-lock-only` in `apps/web`. Undo the fix and it goes red, which is the check this change currently has no answer to.

Smaller things, none blocking:

- `js-yaml` 4.3.1 is one release short. GHSA-2883-xcg3-v3hh covers `>=4.0.0 <4.3.2`, and 4.3.2 published on 2026-08-26. Not exploitable here since js-yaml only arrives via eslint, but if the block is being touched anyway it may as well land on 4.3.2.
- `postcss-selector-parser@6.1.2` via tailwindcss is affected by GHSA-w9m9-85wc-3x92 (fixed in 6.1.3) and is not in the block.
- The CHANGELOG entry needs `(#366 by @Thedude7054)`; every neighbouring entry carries a PR reference.
- Two claims in it do not hold. "Everything here is dev/build-only" is not quite true, since next declares postcss in `dependencies`, so the override is live during `pnpm build`. And the "Dependabot-flagged" provenance is not something I can confirm: the alerts API returns `[]` for every state on this repo, though the advisories themselves are all real. Please reword to cite the advisories directly.
- Heads up on ordering: this conflicts with Dependabot's #311 in both directions, on package.json and the lockfile. #311 moves postcss to `^8.5.26`, which an exact override would silently pin back. Whichever lands first, the other needs a real rebase and a re-read rather than an auto-merge.

ravirajsinh45 (commented):
@Thedude7054 Following up on my own review from yesterday rather than waiting on you, because I got something in it wrong and you should have the correction before you spend time acting on it. Nothing here is new work I am asking for beyond what we already discussed.

First, what holds up. I re-checked the security substance against the advisory API rather than trusting the PR body, and it is right:

- vite 8.0.1 carries five advisories including three high (GHSA-fx2h-pf6j-xcff, GHSA-v2wj-q39q-566r, GHSA-p9ff-h696-f583). `affects=vite@8.3.0` returns nothing. Promoting vite from a transitive nobody pinned to a declared devDependency is the right instinct, and it is the highest-value change in this PR.
- next@14.2.35 pins postcss to 8.4.31 exactly, and 8.4.31 carries four advisories, two of them high. Overriding that pin is genuinely worth doing.
- glob 10.3.10 and brace-expansion 1.1.12 sit inside open advisories while the other copies of each in the tree do not, so the `@10` and `@1` keys are exactly the right scope. That is a detail it would have been easy to get wrong.
- You spotted that the Docker dev stack installs with npm and added a top-level overrides block for it. That is the thing the rest of the repo's tooling misses, and it is the correct diagnosis. Only the shape of one entry is wrong.
- Adding `engines` and correcting the Node prerequisite in docs/contributing.md is housekeeping nobody asked for.

## Correction I owe you

My last review told you that `"postcss": "$postcss"` fails with `Unable to resolve reference $postcss`, and that therefore it was not the way out. That is not true, and I should not have asserted it. Measured on npm 11.6.0 just now, it exits 0. It is also a better fix than the `"postcss@8.4.31": "8.5.25"` I steered you toward. Please disregard that part of what I wrote.

Here are the four forms, with the resolved tree from `npm install --package-lock-only` on a copy of your manifest:

| override | exit | resolved postcss |
|---|---|---|
| `"postcss": "8.5.25"` (current) | 1, EOVERRIDE | aborts, nothing resolved |
| `"postcss@8.4.31": "8.5.25"` (my last advice) | 0 | root 8.5.28 **plus** `next/node_modules/postcss` 8.5.25, two copies |
| `"postcss": "$postcss"` | 0 | root 8.5.28 only, nested copy gone |
| `"postcss@8.4.31": "$postcss"` | 0 | root 8.5.28 only, nested copy gone |

`$postcss` is npm's documented escape from the byte-identical rule. It dedupes postcss to a single copy, 8.5.28 has no advisories at all, and because it tracks the direct dependency it will not need editing every time postcss is bumped. I would take that form.

## Blocker

**`apps/web/package.json:75` aborts `npm install`.** The top-level `overrides` block sets `"postcss": "8.5.25"` while devDependencies line 52 keeps `"postcss": "^8.5.25"`. npm refuses an override on a package the root depends on directly unless the two specs are byte-identical, so resolution stops before any fetch:

```
npm error code EOVERRIDE
npm error Override for postcss@^8.5.25 conflicts with direct dependency
```

Exit 1 on this branch, exit 0 on main's manifest, same command, npm 11.6.0.

The failure scenario: a contributor follows the quickstart and runs `docker compose -f docker-compose.dev.yml up --build`, which is what README.md and AGENTS.md both publish. `docker-compose.dev.yml:222` is `command: sh -c "npm install && npx next dev --hostname 0.0.0.0"`, and a grep confirms it is the only `npm install` in the repo, everything else being pnpm. npm exits 1, the `&&` short-circuits, `next dev` never starts. Postgres, redis, minio and api come up healthy and the web container is exited with nothing on :3000. There is no `restart:` key anywhere in that file, so it stays exited rather than retrying.

CI cannot see this. Both install steps (.github/workflows/ci.yml:174 and :232) are `pnpm install --frozen-lockfile`, which is why all four checks are green on a manifest npm will not resolve. pnpm is happy to override a direct dependency, so every pnpm path in the repo is genuinely fine, and a production self-host is not affected. It is specifically the contributor path that breaks.

Fix is the one line above. Once it is keyed or referenced, your CHANGELOG sentence about the overrides holding on whichever install path is used becomes true, which today it is not.

## Ordering against #392

One update: #311, which I mentioned last time, was closed yesterday and superseded by #392, so please ignore that reference. #392 moves postcss to `^8.5.28`.

I would land #392 first and rebase this onto it. Both PRs rewrite the same two files, so someone regenerates a lockfile either way, and better it is the PR that needs manifest edits anyway. Merging this one first is the worse order: with the unkeyed override present, pnpm applies it to the root importer and records `specifier: 8.5.25 / version: 8.5.25` while package.json reads `^8.5.28`, and `pnpm install --frozen-lockfile` passes on that pair, so nothing in CI would show that the postcss bump had been quietly undone. The same rewrite is already visible in this PR's own lockfile diff, where the importer went from `specifier: ^8.5.25` to `specifier: 8.5.25` against an unchanged package.json. Keying the override fixes that too, so it is one edit for both.

Worth saying that #392 leaves vite at 8.0.1 in its lockfile. Your vite bump is the piece of this that nothing else on the queue delivers, which is why it is worth landing both rather than folding one into the other.

## Non-blocking

- `js-yaml` 4.3.1 is still one release short of GHSA-2883-xcg3-v3hh (high, `>= 4.0.0, < 4.3.2`, patched in 4.3.2, published 2026-08-26). New detail since last time, and the reason I am repeating it: an exact pin is a ceiling as well as a floor. On the npm path, main already resolves js-yaml 4.3.2 on its own, and this override pulls it back to 4.3.1. Not exploitable here, since js-yaml only arrives through eslint and never reaches a shipped artifact, but it is the one pin that crosses a patched boundary. `4.3.2` in both blocks.
- Same mechanism, no advisory attached: nanoid resolves 3.3.19 -> 3.3.18 and brace-expansion 1.1.21 -> 1.1.18. Both pinned versions are clean so nothing is wrong today, it just freezes two transitives a little behind. Caret-shaped floors age better where you only need a minimum.
- `postcss-selector-parser@6.1.2` via tailwindcss is covered by GHSA-w9m9-85wc-3x92 (low, fixed in 6.1.3) and is not in either block. While you are in there.
- The CHANGELOG entry needs `(#366 by @Thedude7054)` under [Unreleased]; both neighbouring bullets carry a PR reference. Two claims in it also need a touch: the "so they hold whichever install path is used" sentence is the reverse of what npm does today, and "no runtime behaviour" is a stretch given next declares postcss under `dependencies`. Citing the GHSA ids directly would be stronger than pointing at Dependabot.
- docs/contributing.md says the `engines` field "enforces the same". It does not: pnpm and npm both warn and exit 0 on an impossible engine, and there is no `.npmrc` setting `engine-strict=true`. Either soften the word or add the `.npmrc`.
- Still worth adding: one CI step running `npm install --package-lock-only` in `apps/web`. It is the only check that would catch this class of thing, and today deleting the whole overrides block leaves all four checks green.

For transparency on what I did and did not run: the EOVERRIDE reproduction, the resolved-tree comparisons and the merged-with-#392 check were all run against copies in a scratch directory with npm 11.6.0 and pnpm 10.15.0 on macOS. I could not start the compose stack this time, so the container exiting is inferred from the reproduced exit 1 plus the `&&` and the missing `restart:` key, not watched end to end. I also did not run the web test suite, so the "487 tests" line rests on your green CI rather than on my own run.

The substance of this PR is good and I would like it in. It is one line and a version bump away.

ravirajsinh45:
Thanks @Thedude7054, and apologies for the delay on this. It shipped as #414, which I opened from your branch so it could go out with the rest of today's dependency work. Closing this as a duplicate of that.

Almost all of it went in as you wrote it: vite 8.0.1 to 8.3.x, vitest promoted to a direct devDependency instead of being reached transitively, the transitive pins (`nanoid@3` 3.3.18, `glob@10` 10.5.0, `brace-expansion@1` 1.1.18), the `engines` field on `apps/web`, and the contributing-guide prerequisite.

**On point 1, you were right and my review was wrong.** `pnpm install --frozen-lockfile` does succeed on your branch and there is no importer mismatch. I should have run it before writing that.

**The one change: the postcss override needed a key.** `postcss` was already a direct devDependency at `^8.5.25` before this PR, and npm refuses an unkeyed override on a name that is also a direct dependency unless the two specs match exactly:

```
npm error code EOVERRIDE
npm error Override for postcss@^8.5.25 conflicts with direct dependency
```

That is `npm install --package-lock-only --ignore-scripts` against `apps/web/package.json` exactly as this branch has it. It reproduces on both npm majors, so it is not a version quirk:

| manifest | npm 10.8.2 (node 20) | npm 11.19.0 (node 24) |
| --- | --- | --- |
| this branch | exit 1, EOVERRIDE | exit 1, EOVERRIDE |
| as merged | exit 0 | exit 0 |

npm rejects the whole manifest, not just the offending entry, so `nanoid`, `glob`, `brace-expansion` and `js-yaml` stop being enforced along with it. That is the opposite of what the PR is for, and it only bites on the npm path you correctly identified in point 3. Your isolated check almost certainly passed because an isolated manifest has no direct `postcss` devDependency for the override to collide with; the collision exists only in the real one.

The fix is to key the override to the copy that is actually vulnerable: `"postcss@8.4.31": "^8.5.28"`. 8.4.31 is the exact version `next@14.2.35` pins, so the override lands on that transitive copy and leaves the direct devDependency alone. `postcss-selector-parser@6.1.2` is pinned to 6.1.3 the same way, and `js-yaml` went to 4.3.2.

**No CI guard ships with this, which is worth recording here.** The obvious regression test is to make CI run the npm resolve, since a pnpm-only gate is exactly what let this through. I could not make it reliable. npm 10.9.8, which is what the runner ships, dies on this tree with an internal `Cannot read properties of null (reading 'edgesOut')` regardless of the overrides block, including with the block removed entirely. I bisected that rather than assuming, and the manifest turned out innocent of it. A check that fails identically on a correct manifest and a broken one is worse than no check, so the EOVERRIDE fix lands without one.

You are credited on the merge commit. The npm/pnpm split you found in point 3 is the reason this bug was findable at all.

PR #367, opened 2026-09-14
chore(deps): bump pytest to 9.0.3 (dev-only)

Thedude7054:
## Summary

Bumps the API's test tooling to a patched pytest. `pytest==8.4.2` is pinned in `apps/api/requirements.txt` and is caught by CVE-2025-71176; `9.0.3` clears it. Dev-only — nothing in the running service changes.

## Changes

- **`pytest==8.4.2` → `pytest==9.0.3`.**
- **`pytest-asyncio==0.26.0` → `1.4.0`.** `0.26.0` declares `pytest<9,>=8.2`, so pytest 9 cannot install without moving it. `1.4.0` is the current maintained 1.x line (`pytest<10,>=8.4`). The suite exercises asyncio via `asyncio.run()` inside sync tests rather than the plugin's markers, so there is no asyncio-plugin behaviour surface to regress here.
- CHANGELOG entry under `[Unreleased]`.

## Testing

Reproduced the CI backend job locally (Python 3.12 + Postgres 15):

- `pip install -r apps/api/requirements.txt` resolves cleanly with the new pins.
- `alembic upgrade head` — all migrations apply.
- `python -m pytest apps/api/tests/` — **475 passed**, 2 unrelated deprecation warnings.

## Checklist

- [x] One focused change; dev/test dependencies only
- [x] CHANGELOG entry under `[Unreleased]`
- [x] Backend suite run and green under the new versions
- [x] No documentation impact
- [ ] Screenshots — N/A (no UI change)


Thedude7054:
## Response to review

Correct — `apps/api/requirements.txt` is installed by the production Dockerfile (and always has been), so "dev-only" was the wrong framing; pytest was shipping in every image. Rather than just reword, I fixed it properly:

- Test-only pins moved out of `requirements.txt` into a new `apps/api/requirements-dev.txt` (`-r requirements.txt` + `pytest==9.0.3` + `pytest-asyncio==1.4.0`). The Dockerfile continues to install `requirements.txt` alone — the production image no longer carries the test runner.
- CI's backend-test job (and its pip cache key) now installs the dev file — the single intentional `.github/workflows/ci.yml` change in this PR.

Verified in a Python 3.12 + Postgres 15 mirror of the CI job: runtime `requirements.txt` resolves with `pytest` / `pytest_asyncio` not installed (import probe `False`), the dev file installs cleanly, `alembic upgrade head` applies, and `python -m pytest apps/api/tests/` is **475 passed**.

ravirajsinh45 (changes requested):
@Thedude7054 The dependency reasoning here is exact and I could not fault it. I re-derived the advisory independently, checked that pytest-asyncio 0.26.0 really does pin `pytest<9` so it has to move, and confirmed your "dependencies-only" claim about the 0.x to 1.x jump: the suite has zero `@pytest.mark.asyncio`, zero `async def test`, zero `pytest_asyncio` imports and no `asyncio_mode` anywhere. The underlying point was worth catching, because `Dockerfile.prod:38` really was shipping pytest into the production image.

One blocker, and it comes from a naming slip in your own comment.

**The dev image loses pytest too.** `apps/api/requirements-dev.txt:1-3` says "The production image (apps/api/Dockerfile) installs requirements.txt alone". `apps/api/Dockerfile` is the **dev** image. The production one is `apps/api/Dockerfile.prod`. Both install `requirements.txt`, and this PR changes neither, so removing the pins from that file takes pytest out of both.

Built it from this branch to be sure:

```
$ docker build -f apps/api/Dockerfile -t ff-367-dev:test .
$ docker run --rm ff-367-dev:test python -m pytest --version
/usr/local/bin/python: No module named pytest
```

`docker-compose.dev.yml` builds api and the four Celery services from that file, so after this merges the command five checked-in documents tell people to run stops working:

- `.github/PULL_REQUEST_TEMPLATE.md:11`, the checklist every contributor is asked to tick
- `CONTRIBUTING.md:24`
- `docs/contributing.md:83`, `:86`, `:89`
- `AGENTS.md:95`

`AGENTS.md:92`'s bare local `python -m pytest` breaks the same way for anyone building a venv from requirements.txt.

The break is delayed, which makes it worse rather than better: anyone with a pre-merge image keeps working, and only a fresh clone or the next `--build` fails, with nothing in the diff pointing at why.

It is not a one-line edit. `Dockerfile:12` copies only requirements.txt into the build context, and requirements-dev.txt opens with `-r requirements.txt`, so the dev Dockerfile needs to COPY both and install the dev one. `Dockerfile.prod` stays on requirements.txt, which is what actually delivers the goal.

A one-line `docker run --rm <dev image> python -m pytest --version` appended to the docker-build job would go red on exactly this mutation, if you want the guard.

Smaller things:

- `ci.yml:72` narrows `cache-dependency-path` to requirements-dev.txt. setup-python hashes that file's literal contents and does not follow the `-r` include, so the runtime pins Dependabot rewrites monthly stop busting the cache. Listing both paths keeps it honest.
- The suite prints `473 passed, 2 skipped`, not 475. 475 is the collected count. The CI floor greps `\d+ passed`, so it reads 473. Worth correcting in the CHANGELOG and the body.
- The CHANGELOG entry needs `(#367 by @Thedude7054)`, matching the form your own #286 entry uses.
- `requirements-dev.txt` has no trailing newline.
- Ordering note: this conflicts with #366 on CHANGELOG.md only, since both insert at the head of the same block. Trivial, just worth merging them deliberately rather than together.

ravirajsinh45 (commented):
@Thedude7054 Picking this back up. Nothing has changed on your side yet and our last round is only a day old, so treat this as an update to our position rather than a nudge.

First, we got something wrong and you were right.

Our last review said the suite prints `473 passed, 2 skipped`, not 475, and that 475 was only the collected count. That is false. The Backend Tests log for run 34826001941, on this exact head, prints:

```
platform linux -- Python 3.12.14, pytest-9.0.3, pluggy-1.6.0
======================= 475 passed, 2 warnings in 9.16s ========================
Tests passed: 475
```

No skips. Collected and passed coincide here. Your number is correct in the PR body and in the CHANGELOG, and that nit is retracted.

The restructure you did in response to the first round is the right call, and it lands. `apps/api/Dockerfile.prod:38` really was installing `requirements.txt` with `pytest==8.4.2` in it, so the production image was shipping the test runner, and after this PR it does not. We also checked that nothing was quietly leaning on pytest's transitive closure at runtime: building from your `requirements.txt` alone and importing `apps.api.main`, `apps.api.tasks.celery_app` (the target the four Celery services run) and `packages.transcoder` all import clean, and nothing in `apps/api` or `packages` imports `packaging`, `pluggy`, `iniconfig` or `yaml` outside `apps/api/tests/`. Resolved runtime versions are unchanged. The CI scoping is right too: Backend Tests is the only job that runs pytest and it is the only one you switched, and leaving the Lint job on `requirements.txt` is correct rather than an oversight, since all it does is `python -m py_compile apps/api/main.py`.

One blocker, unchanged from last time.

**The dev image loses pytest.** `apps/api/Dockerfile:12-13`

`apps/api/requirements-dev.txt:1-3` says "The production image (`apps/api/Dockerfile`) installs requirements.txt alone". `apps/api/Dockerfile` is the **dev** image. Production is `apps/api/Dockerfile.prod`. Both install `requirements.txt`, this PR changes neither, so pulling the pins out of that file takes pytest out of both.

Failure scenario, concretely. Merge `901f96ed`, then on a fresh clone:

```
$ docker compose -f docker-compose.dev.yml up -d --build api
$ docker compose -f docker-compose.dev.yml exec -w /workspace api python -m pytest apps/api/tests/ -q
/usr/local/bin/python: No module named pytest
```

That second command is the one six places across four checked-in documents tell contributors to run: `.github/PULL_REQUEST_TEMPLATE.md:11` (the checklist every contributor is asked to tick), `CONTRIBUTING.md:24`, `docs/contributing.md:83`, `:86`, `:89` and `AGENTS.md:95`. `AGENTS.md:92`'s bare local `python -m pytest` breaks the same way for any venv built from `requirements.txt`.

The Docker daemon is down on this machine today, so the `No module named pytest` line above is from the build we ran in the previous round rather than one we re-ran now. What we did re-verify on this head, without Docker: the diff touches four files (`.github/workflows/ci.yml`, `CHANGELOG.md`, `apps/api/requirements-dev.txt`, `apps/api/requirements.txt`) and neither Dockerfile is among them; `Dockerfile:12-13` is byte-identical to main; `requirements-dev.txt` is never COPYed into the build context; and reproducing `Dockerfile:13`'s install in a venv from your `requirements.txt` gives `ModuleNotFoundError: No module named 'pytest'`, so nothing pulls it in transitively. There is no fourth install path in the repo, only `ci.yml:72/75`, `ci.yml:215/218`, `Dockerfile:12-13` and `Dockerfile.prod:37-38`.

The break is delayed, which makes it worse rather than better. Anyone holding a pre-merge image keeps working, and only a fresh clone or the next `--build` fails, with nothing in the diff pointing at why. `docker-compose.dev.yml:99`, `:126`, `:159` and `:186` build the four Celery services from the same file, so they lose the runner too, though nobody runs pytest in those.

Nothing in CI catches it. The Docker Build job (`.github/workflows/ci.yml:248-264`) builds the four images and then parses the two compose files as YAML, but never executes anything inside an image. That is why all four checks are green on this head while the dev image is broken.

The fix is not one line, since `requirements-dev.txt:10` opens with `-r requirements.txt` and `Dockerfile:12` copies only the runtime file:

```dockerfile
COPY apps/api/requirements.txt apps/api/requirements-dev.txt /workspace/apps/api/
RUN pip install --no-cache-dir -r /workspace/apps/api/requirements-dev.txt
```

Both files are in the build context: `docker-compose.dev.yml:66` and `ci.yml:249` both build with context `.` at the repo root, and the repo has no `.dockerignore`. `Dockerfile.prod:37-38` stays on `requirements.txt`, which is what actually delivers your goal. pip resolves the `-r` include relative to the referencing file, which we confirmed by installing the dev file from a different working directory and getting the full runtime set alongside pytest 9.0.3.

If you want a guard so this cannot come back silently, one line after `ci.yml:249`: `docker run --rm freeframe-api:test python -m pytest --version`. It goes red on exactly this mutation and on nothing else.

Non-blocking, all small:

- `apps/api/requirements-dev.txt:2` names `apps/api/Dockerfile` as the production image. Naming both would have surfaced the blocker: production is `Dockerfile.prod`, dev is `Dockerfile`, and CI and a local venv install the dev file. Same correction in the CHANGELOG entry, which says "the image installs the runtime file alone" in the singular.
- `apps/api/requirements-dev.txt:12` has no trailing newline. The next `echo 'pytest-cov==...' >>` produces `pytest-asyncio==1.4.0pytest-cov==...`, which pip rejects rather than installing either pin.
- `.github/workflows/ci.yml:72` narrows `cache-dependency-path` to `requirements-dev.txt`. setup-python hashes the files that path matches and does not follow the `-r` include, so a Dependabot bump that only rewrites `requirements.txt` no longer busts the cache. The install stays correct, since pip's own cache is keyed by URL, so this costs a cache hit rather than correctness. Listing both paths keeps the key honest.
- The CHANGELOG entry needs `(#367 by @Thedude7054)`, matching the form of most of its neighbours in `[Unreleased]` and of your own earlier entry. While you are in there, 475 is right today but goes stale on a rebase, since main collects more. Either drop the number or re-check it at merge.
- Not for this PR, just worth confirming after it lands: `.github/dependabot.yml:27` names only `apps/api/requirements.txt` in its comment. Please check Dependabot keeps the new dev file patched too, since an unwatched dev file is how the advisory you are fixing here would come back.

Ordering: this conflicts with #366 on `CHANGELOG.md` only, both inserting at the head of the same block, so we will merge them one at a time. No conflict with the open Dependabot PR #391, and #391 does not put the pytest pins back.

ravirajsinh45:
Thanks @Thedude7054, and sorry for the wait on this one. It shipped as #413, which I opened from your branch so it could go out with today's release rather than sit any longer. Closing this as a duplicate of that.

Your diff went in as you wrote it: the new `apps/api/requirements-dev.txt` (`-r requirements.txt` plus pytest 9.0.3 and pytest-asyncio 1.4.0), the two test pins removed from `requirements.txt`, and CI switching both its install step and its pip cache key over to the dev file. Your second round was the right call too. "Dev-only" was the wrong framing precisely because `Dockerfile.prod` installs `requirements.txt`, and the CHANGELOG entry now says so in those terms.

Two things I added on top, both found while checking the change rather than reading it:

**The dev image could no longer run the test suite.** `apps/api/Dockerfile` still copied and installed `requirements.txt` alone, so `docker exec -w /workspace freeframe-api-1 python -m pytest apps/api/tests/`, which is what AGENTS.md, CONTRIBUTING.md and the PR template all tell contributors to run, came back with no pytest installed. CI stayed green because the Backend Tests job installs with pip directly on the runner and never touches that image. The dev Dockerfile now copies both files and installs `requirements-dev.txt`. `Dockerfile.prod` is deliberately untouched, since that is what keeps pytest out of production.

**A check so neither half can regress quietly.** The Docker Build job now asserts both directions: `python -m pytest --version` has to succeed in the dev image and has to fail in the prod one. Asserting only one direction would stay green for a change that fixed one side by breaking the other, which is roughly how the gap above got in.

You are credited on the merge commit. Thanks for taking this from a version bump to the actual fix.

PR #368, opened 2026-09-15
feat(web): rename an asset from the uploads panel, while it uploads

### LINKED ISSUE #330: Name an asset from the upload list, instead of only in the dialog before it
### Problem

The asset name can only be set in one place: the "Upload asset" dialog, before the transfer starts, and only when a single file is selected. Pick two files and the name field disappears — each asset is named after its file, and correcting that means opening each one afterwards and renaming it there.

That is the common case rather than the exception. Files arriving from a camera, an export preset or a shared drive are named A047C012_230915_R1AB.mov or Sequence 01_v3_FINAL_bs.mp4, and the review link that goes to a client shows whatever the file was called.

Dropping files onto the project makes it sharper. There is no dialog on that path by design — dragging a file onto a project has already said which file and which folder, and asking again would spend the click the drop just saved — so every dropped asset takes its filename and the naming step is gone entirely.

The dialog's own layout points at the answer: it already lists the files, one per row, with the size on the right. The row is the natural place to put the name, and there is nothing in the way of having it there for ten files as readily as for one.

Image
Image
these are the 2 dialogues currently popping up when using the upload button. I suggest uploading right away just like in the other dropzones - and offer the rename feature which is quite useful directly in the upload panel. a small pencil button would suffice and i could instantly rename the asset right in that view. saves clicks and makes for a fast batch upload and rename workflow

Image

### Proposed solution

Make the asset name editable per row, in the list, wherever that list is shown.

Each row carries a small pencil next to the name. Clicking it turns the row into a text field, in place. It edits the asset name only — the file being uploaded is untouched, and the extension is not part of what is edited.
The same row works for one file and for ten. One layout, one interaction, whatever the selection is.
The dedicated "Asset name" field below the list is then redundant for the single-file case and can go, which makes the dialog shorter rather than longer.
Two things follow from it:

"Change files" becomes "Cancel", and closes. Today it clears the selection and returns to the picker inside the same dialog, which is a third thing between "upload these" and "not now". With the names editable in place there is no reason to swap the selection rather than start again, and two buttons that mean "yes" and "no" are easier to read than three.
The same list can be shown after a drop, if you want the naming step there too — the rows are already uploading by then, so it costs nothing and blocks nothing. Renaming a row that is at 40% is just a field on the asset.

### Alternatives considered

Rename afterwards from the grid. What we do today. It works, and it is the reason this is a papercut rather than a bug — but it is one dialog per asset after the fact, on exactly the assets whose names were worst to begin with.

Ask for names before the upload starts, for dropped files too. This is the obvious way to give the drop path a naming step, and we think it is the wrong one: it puts a dialog in front of the gesture whose whole point was not needing one. Naming during the upload gets the same result without paying for it.

Derive better names. Strip camera-roll prefixes, _FINAL, version suffixes. Guesswork, and wrong for someone whose naming scheme is deliberate.

Happy to open a PR for this if the shape is right. Reported from a self-hosted instance where most uploads arrive from a camera and go straight to a client review link, so the filename is what the client sees.

Lennart-Pingpong:
## Summary

Closes #330. The name is decided in the "Upload asset" dialog before a byte moves, which is the one moment the file is not on screen next to it. Since #327 a dropped file skips that dialog altogether and arrives named after whatever the camera or the NLE called it, with no way to correct it until the transfer ends.

A row in the uploads panel now takes a click on the name, or on a pencil beside it, and edits in place. Enter saves, Escape discards, leaving the field saves, which is the same key handling the folder rename in the sidebar already has.

This follows the shape you set in the issue: the per-row pencil in the panel, branched off `main` independently of #327, and no editable names in the pre-upload dialog list.

## Your three conditions

- **A queued row has no `assetId` yet.** You suggested gating on it or having initiate read the live entry. It does the second. The pencil is offered from the moment the row appears. Until `/upload/initiate` answers, the row is the only record of the name, and it is the record initiate reads, so an edit there is stored without a request.

  One detail that is not what it looks like: there is no window between "row appears" and "initiate is sent". `startUpload` runs synchronously up to its first `await`, so initiate is already on the wire before the row renders. The window that exists is the round trip. A rename made during it lands in the row while the request carrying the old name has already left, so once initiate returns the store compares and sends one `PATCH`. If that correction fails, the upload does not: the bytes are the point, and the name can be typed again. Covered by `upload-rename-before-initiate.test.ts`.

- **History rows get no pencil.** Rows now carry `fromHistory` rather than being told apart by an id prefix. The reason turned out to be stronger than a 403 for a reviewer, see "Found along the way" below.

- **The grid behind the panel refreshes.** The panel has no `mutateAssets`, so a successful `PATCH` is followed by a keyed `mutate` over every key containing `/assets`. Without it the card kept the old name until the next navigation.

And the one you asked to fold in: single-file upload stripped the extension and multi-file did not. `assetNameFromFile` in `lib/utils.ts` is now used by the dialog's single-file path, its multi-file path, and the drop path from #327. A name that is nothing but an extension keeps it, since stripping `.gitignore` leaves an asset called "".

## Behaviour on failure

The rename is optimistic: the row shows the new name at once, since snapping back for a round trip reads as the edit having failed. If the server refuses, the old name comes back and the reason is shown in the row's error line, next to the status. An empty or unchanged name sends nothing.

## Found along the way

The upload history is not "my uploads". `fetchHistory` calls `/me/assets` with no `filter`, which takes the `else` branch in `routers/me.py`: every asset in every project the user is a member of, in whatever role, plus anything shared with or assigned to them. On a single-user instance nobody notices. On a shared one, the Complete tab lists colleagues' material from projects where the user only reviews.

That is why history rows stay without a pencil: most of them are not the user's to rename. It is also a question on its own, whether the panel should ask for `filter=owned`, so I have left it out of this PR and can file it separately if you agree it is worth one.

## Tests

22 new tests plus 5 for the helper. Every guard was checked by removing it: the test must go red.

- `stores/__tests__/upload-rename.test.ts`: the store action (trim, no-op on unchanged or empty, rollback on refusal, the `mutate`, other rows untouched, queued row renamed without a request, history row refused)
- `stores/__tests__/upload-rename-before-initiate.test.ts`: the round-trip window
- `components/layout/__tests__/uploads-panel-rename.test.tsx`: click on name, pencil, Enter, blur, Escape, one request per edit, visible error, queued row offered, history row not offered

Two of these were green without testing anything when first written. The clearer case: the history row fixture was `complete` while the panel opens on Active, so the row was never rendered and "no pencil on it" held trivially. The test now switches tabs first and fails when the guard is removed. jsdom has no `IntersectionObserver`, which the Complete tab needs for its infinite scroll, so the test provides one.

Not covered by a test: the one-line switch of the drop path in `page.tsx` to the helper. There is no test that renders the project page, and the helper itself is covered.

`tsc --noEmit` clean, `pnpm test` 514 passed, `pnpm build` succeeds, run in `node:20` the way CI runs them.

## CHANGELOG

Two entries under `[Unreleased]`, one in Added (the rename) and one in Fixed (the naming helper), both `(#330, #368 by @Lennart-Pingpong)`.


ravirajsinh45 (changes requested):
@Lennart-Pingpong This follows the shape we agreed in #330 and the store logic is well judged. The answer to the queued-row question is the better of the two I offered, and the detail about there being no window between the row appearing and initiate going out, only the round trip, is right and worth having written down. Folding in the single-file vs multi-file naming split was the right call too.

Four things to fix. The first is the one that will be visible to users.

**1. `upload-store.ts:778-782` blanks every `/assets` consumer behind the panel.** The refresh after a successful rename is a three-argument call:

```js
globalMutate((key) => …includes('/assets'), undefined, { revalidate: true })
```

swr short-circuits to a pure revalidate only when fewer than three arguments are passed. With three, it falls through to `populateCache`, which defaults to true, and writes `data: undefined` into every matching key before refetching.

So: rename a row while on `/projects/X` and the grid behind the panel is replaced by skeletons for the length of the refetch, because `asset-grid.tsx` renders skeletons instead of cards when `isLoading`. On the review screen it is worse, since `use-comments` does `data ?? []` and the review page passes no `isLoading`, so the comment list shows "No comments yet" instead of a spinner. A probe against real swr gave `{during:'SKELETON', fetches:2}` for the three-argument form and `{during:'Asset A', fetches:2}` for the one-argument form. Both revalidate; the extra arguments buy nothing and cost the blank. The fix is deleting two arguments.

Once that is in, the filter is still a substring test over the whole path, so a rename also refetches the open asset's comments. Narrowing it would be nice but is not the blocker.

**2. `upload-store.ts:547-552`: a refused mid-initiate correction is silent, and then blocks the retry.** The correction PATCH is `.catch(() => undefined)`, but `applyName(trimmed)` has already run and the function returns `null`, which the panel reads as success, so no error line renders. The sibling path at `:767-772` does the opposite and rolls back.

Concretely: drop a file onto a folder, rename during the round trip, and if that PATCH is refused (a 502 through a proxy, a dropped connection, or a 500 on a name past the `String(255)` column, since the input has no `maxLength` and `AssetUpdate.name` is an unconstrained `Optional[str]`) the row shows the new name for the rest of the session while the asset, the grid, search and the share link all show the old one.

The justification in the comment, that the name is a correction the user can make again, is not available either: `:745` returns early on `trimmed === row.assetName`, and the row already holds the refused name. A probe confirmed it, retyping the same name sends zero requests. Roll the row back to the sent name on rejection and surface the reason, the way the other path does.

**3. The `fromHistory` guard's producer is unpinned.** `mergeHistoryAssets:459` is the only place `fromHistory: true` is set, and no test reaches it: both tests hand-build the fixture and nothing calls `fetchHistory`. Delete that one line and the suite is still 514 green, while in the app every Complete-tab row gains a pencil, which per your own "found along the way" ends in a 403 for a reviewer. The guards themselves are pinned; only the producer is not. One test that stubs `/me/assets`, calls `fetchHistory()` and asserts the produced row offers no pencil closes it.

**4. The `### Fixed` entry's subject has no regression test.** Revert `page.tsx:408` to `: file.name;` and the suite is 514 green, so the surviving mutation is the bug itself: three clips selected upload as `A047C012.mov` while one clip gives `A047C012`. I am not asking you to render the 1359-line page. Extracting the two-line rule from `handleStartUpload` into a small pure helper and testing the `count > 1` branch would do it. Note that `:354` is what actually decides the single-file name, so `:408` is the genuinely new behaviour and the only one worth pinning.

Non-blocking:

- `:752` and `:770`: `previous` is captured at entry and the rollback writes it unconditionally, with no check that this call is still the latest. Rename A to B, then B to C before B settles, and a late failure can restore a stale name.
- `:760-765`: `applyName` runs before the `!row.assetId` check and `null` reads as success. Right for a queued row, wrong for one whose initiate already failed.
- `uploads-panel.tsx:131-141`: commit and cancel unmount the focused input without returning focus to the trigger, so `document.activeElement` becomes `<body>` and the next Tab restarts at the top of the document.
- A correction to something I was going to raise and checked instead: the pencil's `opacity-0 group-hover:opacity-100` is not invisible on touch here, since this Tailwind config has no `hoverOnlyWhenSupported`. It is fine as written.
- Your "the upload history is not my uploads" finding is correct, `/me/assets` with no filter does take the else branch. Please do file it separately, it is worth an issue on its own.
- The PR body still credits both CHANGELOG entries as `(#330 by @Lennart-Pingpong)`; the file itself is right.
- Ordering: this conflicts with #312 in CHANGELOG.md, uploads-panel.tsx and upload-store.ts. The one that needs care rather than a textual resolve is `mergeHistoryAssets`, where both PRs append a key to the same row shape.

Lennart-Pingpong:
All four fixed. The first one is the one I am most glad you caught, because it is visible to everyone and I had read the three-argument form as harmless.

**1. `globalMutate` blanked every `/assets` consumer.** Exactly as you describe: below three arguments swr short-circuits to a pure revalidate, with three it falls through to `populateCache` and writes `undefined` into every matching key first. One argument now. There is a test that asserts the call has exactly one argument, since the mistake is invisible in any assertion about *what* was revalidated — both forms revalidate the same set.

I left the matcher as the substring test. Narrowing it so a rename does not also refetch the open asset's comments would be nice, but as you said it is not the blocker, and doing it properly means knowing which key shapes are meant to be covered.

**2. The refused correction was silent and then blocked the retry.** Both halves are fixed together, because they are the same thing: the row now goes back to the name the asset actually has, and the reason is recorded on the row as `renameError`, which the panel renders where it renders the interactive path's error. Putting the name back is also what unblocks the retry — with the refused name left in place, `renameUpload`'s early return meant a second attempt at the same name sent nothing at all. Two tests: one that the name comes back with the reason, one that the same name can then be retried and succeeds.

The asymmetry you pointed at was real and I had not seen it: the sibling path at `:767-772` rolls back, this one did not, and the comment justifying it ("a correction the user can make again") was not true precisely because of the early return.

**3. The `fromHistory` producer.** You are right that only the guards were pinned. The new test stubs `/me/assets`, calls `fetchHistory()` and asserts the produced row offers no pencil, so deleting `fromHistory: true` from `mergeHistoryAssets` now reddens it.

**4. The `### Fixed` entry's subject.** Extracted as you suggested: `uploadNameForFile(fileName, typedName, fileCount)` in `lib/utils.ts`, tested on the `count > 1` branch, and used by `handleStartUpload`. Reverting it to the raw filename reddens a named test. You are right that `:354` is what decides the single-file name — that branch is in the helper too, but the multi-file one is what the entry is about.

**The non-blocking ones.** I have left these, and would rather they were their own issues than grow this PR further; say the word if you would prefer them here:

- `:752`/`:770`, `previous` captured at entry and restored unconditionally, so a late failure can restore a stale name after A→B→C. Real, and it needs a per-call generation to fix rather than a line.
- `:760-765`, `applyName` running before the `!row.assetId` check, so `null` reads as success for a row whose initiate already failed.
- `uploads-panel.tsx:131-141`, focus not returned to the trigger on commit or cancel.

Thank you for the correction on `hoverOnlyWhenSupported` — I would have changed that for no reason.

**The `/me/assets` finding is filed separately**, as you asked. The short of it: the `fromHistory` flag this PR carries is a workaround for that endpoint rather than a design, and anything else added to the panel later has to remember the same thing.

**Ordering.** This still conflicts with #312 in `CHANGELOG.md`, `uploads-panel.tsx` and `upload-store.ts`. `mergeHistoryAssets` is the one that needs care rather than a textual resolve, as you said — both PRs append a key to the same row shape, and this one now adds `renameError` to it as well. Happy to rebase this onto #312 or the other way round, whichever order you want to take them in.

ravirajsinh45 (approved):
@Lennart-Pingpong All four are fixed. I checked each one by mutating the code and running the suite rather than by reading the diff, so what follows is measured. Baseline at `0535d43b`: 70 files, 530 tests green.

**1. The three-argument `globalMutate`.** `upload-store.ts:809` is one argument now, and you pinned the right thing. Putting `, undefined, { revalidate: true }` back reddens `revalidates without blanking what is on screen first` and nothing else. Asserting the arity rather than the effect is what makes that test work at all: both forms revalidate the same set, so no assertion about *what* got refetched could have caught it, and the swr mock forwards `...args` verbatim so the length check is faithful. The `typeof key === 'string'` guard costs nothing here, every asset list key is a template string and there is no array key anywhere in `apps/web`.

**2. The refused mid-initiate correction.** `upload-store.ts:562-570` puts `sentName` back and records `renameError` on the row, and `uploads-panel.tsx:259-262` renders it in the always-rendered detail line, so it shows on a pending, uploading or complete row alike. Swapping the try/catch back for `.catch(() => undefined)` reddens exactly two tests, `puts the refused name back and says why` and `lets the same name be retried after a refusal`, and the second one fails on `api.patch` never being called, which is the early-return symptom itself. Both halves really are closed, and you were right that they are one thing: putting the name back is what unblocks the retry.

**3. The `fromHistory` producer.** `uploads-panel-rename.test.tsx:181-208` goes through `/me/assets` and `fetchHistory()` as asked, and deleting `fromHistory: true` from `mergeHistoryAssets` reddens that test alone. The `getByText('Someone elses cut')` alongside the `queryByRole` keeps it honest, since it fails too if the row stops rendering rather than the pencil appearing.

**4. The `### Fixed` entry's subject.** `uploadNameForFile` at `lib/utils.ts:200-210` is exactly the shape we discussed, and changing the multi-file branch to return the raw filename reddens a named test. Two things it got right beyond the ask: an all-whitespace typed name now falls back instead of naming the asset "   ", and the single-file fallback is the stripped name rather than the raw one. The call site at `page.tsx:406` is still unpinned, reverting it to `file.name` leaves the suite green, but that was the trade we proposed when we said we were not asking for the 1359-line page to be rendered, so it stays where we put it.

CHANGELOG is right: both entries are under `[Unreleased]` and credited correctly.

One thing to own. The rollback in item 2 is something I asked for, and it brought a narrow window with it: the new catch writes `sentName` with no check that it is still the latest rename, where before it wrote nothing at all. Rename during the initiate round trip, rename again once `assetId` exists and let that second one succeed, then let the correction fail, and the row snaps to a name neither you nor the server holds, with an error about a rename that no longer matters. It is recoverable by retyping and it is the same class as the `:788` item you deferred, so it is not a blocker, but it belongs in that same issue rather than a separate one, because one per-call generation closes both.

This is ready. Non-blocking below. I pushed the second one to your branch as `1cc651a7`, since it was a missing test rather than a decision; the rest is yours unless you want it here:

- `upload-store.ts:566`: the unconditional `sentName` rollback above. Same issue as the deferred `:788` item.
- `uploads-panel.tsx:259`: the screen half of fix 2 has no test. I deleted `?? upload.renameError` from the render and all 530 tests still passed. One line in `uploads-panel-rename.test.tsx` seeding `renameError` on a row and asserting the text closes it, and it is worth having before this file is rebased against #312.
- `uploads-panel.tsx:145`: no `maxLength` on the input, against `Asset.name`'s `String(255)` and an unconstrained `AssetUpdate.name`. Not a regression, the existing rename dialog has neither a limit nor any error handling, so this path is already the better of the two.
- `uploads-panel.tsx:170`: the name and the pencil are two tab stops with the same `onClick`, and the name button announces only the asset name, since `title` maps to the description. `tabIndex={-1}` on the name collapses it to one stop without touching the mouse affordance.
- The three you deferred are fine as their own issues, as is the `/me/assets` finding you filed.

On ordering: let us take this one first since it is ready, and rebase #312 onto it. `mergeHistoryAssets` and the row shape are the parts that need care rather than a textual resolve, and that direction means #312 resolves against a row that already carries both `fromHistory` and `renameError`.

Approving and merging this first, as above.

PR #369, opened 2026-09-15
fix(web): open the uploads panel beside the sidebar, not over it

Lennart-Pingpong:
Summary
With the sidebar expanded, opening Uploads draws the panel on top of it and cuts every label back to its first letter ("P", "N", "U"), which undoes the one thing expanding the sidebar is for.

The panel's left offset is [--ff-left:52px], the collapsed rail's width. It has been 52px since the panel was added, and #347 carried the number over into the safe-area form. The layout owns sidebarCollapsed, so it now passes it to UploadsPanel, which picks [--ff-left:52px] or [--ff-left:220px]. Both are written out as whole class names because Tailwind only generates what it finds literally.

Test
`app/(dashboard)/__tests__/layout-uploads-panel-rail.test.tsx` renders the layout with the panel open, toggles the sidebar, and asserts the offset follows. Removing the prop in the layout reddens one of its two assertions.

`app/(dashboard)/__tests__/uploads-panel-offset-css.test.ts` runs the repo's own Tailwind over the panel file and asserts the collapsed offset is emitted unconditionally and the expanded one only inside `@media (min-width: 768px)`. That is the level the constraint actually lives at: a computed class name renders identically and emits no rule at all.

tsc --noEmit clean, vitest run green (574), in node:20.

before the fix 
image

after the fix
Bildschirmfoto 2026-09-15 um 22 54 31


ravirajsinh45 (changes requested):
@Lennart-Pingpong Every factual claim in the description held when I checked it, including the two easiest to get wrong: the 52 and 220 constants match `sidebar.tsx:78` and `layout.tsx:70` exactly, and both class names are written out literally so Tailwind emits them. The refactor also preserves #347's and #341's work rather than quietly undoing it, since `left-safe`, `h-dvh` and `pb-safe` all sit outside the ternary and cannot drift apart. The comment at `:263-267` explains the non-obvious constraint at the exact place someone would otherwise tidy it away, and my own mutation testing proved that comment right.

One blocker, and it is one gate away.

**`uploads-panel.tsx:269-270` applies the 220px offset at every viewport, against a hard `w-[380px]`.** `grep -nE '\b(sm|md|lg|xl):'` returns nothing in uploads-panel.tsx, sidebar.tsx or `app/(dashboard)/layout.tsx`, so on a phone the panel spans 220 to 600 CSS px.

Measured in headless Chrome against the repo's own compiled Tailwind, rendering the head markup at `--ff-left:52px` against `--ff-left:220px` with transitions disabled:

| viewport | element | before | after |
|---|---|---|---|
| 402px | X close | 387-415, 13px clipped | 555-583, fully off |
| 402px | "Complete" tab | 242-329, visible | 410-497, fully off |
| 430px | X close | 387-415, fine | fully off |
| 430px | "Failed" tab | 329-417, fine | fully off |

`document.documentElement.scrollWidth === window.innerWidth` at 393, 402 and 430, so there is no horizontal scroll and the overflow cannot be reached by scrolling, swiping or zooming. The only way out is a backdrop tap. It is two taps to get there with the panel closed: Collapse/Expand on the rail, then Uploads.

I should be straight that at 430px the pre-PR panel is entirely healthy, so "it was clipped anyway" is not available as a mitigation. What does mitigate it is that the trigger needs the sidebar manually expanded, which is not the default and is not persisted (`useState(true)` at `layout.tsx:22`), so a fresh load behaves exactly as today.

The fix compiles under the same literal-class constraint. I ran the repo's tailwindcss over it and `.md\:\[--ff-left\:220px\]` is emitted inside `@media (min-width: 768px)`:

```
'[--ff-left:52px]', !railCollapsed && 'md:[--ff-left:220px]'
```

A phone-width screenshot with the sidebar expanded is what I would ask for alongside it. Both attached captures are desktop width, which is exactly the gap that let this through.

Non-blocking:

- Three mutations survive the new test. The one worth fixing is replacing the two literals with a computed `` `[--ff-left:${railCollapsed ? 52 : 220}px]` ``: the test still passes while the CSS rule disappears entirely, which is the precise failure your own comment warns about. Asserting on computed style rather than the class string would close it.
- The description says "Removing the prop in the layout turns it red". I reverted `layout.tsx:93` and got 1 failed, 1 passed, so one of the two goes red rather than both. Minor, but worth saying accurately.
- `notification-drawer.tsx:119` is the identical construct and keeps the bug, character for character. Its own backdrop at `:116` even carries the comment about the 52px rail. Happy to take that as a follow-up issue rather than growing this PR.
- The panel's backdrop at `:247-248` is `fixed inset-0 z-40` with no `pointer-events-none`, above the sidebar's z-30, so the rail underneath is not clickable while the panel is open. Pre-existing, and only relevant because the fix invites people to toggle the rail with the panel open.
- CHANGELOG, credit, test and screenshots are all in order.
- Ordering: #370's CHANGELOG hunk is at the identical position, so whichever lands second takes a trivial conflict. #312 and #368 both touch this file but at non-overlapping hunks.

Lennart-Pingpong:
The collapsed offset is the base rather than a branch, so below `md` the panel sits where a collapsed rail puts it whatever the rail is doing. A phone-width screenshot with the sidebar expanded is attached.

Thank you for being straight that the pre-PR panel is healthy at 430px. That is the part that made this a blocker rather than a rough edge, and it would have been easy to leave out.

**The surviving mutation.** You are right that the test did not hold the thing that matters. My first attempt at closing it asserted on the component's own source text, because a computed class produces the *identical* className at runtime and no rendering assertion can tell it from the literal — computed style included, since the rule Tailwind never emitted is absent either way. I threw that away: it pins the characters rather than the thing they are supposed to cause, and it cannot see a class that is spelled correctly and still never reaches the stylesheet.

What is in there now runs the repo's own Tailwind with the repo's own theme and asks the output — the collapsed offset in no media query, the expanded one inside `@media (min-width: 768px)` and in no unconditional rule.

**One thing you may want to know about the check we both ran.** Narrowing `content` to the panel file is not tidiness. The repo's globs cover `app/**`, which includes the test files, so the sibling test's own assertion strings (`'[--ff-left:52px]'`, `'md:[--ff-left:220px]'`) are enough to make Tailwind emit both rules on their own. Run over the whole project the check passes for a panel that has stopped asking for them — which means "I ran the repo's tailwindcss over it and the rule is emitted" cannot, by itself, distinguish the mutation from the fix. Your conclusion was right; the method could not have told you if it had not been.

Narrowed, computing the literal reddens both assertions and deleting the `md:` reddens the breakpoint one.

While measuring that: exactly two utilities in the whole bundle come only from test files (`.invert`, `.filter-none`), so as bundle noise this is nothing. It is only the check that it invalidates.

`test/setup.ts` grew one guard, `typeof window !== 'undefined'` around the matchMedia shim, because setupFiles run for every test file including one that declares `@vitest-environment node` to get Tailwind rather than a DOM.

**The description claim.** You are right, reverting `layout.tsx:93` gives 1 failed, 1 passed rather than both. Corrected in the body.

**`notification-drawer.tsx:119`** — happy to take that as a follow-up issue rather than growing this PR, as you suggested. Same for the backdrop's missing `pointer-events-none` at `:247-248`; it is pre-existing, and you are right that this fix makes it easier to run into.

ravirajsinh45 (approved):
@Lennart-Pingpong The blocker is closed, and closed in the way that makes it hard to regress.

`uploads-panel.tsx:276-277` now has the collapsed offset as the unconditional base with the 220px value gated behind `md:`. I checked that in the compiled CSS rather than in the source string: building `apps/web/app/globals.css` with the repo's real content globs emits `.\[--ff-left\:52px\]` at top level and `.md\:\[--ff-left\:220px\]` only inside `@media (min-width: 768px)`, later in the file, so it wins from 768 up on ordering alone. Then I loaded that stylesheet in headless Chrome against the panel's exact class string. At 767 CSS px, `--ff-left` computes to 52px, the panel is 52 to 432, the close button 396 to 415, and `documentElement.scrollWidth === innerWidth`. At 768 it is 220px, the panel is 220 to 600, flush against the rail's right edge, the close button 564 to 583, still no horizontal overflow. 220 + 380 = 600 fits inside 768, and `md` is the stock 768 here since `tailwind.config.ts` has no `screens` override and `plugins: []`, so there is no viewport at or above the gate where the panel overflows, and below the gate the panel's left is byte-identical to what shipped before this PR: the same single unconditional rule.

It is also not the bug moved somewhere else. Below `md` an expanded rail is still covered, but the panel is modal: its backdrop is `fixed inset-0 z-40` at `uploads-panel.tsx:248`, over the sidebar's `z-30` at `sidebar.tsx:75`, so the rail is inert underneath either way. The comment at `:268-272` states that constraint at the exact place someone would otherwise tidy the gate away.

**On the test, and on a correction I owe you.** You are right about the check we both ran, and our evidence last round did not support the conclusion it was attached to. I confirmed it: compiling with only `layout-uploads-panel-rail.test.tsx` as `content` emits both `.[--ff-left:52px]` and `.md:[--ff-left:220px]` from the test's own assertion strings, so "I ran the repo's tailwindcss over it and the rule is emitted", run over the whole project, cannot distinguish the fix from the computed-literal mutation. Narrowing `content` to the panel file is the substantive part of the new test, not tidiness. Your bundle-noise figure is exact as well: compiling with and without `**/__tests__/**` differs by exactly `.invert` and `.filter-none`.

The test itself now holds the thing that matters. `uploads-panel-offset-css.test.ts:40-43` really runs `postcss([tailwindcss({ ...config, content: [PANEL] })])` against the repo's own config, and `mediaContexts()` re-parses the output and keys each `--ff-left` declaration on its enclosing at-rule. Mutating the fix: computing the literal reddens 4 of the 5 tests, deleting `md:` reddens 3, removing `!railCollapsed &&` reddens 1, and removing `left-safe` from the panel reddens 3. One mechanism worth writing down, since it collapses two of your constraints into one: the two offsets coexist on the element only because of the `md:` prefix. `cn()` is tailwind-merge, and dropping the prefix makes twMerge delete the base outright, leaving `w-[380px] [--ff-left:220px]`.

`tsc --noEmit` is clean, the whole web suite is green at 69 files and 501 tests, and the branch rebased onto `216e0a5f` merges clean. This is ready to merge. Everything below is non-blocking. I pushed the first two to your branch as `a7d0a517` rather than send you a round for them, so they are written as reasoning rather than as a request; the rest is ours to carry.

Non-blocking:

- `CHANGELOG.md:18` ends "It now opens against whichever edge the sidebar has", with no breakpoint qualifier, but after the gate that is true only from 768px up: on a phone the panel still opens at 52px over an expanded rail. The file's own style qualifies exactly this two entries above, where the #341 entry says "Above `sm`, including a landscape phone, everything stays inline exactly as before". One clause closes it, something like "on a tablet and up; on a phone it stays where a collapsed rail puts it, because 220 + 380 does not fit". It stays under `[Unreleased]`.
- `uploads-panel-offset-css.test.ts:56` is `if (decl.value.trim() !== offset) return`, and postcss keeps `important` as a separate flag on the declaration, so writing the base as `'![--ff-left:52px]'` compiles to `--ff-left: 52px !important` unconditionally, the plain `md:` rule can never win, and the panel sits at 52px over an expanded rail at every width. I applied that mutation and the whole directory stayed green at 4 files and 15 tests. Also requiring `decl.important === false` for the collapsed offset closes it. Today's code is correct and measured correct; this is only about what the test would catch next time.
- Two small things in the body. "vitest run green (574)" does not reproduce at `2457ec00`: a full run on this tree reports 69 files and 501 tests, and `find` counts 69 test files, so nothing is being skipped. And "reddens one of its two assertions" went stale in the same push that fixed it, since the file now has three tests and four assertions; reverting `layout.tsx:93` reddens exactly one of the three.
- The follow-up issues for `notification-drawer.tsx:119` and the backdrop's missing `pointer-events-none` are ours to file, not work for this PR. Neither exists yet and we will open them.
- Ordering: #370's CHANGELOG hunk sits at the same position, so whichever of the two lands second takes a trivial conflict.

Approving and merging. Your correction of our verification method was the most useful thing in this round, and it is the reason the test is worth having.

ravirajsinh45:
@Lennart-Pingpong Both of ours are filed: #382 for the notification drawer, which carries the identical hardcoded `left-[52px]` this PR just fixed for the panel, and #383 for the backdrop still covering the rail the panel now deliberately sits beside. The pattern you landed here, and the compiled-CSS test, are referenced from #382 as the shape to copy.

Issue #356, opened 2026-09-11
The comment composer's Send button is covered by the attribution badge and cannot be clicked


ravirajsinh45:
On the asset review page, the main comment composer's **Send button cannot be clicked**. Every click lands on the floating "Powered by FreeFrame" attribution link instead and opens the repo in a new tab.

Pressing Enter still posts the comment, which is presumably why this has gone unnoticed. The button's own tooltip is "Send (Enter)".

## Measured

At 1512x805, `document.elementsFromPoint` at the Send button's own centre returns, top-most first:

```
SPAN    "Powered by FreeFrame"
A       "Powered by FreeFrame"   z-index: 20   x=1323 y=759 w=173 h=30
svg     (the send icon)
BUTTON  (the send button)        x=1468 y=765 w=28  h=28
```

The badge is `fixed` over the whole dashboard and is 173px wide against the button's 28px, so it covers it completely and wins hit-testing outright.

## Where it comes from

`apps/web/app/(dashboard)/layout.tsx`:

```jsx
<PoweredByBadge className="fixed bottom-safe right-safe [--ff-bottom:1rem] [--ff-right:1rem] z-20 ..." />
```

Introduced in #264. Not a regression from the safe-area work in #347, which only swapped `bottom-4 right-4` for the safe-area equivalent; both resolve to the same `1rem` on a desktop.

## Scope

* Affects the main composer on the asset review page.
* The nested reply composer's send button sits higher up and is unaffected.
* The asset page already renders its own centred "Powered by FreeFrame" in the header, so the floating badge is a duplicate on this route.

## Possible fixes

1. Give the badge `pointer-events-none` and put `pointer-events-auto` on the inner link only. Simplest, but the whole link still sits over the button visually.
2. Hide the floating badge on the asset review route, since the header already carries one.
3. Offset the composer or the badge so they no longer overlap.

(2) looks right given the duplication, but it is worth a moment's thought about what admins who disable branding expect to see.

PR #370, opened 2026-09-15
fix(web): stop the credit covering the asset navigator

Lennart-Pingpong:
Summary
On the asset viewer, "Powered by FreeFrame" is drawn straight across the "‹ 5 of 7 ›" navigator whenever the asset has a neighbour.

The credit is placed with absolute left-1/2 -translate-x-1/2, and the navigator is centred in the same spot by the two flex-1 blocks either side of it. Both came in separately (the credit with #203) and each is centred correctly on its own.

They now share one centre group, navigator first and credit beside it. An asset with no neighbour renders no navigator, so the credit is centred alone as before, and the badge still renders nothing when an admin turns the credit off.

Test
`app/(dashboard)/projects/[id]/assets/[assetId]/__tests__/top-bar-credit.test.tsx` mounts the real page behind mocks and asserts the credit and the navigator sit in one flex row with nothing absolutely positioned. Red at `216e0a5`, and red for the credit-renders-first, drop-`shrink-0` and drop-the-`lg`-gate mutations. Checked by: Chrome, at desktop width and narrowed to iPad-portrait width.

tsc --noEmit clean, vitest run green (574), in node:20.

before the fix
image

after the fix
 
Bildschirmfoto 2026-09-15 um 23 01 25


ravirajsinh45 (changes requested):
@Lennart-Pingpong The bug is real and materially worse than the PR claims. On main, `document.elementFromPoint` at the centre of the "next asset" chevron returns the badge's anchor, so the credit was not merely drawn across the navigator, it was winning the hit test. The diagnosis is right on mechanism and on history too: the credit did arrive with #203 as `absolute left-1/2 -translate-x-1/2`, the navigator is centred by the two `flex-1` blocks either side, and each was correct alone. Both degenerate cases are handled properly as well, since `PoweredByBadge` returns null when the credit is off and `gap-3` then has nothing to space.

Two things to fix.

**1. The credit is now an unshrinkable in-flow item, and the controls are what yield.** `page.tsx:521` moves the 147px credit out of absolute positioning into the bar's flow, growing the unshrinkable middle from 103.3px to 262.4px. The left block collapses first, and the right block at `:549` never yields: it is `shrink-0` with no `min-w-0`, so it is floored at min-content and the surplus spills past the right edge. Nothing can scroll to it, because the page root at `:495` is `absolute inset-0 flex flex-col overflow-hidden` and the shell's only scroller sits outside that clipped root.

Measured in headless Chrome against the dev server's own compiled CSS, rail collapsed:

- Asset name: 1024px viewport 330.5px visible becomes 220.2px; at 836px it is about three characters; at 800px it is gone. On main it only reached zero at 640px.
- Right-edge clipping starts at 744px instead of 596px. At 744px, iPad mini portrait, 15.8px of the 32px sidebar toggle is past the clip and `elementFromPoint` at its centre returns BODY. At 700px the toggle is entirely gone and 27px of Share with it.
- Expand the rail to 220px and every threshold moves up 168px, so a ~920px window loses the toggle where main kept it.

So a reviewer on an iPad in portrait, or with two windows side by side, cannot open or close the comment sidebar. Make the credit the thing that gives way rather than the controls: either breakpoint-gate the badge (`hidden lg:inline-flex`) or let the wrapper shrink instead of `shrink-0`.

**2. The invariant this adds is testable, and the stated reason it is not is not quite right.** The body says overlap is a layout property jsdom does not compute. True, but the thing this fix establishes is structural rather than geometric: one flex row holding the navigator and then the credit, with nothing absolutely positioned. I wrote that test to check, mounting the real default export behind mocks, and it is green here and red at `ebd47438` with "expected false to be true", and red again for both the credit-renders-first and the drop-`shrink-0` mutations. The repo has already done exactly this for the other half of this same bug: `layout-attribution-badge.test.tsx`, added by #358, mocks the shell and asserts where the credit renders. One note if you write it, the assertion has to be `credit.parentElement.contains(counter)` rather than "share a parent", because at this head the counter's parent is the inner navigator div.

As it stands, reverting the wrapper re-introduces the overlap with CI green.

Non-blocking:

- The comment at `:519-520` and the body both say the credit is "centred alone, as before" with no neighbour. That holds. With a neighbour it is the group that is centred, so the navigator itself is no longer on the bar's centre line. Worth saying in the comment, since the next reader will measure it.
- Two other places now describe the old behaviour: `layout.tsx:83-84` and `layout-attribution-badge.test.tsx:12` both justify suppressing the floating badge with "carries its own credit centred in the top bar", and the #356 CHANGELOG entry four lines above yours says the same. All three are true only for an asset with no neighbour after this change.
- The body still has `<!-- Lennart: Screenshots einfügen -->` sitting where "Checked by:" should be. Invisible on the rendered page but present in the source of a public repo, and the screenshots did get attached, so it just needs deleting.
- The asset-name span has `truncate` with no `title`, and blocker 1 makes that bite 159px earlier. Pre-existing, but a `title` would cost nothing here.
- CHANGELOG entry, section, voice and credit are all correct.

Lennart-Pingpong:
Both fixed, and the second one is a fair correction to something I asserted rather than checked.

**1. The credit was the wrong thing to make unshrinkable.** Your reading of the mechanism is right: the middle group does not shrink, the right block is `shrink-0` with no `min-w-0` so it is floored at min-content, and the page root clips what spills while the shell's only scroller sits outside that root — so there is no gesture that reaches the toggle at all.

Gated at `lg`, per your first suggestion. Measured against the repo's own compiled Tailwind, with the bar reconstructed from `page.tsx`:

| | middle group |
|---|---|
| before this PR (credit absolute) | 100px at every width |
| this PR as submitted | 244px at every width |
| with `hidden lg:inline-flex` | 100px below 1024, 244px at and above |

So below `lg` the bar is back to the flow it had before the credit joined it, which is the state your thresholds were measured against. I did not try to reproduce your 744px and 700px figures: my reconstruction's right-hand block is narrower than the real one, it did not overflow at those widths, and numbers from a rough copy of the bar would be worse than none. Yours are from the real page.

I went with the breakpoint rather than letting the wrapper shrink because "Powered by Free…" is not an attribution, and because out of flow is what the credit already was — this restores that below `lg` rather than inventing a third behaviour. It does mean no credit at all on this route below 1024px. The floating badge deliberately does not come back to fill that gap: it would land on the send button again, on exactly the widths where the composer is hardest to hit. If you would rather have the credit present and shrinking, say so and I will change it.

**2. You are right that it is testable, and I was wrong to say otherwise.** What I checked was whether the overlap could be asserted; what the fix establishes is structural, and that is a different question. The test mounts the real default export behind mocks and asserts `credit.parentElement.contains(counter)` — with your note taken, since "share a parent" does not hold: the counter is one level further down, inside the navigator's own div.

It is red at `216e0a5`, and red for each of the credit-renders-first, drop-`shrink-0` and drop-the-`lg`-gate mutations. `layout-attribution-badge.test.tsx` from #358 was the model for the mocking, as you said.

**The non-blocking ones, all taken:**

- The comment in the bar now says that with a neighbour it is the *group* that is centred, so the navigator sits a little left of the bar's centre line. That is the cost of the two not overlapping, and the next reader will measure it.
- `layout.tsx:83-84`, `layout-attribution-badge.test.tsx:12` and the #356 CHANGELOG entry all said "centred in the top bar". All three now say "in the top bar", and `layout.tsx` also records why the floating badge does not return below `lg`.
- The asset name has a `title` now. It is the only place the full name appears and it is the first thing the bar gives up.
- The CHANGELOG entry carries the hit-test detail you found, which is the part that makes this more than cosmetic.

ravirajsinh45 (approved):
@Lennart-Pingpong This is good work. Both blockers are closed, and since you measured your claims I checked each one rather than taking it on trust.

**The `lg` gate does what you say it does.** `cn` is twMerge (`apps/web/lib/utils.ts:4`), and `hidden` and `inline-flex` are both in twMerge's display group, so the component's own base `inline-flex` is dropped outright: the merged class list has no `inline-flex` left to compete with `display:none`. A `display:none` child is not a flex item either, so the `gap-3` at `page.tsx:539` contributes nothing below the gate. Below 1024px the middle group is the navigator alone, which is byte for byte the flow this page had before #203 put the credit in the bar, exactly as your table says. `lg` is the stock 1024px, since `tailwind.config.ts` defines no `screens` override. Both widths we measured last round sit below it: 744px with the rail collapsed, and about 912px with it expanded, which confirms our +168px (the rail is 52px against 220px, `sidebar.tsx:78`). Every width where we actually demonstrated the failure is now credit free.

You were also right not to reproduce the 744px and 700px figures from a reconstruction. Numbers from a rough copy of the bar would have been worse than none.

**The test is real, and every mutation you list is red.** I ran them one at a time: `page.tsx` restored from `216e0a5f` gives 4 failures, the credit rendered before the navigator gives 1, dropping `shrink-0` from the group gives 1, dropping `hidden lg:inline-flex` gives 1. The assertion is `credit.parentElement!.contains(counter)`, which is the shape that holds at this head. The whole web suite passes at `c8abfb61`, and the new file is picked up by the default glob despite `(dashboard)`/`[id]`/`[assetId]` in its path, so CI genuinely runs it.

**No regression on the other half of the bug.** The diff against `layout.tsx` touches only the comment block. `isAssetViewer` and the gate on the floating badge are unchanged, and the four tests in `layout-attribution-badge.test.tsx` still pass. Declining to un-gate the floating badge below `lg` to fill the new gap is the right call: that is precisely what #358 removed. The viewer's own copy is in normal flow with nothing absolute or fixed, so it cannot overlap anything.

All four non-blocking items from last round are done: the comment now says it is the group that is centred and what that costs, the three stale "centred in the top bar" descriptions are corrected, the PR body's leftover comment is gone, and the asset name carries a `title`.

Nothing blocking. Three small things. I pushed the first two to your branch as `a196edfd` rather than send you a round for them, so what follows is the reasoning rather than a request. The third is a question for you.

Non-blocking:

1. **`CHANGELOG.md:18-19` carries the #370 entry twice.** `1c13cb57` added a credited copy above the original instead of amending it in place, and `c8abfb61` then updated only line 18. Line 19 is the pre-gate wording, with no PR number and no credit, so it now also describes behaviour the PR no longer has. Deleting line 19 fixes it. Keep the surviving entry where it is, under `[Unreleased]`.

2. **The assertion guarding the gate is a substring match, so the blocker can come back green.** `top-bar-credit.test.tsx:176` is `expect(credit.className).toContain('hidden')`, and `overflow-hidden` satisfies that. Change `page.tsx:563` to `className="overflow-hidden lg:inline-flex"` and twMerge keeps the base `inline-flex` (I ran it: the merged output is `inline-flex … overflow-hidden lg:inline-flex`), so the credit is `display:inline-flex` at every width and blocker 1 is fully restored with the test still passing. `expect(credit.className.split(/\s+/)).toContain('hidden')` closes it. Line 177 is fine as is, nothing else contains `lg:inline-flex`.

3. **The gate is keyed to viewport width while the constraint is bar width. That is a limitation of what we suggested, not of what you did.** The mechanism survives above `lg`: `page.tsx:567` is still `shrink-0 flex-1 justify-end` with no `min-w-0`, so the right block is floored at min-content and the surplus still leaves the bar past the clipped edge. We have not measured this and cannot right now, so treat what follows as arithmetic over our round-1 numbers rather than a finding. Those put the expanded-rail threshold at about 912px, leaving roughly 112px of headroom to the 1024px gate. The conditional Compare button at `:586-605` sits in that right block and looks worth 95px to 105px of min-content, which would move the threshold to about 1010px. Whether the real headroom is 112px or nearer 10px depends on whether the asset we measured already had Compare rendered, which we cannot recover. One measurement at 1024px with the rail expanded and Compare present settles it. If it is tight, `min-w-0` on the middle group at `:539`, letting the credit be what gives way, removes the class of failure rather than sitting above it.

Approving and merging. Thanks for the measurements, they made this a short review.

ravirajsinh45:
@Lennart-Pingpong Filed #384 for the third item, the one neither of us could settle without a measurement: the gate is keyed to viewport width while the constraint is bar width, and whether the headroom to 1024px is ~112px or nearer 10px depends on whether the asset we measured had Compare rendered. The issue says what single measurement settles it.

Issue #371, opened 2026-09-16
Skip re-encoding a source that is already browser-safe, when the ladder asks for its own size


Lennart-Pingpong:
## Where this comes from

We run FreeFrame self-hosted to review cuts with clients. The material is episodes exported from DaVinci Resolve — H.264 8-bit, 30 to 40 minutes, around 4 GB each — reviewed at full size on a desktop, because the point of the round is to catch a jump in a grade or a soft frame. There is no GPU in the box and the worker shares 16 cores with everything else the machine runs, so it gets 6.

Two things follow, and both are why this has been a patch on our instance for months:

**An hour of encoding sat between "uploaded" and "watchable".** Three episodes in an evening kept the box busy until the next morning and the client got the link the next day. Copied, the same three are ready in minutes — the difference between a review round inside one working day and one across two.

**The re-encode was actively unhelpful for what the tool is for.** Our reviewers are looking at the picture itself. A rendition that has been decoded and encoded once more carries artefacts the master does not, and an editor cannot tell ours from theirs, so a round can end with someone re-exporting to chase something the review copy introduced. Copying the stream removes the question: what is being judged is the file that was uploaded.

Storage is the third and smallest reason — three renditions of a library of masters is a second copy of it on paid object storage, for rungs nobody here ever plays. #325 settled that half; this is the CPU half.

## Problem

Since #325 a deployment can say how many rungs to build, and a single rung at or above the source resolution resolves to exactly one rendition at the source's own size. That is the shape a review instance wants: reviewers judge finished cuts at full size, and the smaller rungs cost encode time and storage that nobody plays.

What that rendition still costs is a full re-encode — and on this kind of instance the source it is encoded from is nearly always **already the target format**. A master exported from an NLE is H.264 8-bit 4:2:0, which is exactly what every browser plays. The encoder is being asked for a slightly worse copy of what it was handed, at the price of the entire transcode.

Measured on a real master (36 min, 1080p, 3.92 GB) on a 16-core box with the worker capped at 6 cores:

| | wall clock |
|---|---|
| stream copy + HLS segmentation | **355 s** (137 s of which is the upload) |
| re-encode, one rendition, `medium` CRF 20 | **~55 min** |

The output of the first is the master's own picture. The output of the second is that picture decoded and re-encoded once more.

## Proposed solution

[...]

Issue #350, opened 2026-09-10
Desktop/offline review client with selective media sync


loscarinos:
### Problem

FreeFrame currently works very well as a browser-based review system, but reviewing depends on having a working connection to the FreeFrame server.

For professional film and audio workflows this can be limiting. Review often happens while travelling, on set, in screening rooms with unreliable internet, or simply in situations where downloading a video once would be more practical than repeatedly streaming it.

A useful workflow would be:

1. While online, select a project, asset or specific version and choose "Make available offline".
2. The media is stored securely on the local computer.
3. The user can review it without a network connection and create timecoded comments, annotations and possibly approvals.
4. When connectivity returns, those review changes are synchronized with FreeFrame.

The main goal is offline review, not a complete Dropbox/Nextcloud-style filesystem replacement.

### Proposed solution

Consider an optional desktop client, or another supported local-client mechanism, that provides selective offline media synchronization.

A first implementation could deliberately stay small:

- Allow selected projects/assets/versions to be marked "Available offline".
- Download and securely cache the required review media locally.
- Show clearly which versions are available offline and how much local storage they use.
- Allow normal playback and timecoded comments while offline.
- Queue comments, annotations, approvals and other supported review metadata locally.
- Synchronize those changes automatically when the FreeFrame server becomes reachable again.
- Detect when a newer media version exists and offer to download/update the offline copy rather than silently replacing it.
- Allow the user to remove local copies without deleting anything from FreeFrame.
- Preserve FreeFrame permissions: losing server access should prevent further synchronization, and locally cached confidential media should be treated as sensitive data.

This would already cover the core offline-review workflow without requiring a full filesystem mirror.

### Possible later extension

A later, separate enhancement could add an optional watched/synced local folder for review exports.

Example:

1. A user exports `Film_v12.mp4` into a designated local folder.
2. The FreeFrame client detects the new file.
3. It uploads it to the selected project/asset, ideally as a new version.
4. Uploads can continue or retry in the background.

That could provide a Nextcloud/Dropbox-like convenience for production workflows, but it does not need to be part of the initial offline-review implementation.

The existing resumable-upload work discussed in #241 could potentially complement such a client, but offline review should not depend on implementing full two-way filesystem sync.

### Browser caching / PWA only

A service worker could cache some web assets and media, but browser storage limits, eviction behavior and restricted local-file access make this less predictable for multi-GB professional video projects. Offline comment synchronization would still need dedicated application logic.

[...]

Issue #333, opened 2026-09-09
Skip re-encoding a source that is already browser-safe


ravirajsinh45:
Second half of #319, reported by @Lennart-Pingpong. The configurable ladder shipped in #325.

Every upload is re-encoded even when the source could be served as it is. On the 36-minute master in the original report that is the difference between minutes and roughly an hour of encode time, plus the storage for renditions nobody needed.

Deciding a source is "browser-safe" needs more than a codec check:

- H.264 in a profile and level the browser will decode
- `yuv420p` pixel format
- a `moov` atom the player can reach without downloading the whole file
- an audio stream the HLS segmenter can cut cleanly

Related: #328 asks for configurable audio codec and bitrate, and the copy case overlaps with it. #201 covers the ladder upscaling that #325 already closed off.

PR #372, opened 2026-09-16
feat(transcoder): package a browser-safe source without re-encoding it

Lennart-Pingpong:
Refs #333 

## Summary

`TRANSCODER_SOURCE_COPY=true` (default off) lets a deployment skip the encoder where it has nothing to decide: the ladder has already resolved to a single rendition at the source's own size, and the upload is already H.264 8-bit 4:2:0 — what every browser plays, and what an NLE exports by default. The file is remuxed into HLS with `-c:v copy` instead: the same frames, cut into segments at the keyframes they already carry.

Measured on a 36-minute 1080p master on six CPU cores: **355 s against about 55 minutes**, and what the reviewer gets is the master's own picture rather than a re-encode of it.

This is deliberately not a new policy. It rides on #325: without a ladder that already asks for one full-size rendition, it never engages, so it changes what a transcode *costs* and never what it produces.

## Where this comes from

We run FreeFrame self-hosted to review cuts with clients. The material is episodes exported from DaVinci Resolve — H.264 8-bit, 30 to 40 minutes, around 4 GB each — and they are reviewed at full size on a desktop, because the point of the round is to catch things like a jump in a grade or a soft frame. There is no GPU in the box, and the worker shares 16 cores with everything else the machine does, so it runs with 6.

Two things follow from that, and both are why this patch existed on our instance long before it was offered here:

**An hour of encoding sat between "uploaded" and "watchable".** Three episodes in an evening meant the box was busy until the next morning, and the client got the link the next day. With the copy path the same three are ready in minutes, which is the difference between a review round inside one working day and one across two.

**The re-encode was actively unhelpful for what we use the tool for.** Our reviewers are looking at the picture itself. A rendition that has been decoded and encoded once more carries artefacts the master does not, and an editor cannot tell ours from theirs — so a review round can end in someone re-exporting to chase something the review copy introduced. Copying the stream removes that question entirely: what is being judged is the file that was uploaded.

Storage is the third, smaller reason. Three renditions of a library of masters is a second copy of the library on paid object storage, for rungs that nobody on this instance ever plays. #325 already fixed that half; this one takes the CPU half.

None of that argues the ladder is wrong — it is right for an instance whose reviewers watch on mobile data, which is why the default is untouched and the setting is off unless someone turns it on.

## What it does

`TRANSCODER_SOURCE_COPY=true` remuxes instead of encoding, and only when every one of these holds:

- **The ladder resolved to one rendition at the source's own dimensions** — #325's clamped case, or a rung that matches the source exactly. A configured ladder gets its rungs; a rung below the source is a request for a smaller rendition and is encoded.
- **The video stream is H.264 8-bit 4:2:0.** The pixel format is checked, not just the codec name: High 10 and High 4:2:2 report `h264` too and play in no browser, and Resolve exports High 10 if you let it. Copying one of those would produce an asset that is black for every reviewer, which is the failure mode worth being paranoid about here.
- **The source is SDR.** HDR and Dolby Vision are what the filter graph exists for and take the encoding path unchanged.

Audio is separate: AAC is copied with the video, anything else — PCM is common out of an NLE — is encoded to AAC, which costs seconds and leaves the picture untouched.

## Decisions worth your attention

**It hangs off the ladder rather than standing alone.** An earlier draft simply said "copy whenever the source is browser-safe". That would override a deployment's configured rungs, which is a policy decision the operator did not make. Tying it to the single-full-size-rendition case makes it a strict optimisation of a configuration you already support: same output, cheaper route.

**A copied version records no download MP4.** #350 added one because the ladder is segments and the only single file a download had was the camera master, often an order of magnitude larger than what was watched. For a copied version that gap does not exist — the rendition is the uploaded stream, so the master the download falls back to is the same picture at the same bitrate. Building it anyway would write a third copy of the file and spend exactly the storage this setting was turned on to save. The fallback in `assets.py` and `share.py` already handles a version without the rung, so nothing else changed.

**`-hls_time` becomes a floor.** Segments can only start on a keyframe and a copy cannot make new ones, so a master with a long GOP gets longer segments. Players handle it; the alternative is the re-encode.

**No hardware fallback on this path.** A copy runs no encoder and touches no device, so there is nothing for `_is_hw_runtime_failure` to degrade from; anything that stops it is about the input and would stop a second attempt too.

## Changes

- `packages/transcoder/ffmpeg_transcoder.py`: `source_copy_enabled()` and `is_browser_safe()` next to the other `TRANSCODER_*` readers; one `copy_source` boolean built from values `transcode()` already has; `_build_copy_cmd()` as a sibling of `_build_ffmpeg_cmd()`, reached through the same attempt loop so progress reporting, timeouts, upload, thumbnail and the result are shared.
- `.env.example`: the setting, next to `TRANSCODER_QUALITIES`, with what it trades.
- `CHANGELOG.md`: one entry under `## [Unreleased] / ### Added`.

No new module, no schema change, no new dependency, and the encoding path is untouched.

## Testing

- [x] Backend: 487 passed, including 12 new ones in `apps/api/tests/test_source_copy.py`. They read the ffmpeg command that is actually built, since both mistakes possible here are silent: the setting doing nothing, and the worse one — copying something that should have been encoded.
- [x] Each of the five conditions removed in turn, each time a named test went red: the switch, the ladder condition, the pixel-format check, the HDR exclusion, and the download-MP4 exception.
- [x] Run against real S3 (Hetzner `nbg1`) and real ffmpeg 7.1 with generated masters, four cases: H.264 + AAC copied, H.264 + PCM copied with the audio converted, High 10 encoded, and a three-rung ladder encoded. For both copy cases the **decoded frames of source and result match exactly** (`ffmpeg -f framemd5`), which is the check that proves a copy copied — remuxing to MPEG-TS rewrites SPS/PPS before every keyframe, so comparing the bytes of the elementary stream would report a difference for a file where no picture was touched.
- [x] End to end on our dev instance, uploaded through the browser: a 90-second 1080p H.264 master, `TRANSCODER_QUALITIES=1080p` and the setting on. The whole transcode task took **2.6 s**, the log line names the reason (`already browser-safe (h264/yuv420p) and the ladder asks for 1920x1080: remuxing, not re-encoding`), the version went `processing` to `ready` within one poll, `s3_key_download` is NULL as intended, and the asset plays in the review screen through hls.js at 1920x1080 with no console error.

## Checklist

- [x] `CHANGELOG.md` entry under `## [Unreleased]`, with `(#371 by @Lennart-Pingpong)`.
- [x] `.env.example` documents the setting and its trade-off.
- [x] Default behaviour unchanged: with the setting unset, the built command is byte-for-byte what it was.


ravirajsinh45 (changes requested):
@Lennart-Pingpong This is a strong piece of work and the reasoning behind it is the right shape. Tying the copy to the case where the ladder has already resolved to one rendition at source size is what keeps it an optimisation rather than a policy change, and I agree with that call. I also ran your mutation claim rather than trusting it: each of the five conditions you named, removed in turn against the full backend suite with live Postgres, does take a named test red. That held exactly as described.

Two blockers, both in the same gate at `ffmpeg_transcoder.py:800-806`, and both are the failure class you said you were being paranoid about: a source that passes every condition and comes out wrong.

**1. Rotation is never checked, and a stream copy cannot carry it.** A phone or DSLR clip shot in portrait is H.264 8-bit 4:2:0 SDR at its own size. The orientation lives in the container as a display matrix, not in the stream, and `is_browser_safe` reads only `codec_name` and `pix_fmt`.

Run on a real rotated file with ffmpeg 7.1.5, source 640x480 with `rotation=90`:

| path | output | rotation side data |
|---|---|---|
| encode, today's `filter_complex` | 360x480, upright | baked into the pixels |
| `_build_copy_cmd` | 640x480, sideways | gone |

The encode path gets this right by accident: ffmpeg applies autorotate whenever a filter graph is present, and the ladder always has one. `-c:v copy` has no filter graph, and MPEG-TS carries no display matrix, so the metadata that would have told the player is dropped by the muxer. The master playlist advertises the coded resolution too. I grepped for rotation, displaymatrix, autorotate and transpose across the whole transcoder and the web review components: zero hits. Nothing compensates downstream, and the thumbnail is generated through a filter so it comes out upright, which means a sideways video sits next to an upright poster frame.

**2. Nothing checks that the sync samples are IDRs, so an open-GOP master produces unseekable segments.** Reproduced with a source that passes `is_browser_safe` cleanly:

```
ffmpeg -f lavfi -i testsrc2=...:duration=12 -c:v libx264 -pix_fmt yuv420p \
       -g 50 -x264opts open-gop=1:bframes=3:scenecut=0 open.mp4
```

Run through `_build_copy_cmd` verbatim it yields 6 segments. seg_000 carries SPS, PPS and an IDR; seg_001 through seg_005 carry none, and each one decodes to `non-existing PPS 0 referenced`. A closed-GOP control through the identical command is clean in all six.

This one is nastier than it looks because it does not fail loudly. The parameter sets live in segment 0, so playback from the top works and the job reports success. What breaks is seeking: clicking a comment timecode, or the scrubber's hover preview, lands the player on a fragment with neither the decoder config nor a random-access point. In a review tool, seeking is the product.

Your framemd5 check could not have caught either one, since it decodes from the top. That is worth knowing rather than a criticism of the method.

One correction to something I nearly told you: `-hls_flags independent_segments` is a symptom here, not the cause. Removing it changes nothing. `-bsf:v dump_extra` is not the workaround either, since it collides with the auto-inserted `h264_mp4toannexb` and the segment still starts on a non-IDR recovery point. The fix belongs in the gate: refuse the copy when the stream carries a non-zero display-matrix rotation, and when a keyframe probe shows sync samples that are not IDRs. Both fail safe toward encoding.

The text needs to follow the fix. The comment at `:314-317` ("changes what a transcode costs, never what it produces"), `.env.example:152-156` ("Same picture") and the CHANGELOG entry ("the result is the master's own picture") all enumerate exclusions, and rotation belongs in each.

Non-blocking, worth a look while you are in there:

- Segment length is unbounded and follows the source GOP, with no probe and no cap. The encode path bounds it with `-force_key_frames`. Same root cause as blocker 2, so one keyframe probe in the gate settles both.
- Two conditions survive deletion with the suite green: `and dovi_profile is None`, and excluding `copy` from the hardware fallback list. The first is probably redundant since DV is always 10-bit, so `pix_fmt` catches it, but it is worth a line saying so. The second is a real design question: the comment argues anything that stops a copy would stop a second attempt, and that is not true of remux-specific failures, where an encode would have succeeded. With the setting on, those become a failed asset rather than a slow one.
- A non-16:9 source with a single rung silently never copies. `:803` compares the rung's nominal `w:h` string against the source's, but the encoder applies it with `force_original_aspect_ratio=decrease`, so a rung can resolve to the source's own size without the strings matching. A 1440x1080 source with `["1080p"]` does not copy.
- The copy path's whole HLS output block is unasserted. Deleting `-master_pl_name master.m3u8` leaves all 12 green, although both routers hand the player `master.m3u8`.
- `TRANSCODER_SOURCE_COPY` is missing from the env-var table in `docs/deployment.md`, which is where a CPU-only operator looks when an hour per master is the problem this setting solves.
- On the issue link: `Refs #333` rather than `Closes #371` would be better. #333 is our own split of #319 and is wider than what this implements, and #371 restates it. We will close #371 as the duplicate, so please point the CHANGELOG credit at #333 and #372.
- The CHANGELOG entry omits its own PR number; the house form is `(#333, #372 by @Lennart-Pingpong)`.

None of this touches the core idea, which is sound, and the measured win is worth having.

Lennart-Pingpong:
Both blockers were real and both are fixed. I reproduced each before touching anything, because the whole argument for this setting is that the output is the master's own picture, and neither of these was.

**Rotation.** Your measurement reproduces exactly. A 640x480 source carrying `rotation=90`, run through `_build_copy_cmd` verbatim: out comes 640x480 with no side data, sideways. The same source through the ladder: 360x480, upright. So the gate now reads the display matrix — `side_data_list[].rotation`, which our existing `-show_streams` probe was already returning and nobody was looking at, plus the older `tags.rotate` spelling — and any non-zero rotation encodes.

**Open GOP.** Also reproduces, with your command. Copied verbatim it yields 6 segments, of which `seg_001` through `seg_005` each decode to `non-existing PPS 0 referenced`; a closed-GOP control through the identical command is clean in all six. You are right that this is the nastier of the two: the job succeeds, the asset looks fine, and what is broken is the thing a reviewer does most.

**One probe settles both**, as you said it would. Before anything is copied, bounded to the first 30s of the source:

- the sync-sample timestamps come from `ffprobe -read_intervals`, and
- whether they are IDRs comes from `-bsf:v trace_headers`, because that is the only place the distinction is visible: ffprobe reports an open-GOP recovery point as `key_frame=1`, `pict_type=I`, exactly as it reports a real IDR.

A copy is refused if fewer than two sync samples turn up in the window, if they are more than 12s apart, if any recovery point is present, or if the IDR count does not cover the sync samples. Both of the last two are needed and neither subsumes the other: a frame split into several slices produces one IDR NAL per slice, so an open-GOP master with a single real IDR frame can still out-count its sync samples. There is a test for precisely that, since without it the recovery-point check survived deletion.

The 12s bound is the segment-length cap you asked for — same probe, no extra read. The window is bounded rather than whole-file because a full pass over a 4 GB master is a second download; measured, the probe costs 0.25s on a 22 MB file and 0.41s on a 1.7 GB one. An encoder does not change its GOP structure halfway through a file, and a first window with only one sync sample is refused anyway rather than guessed at. Everything the probe cannot answer refuses the copy: a refusal costs encode time, which is what this setting exists to save, and the other direction costs an asset nobody can seek in.

## The non-blocking ones

**Non-16:9 never copied, and you were right about why.** The rung is applied with `force_original_aspect_ratio=decrease`, so what comes out is the source's own size whenever the scale factor is exactly 1 — which the nominal `w:h` string does not show. It is now stated as a factor rather than a string: the box has to be at least the source in both directions and exactly its size in one, so there is no float comparison and no rounding to match against `force_divisible_by=2`, which cannot bite when nothing is scaled. 1440x1080 and 1920x804 copy now; 1998x1080 still encodes, because it genuinely does not fit inside 1920x1080. This was a false negative rather than a wrong rendition, but it hid the setting from exactly the material it was written for, so thank you for catching it.

**The hardware-fallback exclusion: you are right and I have changed it.** My comment argued that anything stopping a copy would stop a second attempt, and that is not true of remux-specific failures. A failed remux now falls back to the encoder rather than failing the asset, and unlike the hardware case any error is reason enough, since a remux has no environmental failures to tell apart from input ones. The download rung follows the backend that actually ran rather than the gate's decision, so a fallback produces an ordinary ladder with an ordinary `download.mp4`.

**`dovi_profile is None`** is indeed redundant — every DV profile is 10-bit and none reach `yuv420p`, so the pixel format already catches it. Kept with a line saying so, because it is the condition a reader looks for and because "DV is always 10-bit" is a fact about the format rather than about this code.

**The HLS output block** is asserted now: `-master_pl_name master.m3u8`, `-f hls`, the playlist type and the segment type. Deleting any of them goes red.

**`docs/deployment.md`** has `TRANSCODER_SOURCE_COPY` in the env table now, and `TRANSCODER_QUALITIES` with it — that one was missing too, which was mine from #325.

**`Refs #333`** in the commit and in the PR text, and the CHANGELOG credit is `(#333, #372 by @Lennart-Pingpong)`.

**The text follows the fix.** The comment at the switch, `.env.example` and the CHANGELOG entry each name rotation and the keyframe requirement among the exclusions, and `.env.example` also states the bitrate trade, which it had not.

## Testing

- 23 tests in `test_source_copy.py`, 498 in the backend suite.
- Each of the nine conditions removed in turn; each time a named test went red. The recovery-point check survived the first pass — it was covered only by the IDR count — so the multi-slice test above was written for it.
- Against real ffmpeg 7.1.5 and real S3, four masters end to end through the actual transcoder: closed-GOP 1080p copies (framemd5 of the decoded frames matches the source, every segment decodes standalone), open-GOP encodes, rotated encodes and comes out upright, and 4:3 copies — which it never did before. The standalone-segment check is the one framemd5 could not give, exactly as you said, because framemd5 decodes from the top.

## One thing found while testing, not this PR's

A rotated source comes out of the ladder at 360x480 from a 640x480 master, not 480x640. The probe reports coded dimensions, so the clamp builds its rung against 640x480 while the filter graph has already autorotated to 480x640, and `decrease` then scales by 0.75. It is upright and it plays, so it is cosmetic, and it is on `main` today with nothing to do with this setting — your own measurement of the encode path shows the same 360x480. Happy to file it separately if you want it.

## Rebase

Onto `216e0a5`. The only conflict was `CHANGELOG.md`, where #363 removed the sentence about a share link being an uncovered surface from the #327 entry; I took your line. `#371`'s credit commit is dropped in favour of the `#333, #372` one.

ravirajsinh45 (approved):
@Lennart-Pingpong Both blockers are fixed, and I checked the fixes against real files rather than taking the report on trust.

Rotation: `display_rotation()` at `ffmpeg_transcoder.py:347-377` reads `side_data_list[].rotation` with the `tags.rotate` fallback, and the gate consumes it at `:999`. I verified the premise as well as the code, because the claim rests on a field nobody was reading: `ffprobe -show_streams -select_streams v:0` on a file built with `-display_rotation:v:0 90` does return `"rotation": 90` as a numeric side-data entry on 7.1.1, so the probe you already run really was carrying it. `abs(x) % 360` handles -90, 270 and 360 correctly, and a mirror matrix reports 180 rather than 0, so those encode too.

Open GOP: `_copy_keyframe_refusal` at `:617-716` fires on the exact source from our review. I rebuilt both masters with our own command and ran your two probe commands verbatim: the open-GOP file reports recovery points and is refused at `:707`, the closed-GOP control reports none and is allowed. Your point that the recovery-point check and the IDR count are independent, and that a multi-slice frame is why, holds.

The rest of the list checked out too. `rung_is_source_size` at `:380-411` states scale-factor-1 integrally, so 1440x1080 and 1920x804 copy while 1998x1080 still encodes, and a malformed target returns False toward encoding. The fallback at `:1249-1253` and `:1273-1276` makes a remux failure a slow asset rather than a dead one, and `:1302` keys the download rung on the backend that actually ran, so a fallback still produces an ordinary `download.mp4`. `.env.example:152-163`, `CHANGELOG.md:11` and the comment at `:311-321` each name rotation and the keyframe rule, the CHANGELOG sits under `[Unreleased]` with `(#333, #372 by @Lennart-Pingpong)`, `docs/deployment.md:276-277` has both env vars, and grep finds no `#371` left anywhere in the tree.

One thing we got wrong last round, and it is worth naming because you built to it. We told you "one keyframe probe in the gate settles both", meaning the IDR question and the segment-length cap. That was too neat. A single bounded probe answers both questions only for the window it reads, and two of the notes below are the seam that framing left: the cap measures the gaps inside the window but not the one running off its end, and the IDR answer is a statement about the first 30s rather than about the file. The bound itself is a reasonable trade and you argued it with measurements, so this is our framing being loose, not your implementation drifting from it.

Nothing blocking. Merge when you are ready.

Non-blocking, all measured on 7.1.1:

- `:702` counts every line containing `nal_unit_type: 5(IDR)`, and `avformat_find_stream_info` logs one from its own h264 context before `trace_headers` ever runs. My closed-GOP mp4: 6 sync samples, 7 IDR lines, 1 of them `[h264 @ ...]` and 6 `[trace_headers @ ...]`. So the backstop at `:713` tolerates exactly one non-IDR sync sample, which is the shape `test_a_sync_sample_that_is_not_an_idr_is_encoded_without_a_recovery_point` exists for; the fixture cannot see it because it emits only trace_headers-shaped lines. Matching `[trace_headers @` closes it in one line.
- `:675` takes `widest_gap` pairwise over the in-window list, so the distance from the last sync sample to the end of the window is never measured. A file with sync samples at 0.0, 0.52 and 40.0 reports a 0.52s widest gap, passes the 12s cap, and the copy command then writes a single `#EXTINF:40.000000`. Folding the tail in, clamped by `meta.duration_seconds` so a short source is not false-refused, closes it with no extra read.
- `:650` and `:689` bound both reads to 30s, so a master whose GOP structure changes after that window passes. I joined a 35s closed-GOP clip and a 20s open-GOP clip with `-c copy`: the window reports 15 sync samples, a 2.0s widest gap, 17 IDRs and 0 recovery points, while the whole file reports 9 recovery points. The copy yields 27 segments, of which seg_000 to seg_017 decode clean and seg_018 onward each give `non-existing PPS 0 referenced`. ffmpeg exits 0, so the new fallback never fires. Your premise holds for anything a single encoder produced; it is a lossless join of parts from different encoders, a slate prepended to a master or a stitched delivery, that breaks it. Given the setting is opt-in and the bound is stated in the docstring and the CHANGELOG, I am not blocking on it, but `ffprobe -read_intervals "%+10,40%+10"` does return both windows and an `-ss` seek stays a range read, so a second sample looks cheap if you want it.
- `:369` and `:376` return 0.0 when a rotation value is present but unparseable, and the gate reads that as upright. It is the only unanswerable case in this feature that falls toward copying rather than encoding, which is the opposite of what `:318-320` says it does. Not reachable with ffprobe 7.1.1, so robustness rather than a live bug.
- The PR body did not follow the fixes. It still says there is no hardware fallback on this path, which `:1249-1253` now contradicts, and still describes longer segments as something players handle with no mention of the 12s refusal. Its list of conditions names three where the code, `.env.example` and the CHANGELOG all name five.
- One for the docstring: our production image is `python:3.11-slim` plus `apt-get install ffmpeg`, which is bookworm's 5.1.x, not the 7.1.5 you validated on. The gate reads two ffmpeg log-line spellings rather than any stable interface, and if either ever moves the failure is silent in one direction: lose the IDR line and everything is refused, which is safe; lose the recovery line and only the count above stands in the way. I could not test 5.1 here, so I am flagging it rather than claiming it breaks.

Approving and merging. The six notes above are ours to file as issues rather than yours to carry in this PR, and we will open them against #333 so the window bound and the two probe details sit with the rest of that work.

ravirajsinh45:
@Lennart-Pingpong The six notes are filed, all against #333 as said: #377 (the IDR backstop tolerating one non-IDR sync sample), #378 (the segment cap never measuring the gap off the end of its window), #379 (the 30s probe window), #380 (an unparseable rotation falling toward copy) and #381 (the gate parsing ffmpeg log lines against a production image that ships 5.1.x). Each carries the measurement it came from, so none of it needs re-deriving.

PR #390, opened 2026-09-18
feat(transcoder): let a deployment cap the CPU one transcode takes

### LINKED ISSUE #375: Configurable CPU budget for transcodes — one encode takes 10 of 16 cores today
### Problem

#325 made the ladder configurable, which was the right half of the problem: an instance can now decide *how many* renditions to build. What it still cannot decide is *how much of the machine* building them may take.

On a 16-core host one transcode occupies about 10.6 cores with the default three rungs — and 9.5 with a single rung, so trimming the ladder barely helps:

| | wall | CPU time | cores occupied |
|---|---|---|---|
| 3 rungs, as shipped | 6.3 s | 66.9 s | **10.6** |
| 1 rung, as shipped | 4.7 s | 45.0 s | **9.5** |

For a self-host that is the whole machine. FreeFrame sits next to a database, a file server and whatever else the box does, and for the length of a transcode all of them are visibly slower.

What an operator wants to say here is one sentence: **this is a background job, it must not disturb the server.** A review instance is not a render farm — nobody is waiting on the encode with a stopwatch, and finishing in twenty minutes instead of six is a fine trade for a machine that stays responsive throughout. That sentence is currently not expressible.

(Measured on a 16-vCPU AMD EPYC, no usable GPU: 20 s of 1080p30 from `testsrc2`, `libx264 -preset fast`, CRF 20/22/26 — the transcoder's own parameters.)

### Proposed solution

One setting, accepted in the two shapes people actually think in:

```sh
TRANSCODER_CPU_LIMIT=            # unset — today's behaviour, unchanged
TRANSCODER_CPU_LIMIT=6           # six cores
TRANSCODER_CPU_LIMIT=50%         # half of what this process can use
```

The percentage form matters more than it looks. Someone self-hosting on a VPS often does not think in core counts, but does know "FreeFrame may have half the machine, and the rest must stay responsive". A share also survives moving the instance to a different box, where an absolute number quietly means something else. Note "what this process can use", not "what the host has": if a deployment already caps the worker with `cpus:`, a share has to be read against that cap, or 50% on a 4-of-16 worker would hand out 8.

Turning the budget into ffmpeg arguments, measured rather than assumed:

| | wall | CPU time | cores occupied |
|---|---|---|---|
| 3 rungs, unbounded | 6.3 s | 66.9 s | 10.6 |
| 3 rungs, `-threads:v:i 2` | 13.6 s | 66.5 s | 4.9 |
| 3 rungs, `-threads:v:i 1` + filter 1 | 29.2 s | 61.6 s | 2.1 |
| 1 rung, unbounded | 4.7 s | 45.0 s | 9.5 |
| 1 rung, `-threads:v:i 6` + filter 2 | 9.2 s | 45.5 s | 5.0 |
| 1 rung, `-threads:v:i 2` + filter 1 | 15.0 s | 43.3 s | 2.9 |

Note the third column: **total CPU time barely moves.** A budget buys wall-clock time, it does not waste work. That is the property that makes this safe to hand to an operator.

The budget goes to the encoders, split across the rungs; the filter graph gets one thread. That split is measured too — with the encoders held at two threads each, capping the graph made no difference at all:

| `-filter_complex_threads` | unbounded | 4 | 2 | 1 |
|---|---|---|---|---|
| wall | 13.5 s | 13.7 s | 13.6 s | 13.7 s |
| cores | 4.8 | 4.7 | 4.7 | 4.7 |

Scaling is not where the time goes, so leaving it unbounded would only put back the one pool the setting exists to bound.

One honest caveat that belongs in the docs rather than being hidden: *n* rungs always need at least *n* encoder threads, so a budget below the rung count cannot be met exactly. The transcoder should say so once in the log instead of pretending. And measured occupancy lands a little under the nominal budget (a nominal 6 gave 4.9–5.0), because the encoders do not all saturate at once — so this is a ceiling that is honoured, not a target that gets filled.

Two things I have **not** measured and would not claim: the hardware backends (nvenc/vaapi — no such hardware here), where the CPU cost sits in decode and the filter graph rather than the encoder; and whether `TRANSCODING_CONCURRENCY` should interact with this, i.e. whether the budget is per transcode or per worker. My instinct is per transcode, because that is what an operator can reason about, but it is your call — as is the name.

Happy to open the PR once you have a view on those two.

### Alternatives considered

**A cgroup cap (`cpus:` in compose) — what we do today, and measurably worse.** Same test, same resulting occupancy of 4.9 cores:

| | wall | CPU time | cores occupied |
|---|---|---|---|
| `--cpus=5`, threads unbounded | 16.2 s | **80.2 s** | 4.9 |
| no cgroup cap, `-threads:v:i 2` | 13.6 s | **66.8 s** | 4.9 |
| both together | 14.4 s | 67.3 s | 4.7 |

Throttling eleven threads into five cores' worth of quota costs **20% more CPU time and 19% more wall time** than simply running fewer threads — the difference is context switching on threads that are scheduled only to be stopped again. It is also deployment-level: not in the shipped compose file, absent for a bare-metal or systemd install, and not settable at all on some managed platforms. A setting FreeFrame owns works everywhere FreeFrame runs. The two are not exclusive, though, and the third row shows they compose harmlessly.

**A global `-threads` before `-i` — tried, does nothing.** The natural first attempt. It is an input option and reaches the decoder; the encoders never see it:

| `-threads` before `-i` | 1 | 2 | 4 | 6 | 8 | absent |
|---|---|---|---|---|---|---|
| cores occupied | 11.1 | 11.0 | 11.3 | 11.2 | 11.3 | 11.2 |

**A slower preset.** Changes how much work there is, not how many threads do it — x264 derives its thread count from the machine, not from the preset. It also trades picture quality for something that was never the goal here.

**Trimming the ladder with `TRANSCODER_QUALITIES`.** Already possible, and the first table shows it is not an answer: one rung still takes 9.5 of 16 cores.

**Lowering `TRANSCODING_CONCURRENCY`.** Already possible, and it bounds how many transcodes run at once — not what one of them costs. At the default of 2 a single job already takes two thirds of the machine.

**`nice`/`ionice` on the worker.** Lowers priority without lowering demand. It helps an interactive process win a scheduling contest, but sixteen cores at full tilt still means a hot, busy machine, and it does nothing about the memory bandwidth the other services are competing for.

Lennart-Pingpong:
## Summary

Closes #375.

`TRANSCODER_CPU_LIMIT` takes a core count (`6`) or a share of what the process may use (`50%`). Left empty — the default — the ffmpeg command comes out byte-identical to what it is today, so no existing deployment changes.

Unbounded, ffmpeg picks its own thread counts: about 10 of 16 cores for the default ladder, and still 9 for a single rung, so #325 was no way to get the machine back either. On a self-host that is every core the database and the file server were also using, for the length of every upload.

The property that makes this safe to offer is that a cap costs wall-clock time rather than work. Measured, three rungs, 20 s of 1080p30 on a 16-vCPU host: 4.9 cores instead of 10.6, and total CPU time 66.5 s against 66.9 s. The encode finishes later; the machine stays usable while it does.

## Changes

- **`packages/transcoder/ffmpeg_transcoder.py`**
  - `available_cpus()` — what this process may actually use. Reads the cgroup v2 then v1 quota before falling back to affinity and `os.cpu_count()`, because inside a container `os.cpu_count()` reports the *host*: `50%` on a worker already capped at 4 of 16 with `cpus:` would otherwise hand out 8.
  - `parse_cpu_budget()` — core count or share; an unusable value falls back to unbounded **and says so**, the way an unrecognised `TRANSCODER_OUTPUT` resolves to its default. This setting gets written when a machine is already struggling, so the cost of a typo must be a line in the log, not every upload failing through the retry ladder.
  - `thread_plan()` — splits the budget across the rungs, floors every rung at one thread, and reports the shortfall when the budget is smaller than the rung count. A rung with zero threads is not a quieter encode, it is a command ffmpeg rejects.
  - The plan is computed **once** per job, not inside `_build_ffmpeg_cmd`, which runs a second time on a hardware→software fallback; otherwise one job would log its shortfall twice.
  - Applied as `-threads:v:i` per encoder and `-filter_complex_threads 1`.
- **`.env.example`** — documents the setting, both forms, and that it costs wall-clock time rather than work.
- **`CHANGELOG.md`** — entry under `[Unreleased] / Added`.
- **`apps/api/tests/test_transcoder_cpu_budget.py`** and **`test_transcoder_cpu_budget_applied.py`** — 37 test functions, 46 cases with the parametrised ones expanded.

Three decisions in here are measurements, not preferences:

**The position matters more than the value.** A global `-threads` before `-i` is an input option and reaches the decoder; the encoders never see it. Occupancy stayed at ~11 cores for every value from 1 to 8 — a limit that looks like it works and does not. `test_the_budget_never_becomes_a_global_option` pins the placement, not just the presence, because that is the mutation a later simplification would make.

**The filter graph is not where the time goes.** Holding the encoders at two threads each, capping the graph cost nothing: 13.7 s at one thread against 13.5 s unbounded, 4.7 cores against 4.8. So the budget goes entirely to the encoders and the graph gets one thread — leaving it unbounded would put back the one pool the setting exists to remove.

**Software encoders only.** The hardware backends take their thread counts from the driver; what they still spend on the CPU is decode and the filter graph, which `-filter_complex_threads` covers. I have no such hardware and did not measure it, so I did not touch their arguments.

**How it meets `TRANSCODER_SOURCE_COPY` (#372), now that both exist.** A copy gets no thread arguments at all — there is no encoder and no filter graph to bound, and it costs about one core anyway. It also cannot be short of threads, because a copy only happens once the ladder has resolved to a single rung, which is why the plan can be worked out before the path is chosen. A copy that is *refused* — keyframes too far apart, a rotated source — falls back to a real encode, and that encode is capped like any other. All three are pinned by tests, including the last one, which is the case a naive "skip this for copy jobs" would have broken silently.

## Testing

- [x] Backend tests pass — **578 passed**, 46 of them new.
- [ ] Frontend tests + build — **not applicable**, no frontend file is touched.
- [ ] Tested manually in browser — **not applicable**, there is no UI. What it got instead is an end-to-end run of the real `FFmpegTranscoder` against a real file, S3 stubbed, measuring `getrusage(RUSAGE_CHILDREN)`:

| setting | wall | CPU time | cores occupied | result |
|---|---|---|---|---|
| unset, 3 rungs | 7.3 s | 70.9 s | 9.7 | ok |
| `TRANSCODER_CPU_LIMIT=6` | 13.3 s | 62.6 s | **4.7** | ok |
| `TRANSCODER_CPU_LIMIT=3` | 31.0 s | 60.7 s | **2.0** | ok |
| `TRANSCODER_CPU_LIMIT=25%` | 14.8 s | 61.4 s | **4.2** | ok |
| unset, 1 rung | 5.9 s | 48.3 s | 8.2 | ok |
| `TRANSCODER_CPU_LIMIT=4`, 1 rung | 10.7 s | 45.1 s | **4.2** | ok |

Every condition added here was then removed in turn to check a test holds it — eleven mutations: the per-stream placement, both thread arguments, the per-rung floor, the remainder distribution, the full-budget-is-no-budget branch, the percentage rounding and its floor, the no-budget early return, the cgroup read, and the sub-core quota. **One survived the first pass**: truncating the percentage instead of rounding it. The test I had written only used shares that divide evenly, where both behave the same, and the per-rung floor hid the rest. 35% of 16 is 5.6, so truncation quietly hands out five cores where six were asked for — the test now uses that case.

## Checklist

## Screenshots

Not applicable — no UI change.

---

Rebased onto `a0a6bb7` after the round-1 merges; the conflicts were in `.env.example` and `CHANGELOG.md` and were pure adjacency with #372.

On the two open questions from #375: the budget here is **per transcode**. If you want it per worker that is a division by `TRANSCODING_CONCURRENCY` in one place, and the name is a `sed` — say the word and I will push it.


ravirajsinh45 (changes requested):
@Lennart-Pingpong this is a genuinely good feature and the parts that are hardest to get right are right. A few things worth naming specifically, because they are not accidents:

- The default-off contract is airtight and I checked it byte for byte rather than in spirit. With the variable unset, the three rung 4K command is 59 tokens on main and the same 59 tokens here after normalising the temp directory name. `get_cpu_budget()` short-circuits on an empty string before `available_cpus()` is ever called, so an untouched deployment does not even read sysfs.
- Choosing a per output stream option instead of a global `-threads` is the whole feature working rather than appearing to work, and `test_the_budget_never_becomes_a_global_option` pins both the absence of a bare `-threads` and its position relative to `-i`. A global `-threads` before `-i` reaches only the decoder, so a version of this that looked correct would have capped nothing. Moving the pairs before `-i` reddens exactly that test and nothing else, which is the mutation a later cleanup would have made.
- The arithmetic holds at every boundary I tried: a budget below the rung count keeps every rung at one thread and reports the shortfall rather than emitting a zero thread stream, a budget at or above the machine resolves to unbounded with an honest log rather than a silent no-op, a sub core cgroup quota yields 1, percentages round rather than truncate, and the remainder goes to the earliest and largest rungs.
- Reading the cgroup quota before `os.cpu_count()` is the correct order for a containerised worker, and the reasoning in the comment at lines 1173-1179 for hoisting `cpu_plan` out of `_build_ffmpeg_cmd` earns its length: it really does avoid logging one job's shortfall twice on a hardware to software fallback.
- Three tests cover the seam with #372's copy path, including a refused copy that falls back to a real encode and still gets the budget. That is precisely what a naive "skip the plan for copy jobs" would have broken silently.
- The suite is unusually strong for a config feature. I applied 14 mutations to `packages/transcoder/ffmpeg_transcoder.py` and 11 were killed by a named test, usually the one whose docstring claims that exact property.

What I ran: `python3 -m pytest apps/api/tests/test_transcoder_cpu_budget.py apps/api/tests/test_transcoder_cpu_budget_applied.py -q` gives 45 passed and 1 failed, the failure being a macOS artifact rather than anything about your change (see the nits). The neighbouring transcoder suites are green alongside it. What I could not check: there is no NVIDIA, QSV or VAAPI device here, so the hardware claims below come from reading the command construction and dumping the argv, not from a GPU run. The Docker daemon is down, so no real cgroup limit was ever applied and the v1 and v2 paths were read rather than executed. My performance numbers are ffmpeg 7.1.1 on an 11 core Mac against synthetic sources, so the ratio between capped and uncapped graph threads is the claim, not the absolute seconds, and I did not attempt to reproduce your 16 vCPU figures.

Four things to fix before this goes in. The first two are mechanical and in `.env.example` rather than in your logic, so I have pushed them to your branch as `008db7b9` rather than send you a round for them; they are written up below so you can check the call. The last two are yours, because they are decisions about what the cap should mean rather than defects with one right answer.

**1. `.env.example:161`, the inline comment on a blank value makes every untouched deployment log a false warning on every transcode.** Fixed in `008db7b9`.

`TRANSCODER_CPU_LIMIT=` is the only blank valued variable in the file that carries its explanation as a trailing inline comment, and Compose strips an inline comment only when a value precedes it. Measured, not inferred: `docker compose config` on an env_file holding your exact line hands the service `TRANSCODER_CPU_LIMIT='# How much of the machine ONE transcode may take. Empty = unbounded,'`. Feeding that string to your own `parse_cpu_budget` prints `is neither a core count nor a percentage; leaving CPU use unbounded` and returns None.

Scenario: an operator runs the documented quickstart, `cp .env.example .env` then `docker compose -f docker-compose.dev.yml up`, and never touches the setting. `get_cpu_budget()` is called once per job at line 1180, so they get that line on every transcode, forever. Behaviour is correct throughout, this is log noise, but it burns the credibility of the one line whose entire job is to mean "you mistyped this", and it makes false in the shipped default the exact property `test_transcoder_cpu_budget.py:85-91` was written to protect. The neighbouring `TRANSCODER_SOURCE_COPY=false   # ...` is fine precisely because it has a value.

Fix: the house style is already in the file. REDIS_PASSWORD:29, CORS_ALLOW_ORIGINS:104, MAIL_FROM_NAME:115, SMTP_USER:118 and SMTP_PASSWORD:119 all put the explanation on the lines above and leave the assignment bare. Do the same here.

**2. `.env.example:1`, the rebase mangled the header banner.** Fixed in `008db7b9`.

`git diff origin/main...HEAD -- .env.example` replaces six header lines with three damaged ones. Line 1 is now `# =====================================================# FreeFrame — Environment Configuration`, line 2 is `# =====================================================#` with the lone `#` from old line 4 welded onto the end, all three rulers shrank from 60 `=` to 53 while the file's eight section rules are still 60 wide, and the blank line before `# ─── Database ───` is gone.

60 minus 53 is 7, the exact length of a `=======` conflict marker, so a marker was absorbed into the ruler during the rebase and the neighbouring lines were welded together. That means the PR body's description of the `.env.example` conflict as pure adjacency with #372 is not quite what happened, worth knowing in case the same resolution touched anything else. Nothing fails at runtime since every affected line is a comment, but this is the file the README links as the full config reference. The PR absolutely should touch this file, the `TRANSCODER_CPU_LIMIT` block at 158 is exactly where a new setting belongs, so the ask is just to restore lines 1-4 and the blank at 15-16 from `origin/main:.env.example`.

**3. `packages/transcoder/ffmpeg_transcoder.py:1336`, on a hardware backend the operator's number is discarded entirely.**

`-threads:v:{i}` is emitted only inside the `if backend == "cpu" or dv_software:` branch at line 1336, while `-filter_complex_threads` at line 1310 applies to every backend and is the constant 1. So on nvenc, qsv or vaapi, `TRANSCODER_CPU_LIMIT=2` and `TRANSCODER_CPU_LIMIT=14` build byte identical commands. I dumped the argv with `get_backend` forced to nvenc to confirm it: the only thing a budget of 12 produces is `-filter_complex_threads 1`.

Scenario: an NVENC box sets `TRANSCODER_CPU_LIMIT=8`. For an SDR source the graph is `split` plus `scale_cuda` on the device and decode is on the device too via `-hwaccel cuda -hwaccel_output_format cuda`, so nothing at all is capped. For an HDR source the graph at lines 1276-1284 is `hwdownload, format=p010, zscale, tonemap, zscale, hwupload_cuda`, which is the only CPU bound stage in that job, and it gets one thread whether the operator wrote 2 or 14. So the setting is not a dial there: any value means "one thread for the tone map", and the larger the number written, the more misleading the result.

One related correction to the CHANGELOG while you are in there. "what they spend on the CPU is decode and the filter graph, which the same cap covers" is wrong about decode on every backend, not only hardware. There is no global `-threads` anywhere in the file, and `test_the_budget_never_becomes_a_global_option` exists to keep it that way, correctly, because before `-i` it would be a decoder option. So decode is never capped, and on hardware it is doubly uncovered because decode runs on the device.

Either scale the filter threads with the budget on the hardware path too, or skip the cap entirely when the backend is not cpu or dv_software and say plainly in `.env.example` that it has no effect on a hardware backend. Whichever you choose, please pin it with a test that builds the command for `backend="nvenc"`, because right now nothing does: I both added `-threads:v:{i}` to the nvenc branch and gated line 1310 on `backend == "cpu"`, and the suite stayed green for both.

**4. `packages/transcoder/ffmpeg_transcoder.py:442`, the "cost nothing at all" measurement was taken on the SDR graph and is repeated as general in three operator facing places.**

Line 442 is `return per_rung, 1`, so the filter graph gets one thread for any budget on any backend. The docstring at 424-428 justifies that with "capping the graph at one thread cost nothing at all (13.7 s against 13.5 s unbounded; 4.7 cores against 4.8)", and the same claim is in CHANGELOG.md:20 and in the `.env.example` block.

For the split and scale graph that is true and I reproduced it. It is not true for the graph this code builds for every PQ or HLG upload, which is the common case rather than an exotic one: `get_hdr_mode()` defaults to `convert` at line 462 and `arib-std-b67` sets `is_hdr` at line 1056, so an ordinary iPhone clip runs through the `zscale, tonemap, zscale` chain at lines 1266-1275 before the split. Both zscale and tonemap are slice threaded, so pinning the graph serialises the one genuinely expensive filter in the file while the encoders share the rest of the budget.

Scenario: a 16 core self host sets `TRANSCODER_CPU_LIMIT=6` and someone uploads an HLG clip. The three encoders get two threads each, the tone map chain gets one, and the job occupies well under half of what was granted. Measured on my box with the encoders held at two threads each, the ratio and the monotonicity being the claim rather than the seconds: `-filter_complex_threads 1` sits near one core and runs between roughly 1.5x and 3.6x the wall time of the same encoders with the graph given 3 threads or left alone, with fct=3 and fct=6 landing in between in the expected order. The same sweep on the SDR ladder reproduces your number, so the measurement is right about the graph it was taken on.

The cap is never exceeded and total CPU time is unchanged, so the headline promise of the feature holds. What does not hold is the generalisation. Two ways out and either is fine by me. Give the graph a measured share of the budget when `tone_map` is true, for example `max(1, budget // 2)`, which in my runs stayed inside the granted cores and was substantially faster. Or keep the constant 1 deliberately and scope the claim to the SDR graph in the docstring, the CHANGELOG and `.env.example`. Note that simply dropping the option is not the answer: unset, the graph went over the granted budget on my source, which is exactly the unbounded pool this setting exists to remove, so your instinct to bound it is right. One more thing worth dropping from the claim: Dolby Vision tone maps via libplacebo, which is GPU backed and not slice threaded, so I would not expect the same serialisation there and I could not measure it, no libplacebo in this build.

Non blocking, nothing here needs to hold up a merge:

- The cgroup v1 branch is untested in both directions. Deleting the v1 tuple at lines 332-333 leaves the suite green, and so does weakening `quota_s not in ("max", "-1")` to `!= "max"` at line 341, and narrowing `except (OSError, ValueError)` to `except OSError` at line 345. The middle one matters: on an unrestricted v1 host `cfs_quota_us` reads -1, so the computed core count falls below 1 and the clamp returns 1, which makes every budget resolve to unbounded on every v1 deployment with one log line to show for it. Three tests in the existing `fake_read_text` style would pin it: a v1 quota of 400000 over 100000 giving 4, a v1 -1 falling through, and a garbage quota file falling through rather than raising.
- `docs/deployment.md` gains no row, although the table at lines 273-277 carries every other `TRANSCODER_*` setting including the `TRANSCODER_SOURCE_COPY` row #372 added. Worth cross referencing `TRANSCODING_CONCURRENCY` in the same row.
- The cap is per ffmpeg process and the transcoding worker ships at `-c ${TRANSCODING_CONCURRENCY:-2}`, with `docs/deployment.md:293` recommending 4 to 6 on larger boxes, so the machine wide ceiling is the budget times the concurrency. Your wording says "ONE transcode" so it is not wrong, just under specified. One sentence in the env comment naming the variable 23 lines above it would close it.
- CHANGELOG credit reads `(#375 by @Lennart-Pingpong)`, which is the issue. House style when there is both is issue then PR, as the four neighbouring Added entries do: `(#375, #390 by @Lennart-Pingpong)`. The entry itself is correctly under [Unreleased].
- `test_transcoder_cpu_budget.py:218` cannot run on macOS: `monkeypatch.setattr(mod.os, "sched_getaffinity", ...)` raises because the attribute is Linux only. `raising=False` on that one call turns a Mac contributor's run green. Your production guard at line 350 is correct and load bearing, dropping AttributeError from that tuple reddens two tests here, so only the test's own patching needs it.
- Two pre existing things I hit while measuring, both on main and outside your diff, flagged only so nobody mistakes them for regressions here and worth their own issues: `-pix_fmt` and `-crf` at lines 1325-1327 are not stream qualified while `-c:v:{i}` is, so ffmpeg warns about multiple `-pix_fmt` and every rung ends up on the last rung's CRF; and `apply_watermark` at `apps/api/tasks/watermark_tasks.py:94` runs an uncapped software re-encode on the same transcoding queue, so one job there can still take the box that this setting was turned on to protect.

Fix 1 and 2 and pick a direction on 3 and 4 and I will re-review quickly. The core of this, the split, the placement, the parsing and the tests, I would take as is.

On the two I pushed: the explanation moves above the assignment, which is what the file's other blank-valued vars already do (`REDIS_PASSWORD:29`, `CORS_ALLOW_ORIGINS:104`), and I verified the fix the same way I found the problem, by running `docker compose config` against an env_file holding the new lines: `TRANSCODER_CPU_LIMIT: ""`. The header is restored from main, so the working diff for `.env.example` is now the added block and nothing else. Revert either if you disagree.

Nothing here is a reason to think less of the feature. The default-off contract, the per-stream option and the test suite are the parts that are hard to get right, and those are right.

Lennart-Pingpong:
Thank you — all four hold, and I checked each one myself rather than taking it on trust. Two of them I got wrong in a way that is worth spelling out, because the second one is worse than you diagnosed.

**1 and 2.** Both of your commits stay exactly as they are. I reproduced the Compose behaviour independently before accepting it: `docker compose config` against an env_file holding my line hands the service `TRANSCODER_CPU_LIMIT='# How much of the machine ONE transcode may take. Empty = unbounded,'`, and feeding that to my own `parse_cpu_budget` prints the typo warning and returns None. So the one property `test_transcoder_cpu_budget.py` was written to protect was false in the shipped default, on every untouched deployment, forever. On 2 you are right and my PR body was too kind to myself: that was rebase damage, not adjacency, and your reading of the seven missing characters is exactly what happened.

**4, the graph share.** You are right that the measurement was taken on the SDR graph and generalised. It is worse than that. The measurement was taken on a **1080p** source, and what makes a graph cheap is not the absence of tone-mapping but the absence of work. Measured on 16 cores, three rungs, budget 6, three runs each, spread 1-4%:

| source | graph=1 | graph=3 | graph=6 |
|---|---|---|---|
| 1080p SDR | 16.2 s / 4.44 cores | 16.2 s / 4.38 | 16.5 s / 4.39 |
| 4K, scale only (`TRANSCODER_HDR=preserve`) | 30.4 s / 2.99 | 16.2 s / 5.87 | 10.7 s / 9.02 |
| 4K HDR, tone-mapped | 101.5 s / 1.65 | 51.1 s / 3.68 | 30.5 s / 6.44 |

Read the middle row: a 4K source with **no tone-mapping at all** still runs at half speed with the graph pinned, because scaling three rungs from 4K is real work. So the rule cannot hang on `tone_map`, which is where both of us expected it to land. It hangs on how much picture goes through the graph, and the budget is already the best proxy for that which the code has.

So the share is `max(1, budget // 2)` — your number, and it is the only one that stays inside the cap on all three sources. One thread holds the cap only by wasting it: six cores granted, 1.65 taken. The whole budget is another 1.5x faster and overruns the cap on both 4K sources, by 0.44 cores on the tone-map graph and by 3.0 on the scaling one, so it is out. On the 1080p source the share makes no measurable difference either way, which is why nothing is traded away for the sources that do need it. The graph's threads are not subtracted from the encoders': thread counts are ceilings on parallelism rather than reservations, and the stages are a pipeline, so the sum of the ceilings is not what the job occupies. What it occupies is the table above.

**3, the hardware backends.** Fixed by the same change rather than by a second one. With the graph scaling, `TRANSCODER_CPU_LIMIT=2` and `=14` no longer build identical commands on nvenc, and on a hardware backend the filter graph is most of the CPU work there is, since scaling runs on the device and only an HDR tone-map comes back. Three tests now build the command for `backend="nvenc"`, including one that asserts two different budgets produce different commands — you were right that nothing covered that path, and I could add or remove the fix without reddening anything.

Your CHANGELOG correction about decode is accepted and the claim is gone from all three places it appeared. I also measured it, because I wanted to know whether the gap mattered or merely existed. Decoding that ProRes 4K master alone, no filters and no encoder: 11.38 cores and 65 CPU-seconds unbounded. That is not small. It still does not blow the budget, because in a pipeline the decoder can only run as fast as the stages after it take frames — capping it alongside the graph changed the full job by nothing measurable, 31.3 s against 30.5 s. Worth knowing if it ever does matter: `-threads N` before `-i` bounds the decoder only loosely, since ProRes threads on frame *and* slice level. `-threads 2` gave 4.61 cores, not 2. So "capped" would be the wrong word for it even if we used it.

**How the numbers were taken**, since a number without its conditions is not worth much. Real source rather than synthetic: a 4K HLG clip off an iPhone, ProRes 422 HQ 10-bit, 12.5 s, 313 frames. The SDR row is the same file with `TRANSCODER_HDR=preserve`, so the decoder and the source are held constant and only the graph changes. The commands are read out of the actual code rather than retyped. And my first run at this was thrown away: it ran next to an embedding model eating 13 of 16 cores, and reported the same configuration at 6.00 cores and at 2.74. The tool now reads `/proc/stat` per run, subtracts its own children, and discards a run where the machine saturates — and every design is run three times, because reproducible numbers cannot be driven by fluctuating foreign load. That is what the 1-4% spread above is.

**Verified in a container rather than by reading**, since you could not: with `--cpus=6`, `available_cpus()` returns 6 and not the host's 16, `50%` resolves to 3 rather than 8, and `6` on a six-core quota resolves honestly to unbounded instead of pretending. A full transcode of the HLG master at budget 4 under that quota occupied 2.75 cores and produced playable HLS — 21 segments, a master playlist, `bt709`, rotation preserved.

**cgroup v1 is now tested in both directions**, and each test kills the mutation you named:

| mutation | before | now |
|---|---|---|
| v1 tuple deleted | green | **red** |
| `not in ("max", "-1")` weakened to `!= "max"` | green | **red** |
| `except (OSError, ValueError)` narrowed to `except OSError` | green | **red** |

You were right that the middle one is the one that matters: on an unrestricted v1 host `cfs_quota_us` reads -1, the computed count falls below 1, the clamp makes it 1, and then every budget on that host resolves to unbounded with one log line to show for it.

The rest of the nits are done: `raising=False` on the `sched_getaffinity` patch so a Mac run is green, the CHANGELOG credit is `(#375, #390 by @Lennart-Pingpong)`, `docs/deployment.md` has a row cross-referencing `TRANSCODING_CONCURRENCY`, and `.env.example` says plainly that the ceiling for the machine is the budget times the concurrency.

Your two pre-existing findings I will file separately rather than fold in here — the unqualified `-pix_fmt` and `-crf` putting every rung on the last rung's CRF, and `apply_watermark` re-encoding uncapped on the transcoding queue. The second one is the more interesting of the two, since it is the same machine this setting exists to protect.

586 tests pass.

ravirajsinh45 (changes requested):
@Lennart-Pingpong this closes all four, and the way you closed 4 is better than the way we asked for it. We expected the rule to hang on `tone_map` and you showed with the middle row of your table that it cannot: a 4K source with no tone-mapping at all still halves in speed with the graph pinned, because what makes a graph expensive is how much picture goes through it, not which filters are in it. The budget really is the best proxy the code has for that.

What we verified rather than took on trust:

- **The default-off contract still holds byte for byte.** This is the property most worth re-checking, because the one executable line you changed sits inside the decision that produces it. With the variable unset we built the command on `main` and on `ae757d3c` and compared token by token after normalising the temp directory: 4K SDR three rungs 59/59 identical, 4K HLG 77/77, 4K HLG on nvenc 87/87, 1080p SDR 59/59, single rung 35/35. Also identical with `TRANSCODER_CPU_LIMIT=` set empty, which is what Compose actually passes.
- **The hardware path now responds to the operator's number on every backend.** Dumping argv at budget 2 against budget 14 with the temp directory normalised: nvenc SDR, nvenc HDR, qsv SDR, qsv HDR, vaapi SDR and vaapi HDR each differ in exactly one token pair, `-filter_complex_threads 1` going to `7`, with no `-threads:v:i` anywhere. `cpu` and `dv_software` differ in four pairs, the graph plus the three encoder counts. `dv_software` being grouped with `cpu` rather than with the hardware backends is right.
- **The nvenc tests really build an nvenc command.** The mocked `subprocess.run` does not quietly fall back to software: the command carries `h264_nvenc` on all three rungs, `-hwaccel cuda -hwaccel_output_format cuda`, and the full `hwdownload, zscale, tonemap, zscale, hwupload_cuda` chain. That was the specific thing we asked for and it is there.
- **All three cgroup v1 mutations are dead.** Deleting the v1 tuple reddens `test_a_cgroup_v1_quota_is_read`; `("max", "-1")` to `("max",)` reddens `test_an_unrestricted_cgroup_v1_host_falls_through` with `assert 1 == 8`, which is exactly the silent-death mechanism; `except (OSError, ValueError)` to `except OSError` reddens `test_an_unreadable_quota_falls_through_rather_than_raising`. Each reddens one test and leaves the rest green.
- **Blocker 3's fix is pinned.** Putting the old software-only gate back on line 1335 turns `test_a_hardware_backend_still_bounds_the_filter_graph` red. So the fix-list item is genuinely closed, whatever we say about the third test below.
- **The suite runs clean on macOS now**, 54 passed and 0 failed against 45 and 1 before. `raising=False` went onto all three Linux-only patches, not just the line we named, and it is load-bearing rather than hollowing: removing it fails the test here, and swapping the affinity read for `os.cpu_count()` still fails it.
- **The stream-copy seam with #372 is untouched.** A copy job carries no `-threads`, no `-threads:v:i` and no stray `-filter_complex_threads` on a job that has no filter graph, and a refused copy still lands on an encode that carries both the budget and the half-share.
- **Your 1080p SDR row reproduces here almost exactly.** On an 11 core box, three rungs, budget 6, a 1080p H.264 source: graph=1 gives 4.31 cores and graph=3 gives 4.46, against your 4.44 and 4.38 on 16. That is the control that tells us our measuring method matches yours, which matters for the one thing below where we get a different answer.

Three of the things below are mechanical, so we have pushed them to your branch as `789a8186` rather than send you a round for them, the same way we did with `008db7b9`. They are written up so you can check the call, and revert any of them if you disagree. The one that is left is yours, because it is about what your own measurements are allowed to claim.

**1. `apps/api/tests/test_transcoder_cpu_budget_applied.py:230`, the test named after the regression is the one test in the file that cannot fail.** Fixed in `789a8186`.

`assert klein != gross` compares two whole argv lists, and each `_ffmpeg_cmd_for` call makes its own `tempfile.mkdtemp` directory that lands in `-hls_segment_filename` and the playlist path. So the two lists always differ in two tokens regardless of what the budget does.

Measured. At `ae757d3c` the two commands differ in three tokens: `1` against `7`, and the two temp paths. Restore the round-1 defect with `return per_rung, 1` and `test_the_filter_graph_gets_half_the_budget` and `test_a_hardware_backend_still_bounds_the_filter_graph` both go red, while this one passes. Delete the `-filter_complex_threads` emission outright and four tests go red, and this one still passes.

The regression itself is covered, by `test_a_hardware_backend_still_bounds_the_filter_graph` 26 lines above, so this is not a coverage hole. It is worth fixing anyway because the comment above it is the most confident in the file, it tells the next reader this is the guard against exactly that regression, and someone pruning duplication would keep this one and drop the two that work.

What we pushed compares the one token the budget can reach, with an `assert "-filter_complex_threads" in klein` above it so the deleted-emission case fails with a message rather than a ValueError. Your comment is kept, it is a good one, with the reason the old assertion could not fail added under it. Proof that it is now a test: `return per_rung, 1` takes the file from 2 failures to 3, and deleting the emission takes it from 4 to 5, this test being the addition in both.

**2. The cap claims went absolute in this round, and on a real NLE master the cap is exceeded. This one is yours.**

Round 1's text described measurements. This round's text makes promises: `ffmpeg_transcoder.py:450` "at half the budget it stays under the cap on every source tried", `:438` "the only share that stays inside the cap everywhere", and CHANGELOG.md:20 "half is the share that both uses the grant and keeps the promise" plus "which is why a capped transcode holds anyway".

We tested it on a real master rather than a synthetic one: a 1.8 GB ProRes XQ export out of an NLE, 1920x1080, 4:4:4 12-bit, 318 Mbps, PCM audio. First 15 s, three rungs, budget 6, idle 11 core box, ffmpeg 7.1.1, two runs each with under 2% spread, cores measured as CPU time over wall time.

| source, same 15 s of footage | uncapped | graph=1 | graph=3, shipped |
|---|---|---|---|
| re-encoded to 1080p H.264 first | 8.88 | 4.31 | 4.46 |
| the ProRes XQ master itself | 8.53 | **6.90** | **7.46** |

The first row is the control, and it lands on your published numbers. The second row is the same picture at the same resolution through the same ladder, and the shipped configuration takes 7.46 cores against a grant of 6, against an uncapped baseline of 8.53. The cap recovers about one core out of eight and a half, and it is 24% over what the operator asked for.

The cause is not the half-budget change. `graph=1` also exceeds the cap on this master, at 6.90. We checked the obvious suspect and it is not that either: adding `-threads 2` before `-i` leaves the full job at 7.46 and 7.38, and `-threads 1` at 7.47 and 7.39, so throttling the decoder changes nothing, exactly as you argued. What it is, is that the decoder's CPU cost sits outside the budget and on this master it is enormous: decoding alone is 17.3 CPU-seconds of a 44.5 CPU-second job, and the same 15 s as H.264 is a 20.5 CPU-second job in total. Decode and the 4:4:4 12-bit conversion are more than half of what this transcode costs, and none of it is inside the cap.

So `.env.example:178-181` has the right instinct and the wrong scope. It already names ProRes as the exception, but scopes it to hardware backends: "a hardware backend with a heavy master (ProRes 4K decodes at ~11 cores unbounded) is the case where that stops being true". It stops being true on the software backend too, which is the default, and ProRes is what a self-hoster's editors export. The sentence before it, "It is rarely the problem, because the decoder can only run as fast as the rest of the pipeline takes frames from it", is the one that does not survive: the decoder is indeed throttled here, and it still costs 17 CPU-seconds while being throttled.

Nothing about the feature needs to change for this. The code does what it says: it bounds the encoders and the graph. Our ask is to put the absolutes back to the measurements they came from. Something like "on the sources measured above it stays inside the cap" in the docstring, and in `.env.example` and the CHANGELOG a plain sentence that the budget covers the encoders and the filter graph, that decode sits outside it on every backend, and that on a heavy master such as ProRes decode alone can be larger than the budget, so the process ceiling is the budget plus whatever the decoder costs. Said that way it is still a setting worth turning on, and it is one an operator can trust.

Non blocking, none of this needs to hold up a merge:

- **`ffmpeg_transcoder.py:438`, one word. Fixed in `789a8186`.** "Half the budget is the only share that stays inside the cap everywhere" is contradicted by the table three lines above it, where graph=1 reads 4.44, 2.99 and 1.65 against a budget of 6, all inside. We made it "the largest share that stays inside the cap on all three", which is what the table supports, and your next sentence already makes the real point.
- **"a command ffmpeg rejects" is the one justification that is false, and the truth is a better argument.** It appears at `test_transcoder_cpu_budget.py:162`, in `test_a_budget_of_one_still_leaves_the_graph_a_thread`, and in CHANGELOG.md:20 as "rounded to a command ffmpeg would reject". ffmpeg 7.1.1 accepts a zero thread count everywhere we tried it: `-threads 0`, `-threads:v:0 0` and `-filter_complex_threads 0` all run to completion. Zero is how ffmpeg spells auto. Measured on a 1080p encode: `-threads:v:0 0` takes 7.58 cores, `1` takes 1.14, `2` takes 2.48. So your `max(1, ...)` floors are guarding against something worse than a crash, which is the cap silently ceasing to apply, and that is what the comment should say.
- **`test_transcoder_cpu_budget.py:269`, the v1 test is dead on a 4 core host. Fixed in `789a8186`.** `test_a_cgroup_v1_quota_is_read` asserts `available_cpus() == 4` without pinning what the code falls through to, so when the v1 tuple is deleted it falls through to the host count and passes on any host reporting 4. We simulated it: at 11 available CPUs the mutation is caught, at 2 it is caught, at 4 it passes. `ubuntu-latest` is a 4 vCPU runner, so the nit we raised is closed on your box and on ours and may well be open in CI. We added the affinity patch your two sibling v1 tests already carry, so the assertion is reachable only through the v1 read: with the tuple deleted it now fails `assert 9 == 4` on a simulated four vCPU host. The same hole is in `test_a_cgroup_quota_wins_over_the_host_core_count` at line 211, which predates this round and we have left alone.
- **`test_transcoder_cpu_budget.py:228`, the new comment claims coverage that does not exist.** "The production guard for that same absence is a real one and is covered by its own test below." Nothing below covers the `AttributeError` fall-through to `os.cpu_count()`, and on Linux nothing covers it at all. Either drop the sentence or make it true in three lines: `monkeypatch.delattr(mod.os, "sched_getaffinity", raising=False)`, `monkeypatch.setattr(mod.os, "cpu_count", lambda: 7)`, `assert mod.available_cpus() == 7`.
- **`.env.example:172-177`, the hardware sentence is right about HDR and wrong about the common case.** "only the filter graph is capped, which on those backends is most of the CPU work there is" holds for a tone-mapped source, where `hwdownload, zscale, tonemap, zscale, hwupload` really is the CPU work. On an SDR source the graph the code builds for nvenc is `split` plus `scale_cuda` with CUDA frames in and out, so nothing in it runs on the CPU and the number bounds nothing. Worth scoping, because SDR masters are the common upload and that is the operator most likely to read this line and then set the variable.
- **The ProRes parenthetical in the same paragraph picks a source that would not still be on a hardware backend.** The graph is built out of device scale filters, so a codec the device cannot decode fails the hardware attempt and falls back to `cpu`, where the encoders do get the budget. As it happens the general point is stronger than the example: our measurement above is a software backend with a ProRes master, so the parenthetical can simply lose "a hardware backend with".

One piece of housekeeping. We filed the unqualified `-pix_fmt` and `-crf` finding as #393 yesterday, with a measurement: on a three rung ladder the 640x360 stream comes out at 161,984 bps unqualified against 948,288 bps stream-qualified, because the ladder runs largest to smallest and every rung inherits the last rung's CRF. So no need to file that one. Thank you for #396, and for #394 and #395 while you were in there.

What we could not check: no NVIDIA, QSV or VAAPI device here, so every hardware claim above is from the constructed argv and never from a job that ran on that hardware. Docker is down, so the cgroup v1 and v2 readers were exercised through a patched `read_text` rather than against a real quota, and we could not repeat your `--cpus=6` container run. Our box is 11 core Apple Silicon with performance and efficiency cores, so CPU time over wall time is a fair measure of how much of the machine a job occupies but not of throughput, and we did not try to reproduce your absolute seconds from a 16 core Linux host. The 1080p H.264 control above is there precisely so the one number we do dispute can be read against a number of yours that we reproduced.

So the one thing holding this is the wording of the cap claims. Scope those and we will merge. The design, the split, the placement, the hardware fix and the cgroup coverage are all where they should be, and the part of this that was hard to get right is right.

Lennart-Pingpong:
Thank you — the blocking item is done, and the detour I took getting there is the part worth your time. Pushed as `2f5bbd3`. No executable line changes in it.

**2, the cap claims.** You are right that round 2 turned measurements into promises, and I would have scoped the wording on your word alone. I tried to reproduce the ProRes run first anyway, and did not — which turned out to be more useful than reproducing it.

I built your master as closely as I could from the description: 1920x1080 ProRes 4444 XQ, 4:4:4 12-bit, 330 Mbps against your 318, PCM audio, 12.5s of it. Three rungs, budget 6, 16-core Linux host, ffmpeg 7.1.5, three runs each, spread under 3%.

| source, same footage | uncapped | graph=1 | graph=3, shipped |
|---|---|---|---|
| re-encoded to 1080p H.264 first | 12.19 | 4.46 | 4.45 |
| the ProRes XQ master itself | 12.74 | 5.46 | **5.47** |

Your control row and mine agree to 0.02 cores, so we are measuring the same thing. But the ProRes row does not: 5.47 here against your 7.46, and here it stays inside the cap. Same shape of master, same ladder, same budget, opposite verdict.

The difference is not the master. Decode alone costs 24.0 CPU-seconds here against 5.8 for the same picture as H.264, which is your finding exactly — decode is expensive and it is outside the budget. What differs is the divisor. An uncapped decoder runs only as fast as the capped stages take frames from it, so those 24 CPU-seconds spread over a 16.9s job are 1.42 cores, and the arithmetic closes: 4.05 capped + 1.42 decode = 5.47, and on the H.264 row 4.1 + 0.36 = 4.46.

Your host encodes the same ladder in roughly a third of the wall-clock, so the same decode lands in a third of the window. **The faster the machine, the leakier the cap.** That is a prediction, so I tested it the only way I can without your hardware: shorten the job instead of speeding up the machine. One rung instead of three, everything else identical.

| ladder, budget 6 | H.264 master | ProRes XQ master |
|---|---|---|
| three rungs | 4.40 — inside | 5.49 — inside |
| **one rung** | 5.92 — inside | **7.70 — 28% over** |

There is your 7.46, on my box, at 7.70. So the rule is not "ProRes overruns". It is that the process ceiling is the budget plus the decoder's CPU time divided by how long the rest of the job runs, and both a faster machine and a shorter ladder shrink that divisor. Which means the leakiest configuration is a single-rung ladder — the one `TRANSCODER_QUALITIES` and `TRANSCODER_SOURCE_COPY` both steer a self-hoster toward.

The docstring, `.env.example`, the CHANGELOG and the deployment table now say that instead of promising a ceiling: the budget covers the encoders and the filter graph, decode sits outside it on every backend, the process ceiling is the budget plus whatever the decoder costs, and both numbers above are given so an operator can see which end of the range they are on. I kept your framing that it is still worth turning on, with the number that makes the case — 7.70 against 12.74 unbounded, in the worst row I have.

One correction to your wording rather than mine: "decode alone can be larger than the budget" is not what either of us measured. As occupancy it is 2.45 cores here at worst and about 2.9 on yours. It is large enough to break a cap, not larger than one.

**1, the test that could not fail.** You are right, and it is worse than a nit from where I sit: that is the test I wrote last round *for your finding*, and it was green under the exact regression its own comment claims to pin. Your replacement is what it should have been. Thank you for measuring it rather than just reading it.

**The cgroup v1 test, and its sibling.** Also right, also taken. I did go one further: `test_a_cgroup_quota_wins_over_the_host_core_count` has the identical hole, and you left it alone as predating this round — but it is from round 1 of this branch, so it is mine to fix. Verified the way you verified yours, under `taskset -c 0-3` to simulate the runner: with the cgroup v2 read deleted the old test passes and the patched one fails.

**"A command ffmpeg rejects".** Confirmed on my own box before changing it, and it reproduces your numbers: on ffmpeg 7.1.5, `-threads:v:0 0` occupies 7.97 cores, `1` occupies 1.17, `2` occupies 2.50, and no option at all occupies 7.88. Zero is auto, to within noise. You found it in three places; there were seven. The explanation now lives once, next to the measurement table in the module, and the comments point at it: the `max(1, ...)` floors guard against the cap silently ceasing to apply, which is worse than a crash because nobody sees it.

**The coverage claim at `test_transcoder_cpu_budget.py:228`.** Made true rather than dropped — your three lines, as `test_a_host_without_sched_getaffinity_falls_back_to_the_core_count`. Narrowing the except clause to `OSError` alone reddens it and nothing else.

**`.env.example`, the hardware sentence and the ProRes parenthetical.** Both as you scoped them. The graph is now described as worth having for an HDR master whose tone-map runs on the CPU and worth little for an SDR one that scales on the device, and the parenthetical has lost "a hardware backend with", since a codec the device cannot decode falls back to the CPU backend anyway.

**What I could not check.** No NVIDIA, QSV or VAAPI device here either, so my hardware claims are still argv-only, same as yours. My ProRes master is a tone-mapped and downscaled iPhone HLG clip encoded to XQ rather than a real NLE export, which matches yours on codec, profile, pixel format, resolution and bitrate but not on picture content — if the difference between 5.47 and 7.46 matters to you beyond the explanation above, that is where I would look next. And I did not reproduce your per-backend argv diffing; I took that one on trust, since I have nothing to add to it.

Full suite green, 587 passed.

ravirajsinh45:
@Lennart-Pingpong the round-2 blocker is closed and the decode paragraph is the right explanation. We verified rather than read: `2f5bbd3a` is comment-only in the production module (parsed both revisions, stripped every docstring, compared the AST dumps: identical), the affinity pin does close the hole it claims under a simulated 4 vCPU runner, the new `sched_getaffinity` test is real on both platforms, and nine other mutations of the shipped code are all caught.

One thing in it is wrong, and it is the sentence that replaced the blocker, so we have fixed it and the rest rather than send you another round. Pushed as `66ab4697`.

**`ffmpeg_transcoder.py:474`.** "The sources above are all H.264 and all three-rung, which is why they all held" points at a table whose two 4K rows are one ProRes 422 HQ 10-bit master. Your own note says so: "a 4K HLG clip off an iPhone, ProRes 422 HQ 10-bit, 12.5 s, 313 frames. The SDR row is the same file with `TRANSCODER_HDR=preserve`." They held for the length of their jobs (16.2 s and 51.1 s at half the budget), which is the divisor argument the paragraph already makes. Codec was never the reason, and the same docstring has a three-rung ProRes run 50% over at graph=6.

**One real bug, separate and small.** `float()` accepts `nan`, `inf` and `infinity`. None survives `round()`, which raises `ValueError` for NaN and `OverflowError` for inf, and nothing catches either before the job's outer handler. A single mistyped share failed every upload at 0% with an error that never named `TRANSCODER_CPU_LIMIT`, which is exactly what falling back exists to prevent. The `except` branch now resolves to NaN and one `isfinite` check covers both the strings `float()` rejects and the three it should not accept. Removing that guard reddens five tests.

**A third test that could not fail, and two call-site assumptions nothing held.** `test_a_stream_copy_cannot_be_short_of_threads` asserted only the absence of a warning, so deleting the shortfall report outright left the whole applied file green (16 passed). It now asserts the warning is live at that budget first. Its comment also claimed to guard against the copy gate widening to a real ladder; it cannot, because the gate decides the rung count it sees, and `test_source_copy.py` already owns that mutation. Separately: every budget in the file divides evenly across its rungs, so collapsing `cpu_plan[0][i]` to `cpu_plan[0][0]` emitted 9 encoder threads on a 7-core budget and reddened nothing, and every test used a 1080p source, so swapping `len(qualities)` for `len(requested)` cut the budget to a third on smaller uploads and reddened nothing. Budget 7 over three rungs now pins 3/2/2, and a 640x360 upload pins the whole budget on the one rung that survives pruning.

The two whitespace tests keep their cases and lose their claims: `int()` and `float()` strip for themselves, so neither could detect the removal of the `.strip()` it was named for.

**Prose.** `.env.example` and the CHANGELOG carried only the three-rung ProRes figure that stayed inside the cap; both now give the range including the 7.46 we measured on the same configuration. "7.70 against 12.74 uncapped" read as like-for-like and was not (7.70 is one rung, 12.74 is three), so it is labelled. "Because the decoder is not what the pipeline waits on" is not what the null result shows, and your own measurement is the better reason: `-threads 2` gave 4.61 cores, not 2. "Largest share" became "largest share tried", 4.9 and 4.45 are distinguished by their sources, and the upstream report carries `(#390)` the way this file cites every other one.

**One thing we documented rather than fixed, and it is worth your eye.** A cap multiplies wall-clock against an unchanged 4-hour ffmpeg ceiling, and past it the version fails outright: `subprocess.TimeoutExpired` is a `SubprocessError`, so the `except RuntimeError` at `:1506` does not catch it, no fallback attempt runs, and the outer handler returns `success=False`. Three rungs at one thread each run at 0.68x realtime, so a master of roughly three hours becomes a failed upload. Pre-existing, not yours, but this setting is what makes it reachable. `.env.example` and `docs/deployment.md` now say so. Making the timeout scale with the budget belongs in its own change and we will file it.

84 passed across the three transcoder suites, 528 in the API suite against 522 before. Every condition we added was removed in turn to check a test holds it. Revert anything you disagree with. Approving once CI is green.

ravirajsinh45 (approved):
CI green on `66ab4697`: Backend Tests, Frontend Build, Lint and Docker Build all pass. Approving. Shout if you want any of what we pushed reverted, otherwise we will merge this.

PR #397, opened 2026-09-19
fix(web): ask /me/assets for the user's own uploads only

### LINKED ISSUE #374: Uploads panel lists assets the user did not upload (/me/assets called without a filter)
### Describe the bug

The uploads panel's Complete tab is built from `GET /me/assets` (`upload-store.ts:1277`, `fetchHistory`), and it passes no `filter`. With no filter, `list_my_assets` (`apps/api/routers/me.py:23`) takes its `else` branch, which returns every asset in every project the user is a member of, in any role, plus anything shared with them or assigned to them.

So the panel labelled "Uploads" lists assets the user did not upload — for a reviewer who has never uploaded anything, it is a list of other people's cuts.

There is a filter that comes close: `filter=owned` narrows to `Asset.created_by == current_user.id`. It is not passed. It is also not exactly the right question, which is the part that may need a decision rather than a patch: `created_by` is who created the *asset*, so a second version uploaded onto someone else's asset is the user's upload and is not `owned`, while an asset created long ago and never touched since is `owned` and is not a recent upload of theirs.

This surfaced in #368, which adds a rename to the panel and has to carry a `fromHistory` flag purely to suppress the control on these rows, because offering it to a reviewer would answer 403. That flag is standing in for the filter this call does not pass, and anything else added to the panel later has to remember the same thing — silently, until a user hits the 403.

### Steps to reproduce

1. As user A, create a project and upload an asset into it.
2. Add user B to that project as a **Reviewer**.
3. Sign in as B. Upload nothing.
4. Open the uploads panel (the cloud icon in the header) and switch to the **Complete** tab.
5. A's asset is listed there, under B's uploads.

The same thing, without a browser:

### Expected behavior

Checked directly against the two branches' queries on a dev database, with a project member who had created nothing: 1 row from the `else` branch, 0 from the `owned` branch.

### Screenshots

The uploads panel shows the user's own uploads, and a reviewer who has never uploaded anything sees an empty Complete tab.

Two ways there, and the choice is yours:

- **The small one:** `fetchHistory` passes `filter=owned`. One line, no API change, and it makes the panel wrong only in the narrow case of a version uploaded onto an asset someone else created.
- **The exact one:** a filter that means "versions I uploaded" — `AssetVersion.created_by == current_user.id` — which is what an upload history actually is. That is an API change, so it is your call whether it is worth one.

Either way the default of `/me/assets` stays as it is: the command palette (`command-palette.tsx:59`) searches with `?q=` and no filter, and for a search box "everything I can reach" is the right answer.

### Area

Frontend (Next.js)

### Version / Commit

main at 216e0a5

Lennart-Pingpong:
The uploads panel rebuilds recent uploads from `GET /me/assets` and asked for it with no filter. Unfiltered, that endpoint answers with every asset in every project the user is a member of, in whatever role, plus anything merely shared with or assigned to them -- so the panel's history was not "your uploads" but "everything you can reach". A reviewer who has never uploaded anything opened it to a list of other people's work.

`filter=owned` already exists and resolves to `created_by = <this user>`, which is literally "uploaded by them". Both pages ask for it now: fixing only `fetchHistory` would have let the first "load more" put the other assets back, and that is the shape the new test pins. Paging is unaffected -- `skip` and `limit` run against the filtered set, so a shorter list only ends the paging sooner.

Three tests, both mutations checked: dropping the filter entirely and fixing only the first page each turn them red.

One consequence that is not in the diff. The panel refuses to rename a history row (#368), and the reason recorded beside that guard is this very bug -- "most of that list is not theirs to rename". With the filter in place the list is theirs, so that reason no longer holds. The guard is still right, for a narrower reason: a rename goes through `PATCH /assets/{id}`, which requires the editor role on the project, and `created_by` does not prove the uploader still holds it. Left alone here, since it is a separate decision.

Refs #374

## Summary

The uploads panel rebuilds its history from `GET /me/assets` and asked for it with no filter. Unfiltered, that endpoint answers with every asset in every project the user is a member of, in whatever role, plus anything merely shared with or assigned to them — so what the panel labels as the user's uploads was in fact everything they can reach. A reviewer who has never uploaded anything opens it to a list of other people's work.

`filter=owned` already exists on the endpoint and resolves to `created_by = <this user>`, which is literally "uploaded by them". Both pages of the history ask for it now. Closes #374.

## Changes

- `apps/web/stores/upload-store.ts` — `fetchHistory` and `fetchMoreHistory` both send `filter=owned`. Fixing only the first would have let the first "load more" put the other assets back, which is why the test covers both pages rather than the first request.
- `apps/web/stores/__tests__/upload-history-filter.test.ts` — new, three tests: the first page asks for the filter, the second page asks for it too, and the filter survives paging with the skip advancing as before.
- `CHANGELOG.md` — entry under Unreleased → Fixed.

Paging is unaffected: `skip` and `limit` run against the filtered set, so a shorter list only ends the paging sooner. Nothing else changes — the rows themselves, the merge with uploads still in flight, and the empty state behave exactly as before.

One consequence that is not in the diff, and that I have deliberately not acted on. The panel refuses to rename a history row (#368), and the reason recorded beside that guard is this very bug: "most of that list is not theirs to rename". With the filter in place the list *is* theirs, so that reason no longer holds. The guard is still right, for a narrower reason — a rename goes through `PATCH /assets/{id}`, which requires the editor role on the project, and `created_by` does not prove the uploader still holds it. Whether that narrower case is worth keeping the guard for is your call, and it belongs in its own change either way.

## Testing

- [x] Backend tests pass — 532 passed. Nothing backend changed; run to confirm the branch is clean.
- [x] Frontend tests + build pass — 78 files, 623 tests, all green. `tsc --noEmit` clean, `pnpm lint` reports only the three pre-existing `exhaustive-deps` warnings in files this does not touch, `pnpm build` succeeds.
- [ ] Tested manually in browser — **not done, and I would rather say so than tick it.**

What stands in place of the browser check: the endpoint was read rather than assumed (`apps/api/routers/me.py:32-35`, `filter == "owned"` → `Asset.created_by == current_user.id`), and the behaviour was confirmed against a real database when #374 was filed — a reviewer with no uploads of their own gets one foreign asset unfiltered and none with the filter, in a rolled-back transaction on a live instance. Both mutations of the fix were checked: removing the filter entirely and fixing only the first page each turn the new tests red.

What that leaves uncovered is the rendering itself — that the panel still populates and still pages in an actual browser, which no test in this repo can show. If you would rather have that before merging, say so and I will get you a screenshot.
## Checklist

## Screenshots

None — this changes which assets a list contains, not how anything is drawn.


ravirajsinh45:
@Lennart-Pingpong the change is right and the tests hold: all four mutations redden the ones you would expect, and the third test is redundant rather than vacuous (after a `[]` first page the store sets `historyHasMore: false`, so the `setState` overwrites a decision instead of restoring one, and the two requests it records are identical to tests 1 and 2). Rather than send a round for text, we have pushed the fixes as `234e210a`.

**The whole fix was one token with nothing pinning it.** There was no API test for `/me/assets` at all. Changing `if filter == "owned"` to a typo falls through to the unfiltered `else`, puts every reachable asset back in the panel, and left 532 API tests and all three of your new frontend tests green. Two tests against a real database close it.

**The second one pins the tradeoff rather than the fix.** `filter=owned` is `Asset.created_by`; each upload's authorship is `AssetVersion.created_by`, and `POST /assets/{id}/versions` never touches the asset's. So a version uploaded onto a colleague's asset is the user's upload by every meaning of the word and is not returned. #374 names this as the known cost of the narrow filter, so it is now behaviour under test: widening the branch to the `AssetVersion.created_by` form reddens it, which makes that a visible decision rather than a silent one. Your instinct to leave it out of scope was right; leaving it out of the record was not.

**You called the stale prose in your own PR body and left it.** Agreed on the decision, not on the record. The `fromHistory` docblock is where the guard is explained, and "a reviewer's own history is full of assets they cannot touch" is not what a `filter=owned` history is. The guard is still right for the narrower reason you wrote in the body and not in the code: that filter carries no membership term, so an asset outlives the demotion or removal that took the right to rename it. Same reason restated in its own words in `upload-rename.test.ts` and `uploads-panel-rename.test.tsx`, including a test name that prints in CI and a fixture called "Someone elses cut" which is no longer what the row is.

**The release notes contradicted themselves.** The `[Unreleased]` entry for #330/#368 ships in the same release as this one and still told the reader the panel asks `/me/assets` with no filter. Amended to the narrower reason.

Your own entry lost two claims it cannot support: "`created_by = <this user>`, which is literally 'uploaded by them'" is the equivalence the version case breaks, and "the paging behave exactly as before" contradicted your own correct note that a shorter list ends the paging sooner. "A user's own files were outnumbered by files they had no hand in" is now stated as the shape it is rather than a count nobody took.

Two small things not worth a change: your testing note says lint reports the three pre-existing `exhaustive-deps` warnings, and it reports 18. And we are keeping "Closes #374" and filing the `AssetVersion.created_by` filter as its own issue, so the option you did not take stays on the record.

No screenshot needed. This changes which rows a list contains, not how they are drawn, and the paging path is covered.

534 API tests against postgres 15 (532 before), 623 frontend across 78 files, `tsc` and `eslint` clean. Approving once CI is green.

ravirajsinh45 (approved):
CI green on `234e210a`: Backend Tests, Frontend Build, Lint and Docker Build all pass. Approving. Shout if you want any of what we pushed reverted, otherwise we will merge this.

